PR #559, opened 2014-07-02
Show hot topics and recently added votes on MP page

### LINKED ISSUE #453: No way to select further details of votes on touch devices
Section needs some design attention on mobile as well, probably

zarino:
A two-column display of "hot topics" and "recent votes" as defined in [Policies.php](https://github.com/mysociety/theyworkforyou/blob/551-voting-summary/classes/Policies.php#L101)

summary

votes

This pull request also includes some refactoring of the vote / "dream_details" button which links each vote description to PublicWhip. And it fixes #453 by always showing the source links, even before hover.

Right now, votes are shown *without* the little green / red / orange dot to signify for / against / mixed. We'll need **developer help** to pass the right variables through to the template, at which point the dots can be revealed by adding a class of `vote-description__{for,against,mix}` to each `<li>` as applicable.

Commit https://github.com/mysociety/theyworkforyou/commit/3e3caf3bd3762183d16987209754091719928d4c replaces the "external link" icons on the main MP profile page with little iOS-esque arrows. I'm not convinced it works as well. I'll revisit later.


tmtmtmtm:
Most of this seems to have been obsoleted by other work on this page. Is there anything from here we should salvage?

zarino:
> Is there anything from here we should salvage?

@tmtmtmtm: Not from a design perspective.

struan:
I don't think so, no.

PR #643, opened 2014-08-11
Refactoring 1: Autoloading (Utilities)

jacksonj04:
This PR includes changes which move various 'utility' functions from scripts which are `include`d to using autoloaded classes and static methods. A further PR will move existing classes to autoloadable namespaces and remove additional inclusions.

--

Move classes and supporting functions which are currently being made available using `include`, `require` and their variants into classes which are discoverable by the Composer autoloader (using the [PSR-4 autoloader specification](http://www.php-fig.org/psr/psr-4/)). These classes will only be loaded when actually required.

Also, where sensible to do so, breaks documentation out of top-of-file comments and into per-function docblocks which can be parsed by [phpDocumentor](http://www.phpdoc.org/) to produce HTML documentation on TheyWorkForYou internals. If you fancy trying this, run `/vendor/bin/phpdoc` and check out the contents of the [theyworkforyou.dev/docs](http://theyworkforyou.dev/docs) folder (assuming you're under Vagrant).

#### Wait, isn't this just a lot of unnecessary change for the sake of change?

No – at the moment there are loads of functions and classes which are sat in files throughout the TWFY folder structure. These are only made available to the code if the file is explicitly included,which can lead to exciting maintainability problems since it's not clear where any piece of functionality has actually arisen.

With autoloading, all classes are kept in a single location (specifically in this case the `classes` folder) organised based on their namespace. You can easily work out exactly where every single class originates, and since utility functions have been moved into static functions it's equally simple to figure out which file a function comes from without having to grep through the codebase.

##### Okay, so why all the new namespacing and case changes?

Moving the classes into their own `MySociety\TheyWorkForYou` namespace isn't essential, but does mean that all the classes are neatly encapsulated away from [other things](https://packagist.org/) which we may (or may not) choose to use in the future, such as routing layers or database abstraction.

The case changes also mean that the code is more compliant with [PSR-1](http://www.php-fig.org/psr/psr-1/) (because we like coding standards).

#### Will this change how things behave?

No. There should be no functional changes in the site at all, and if there are any then this refactoring hasn't gone at all as planned.

##### If it doesn't have any actual changes, why bother?

Although there aren't any user-facing changes, or any functional changes which make things better from an application standpoint or fix bugs, the overall quality and robustness of TheyWorkForYou's code will improve. By organising classes in a consistent and coherent manner it's also easier for developers to find things.

##### If there are no functional changes, why does Scrutinizer report changes in code quality and coverage?

Moving things like utility functions into classes and away from includes exposes them to analysis tools, which can then be profiled and which may cause changes in coverage and code quality.

#### Will this slow things down?

[No](https://athos.blogs.balabit.com/2011/03/php-autoload-performance/). It might even speed things up since we won't be loading a lot of functionality which is never used, and the optimised autoloader (which is generated on production) is crazy fast.

#### Why wasn't TheyWorkForYou written like this to start with? There must be a good reason!

When TheyWorkForYou was written PHP didn't support proper OOP, namespaces or autoloading. It now does, and we've dropped support (as of 0dcdd86b488d3b75ebb14f64b8fd316b03c63746) for anything below PHP 5.4, so there's no reason to not tidy up the codebase, be more ready for future work and make our code analysis tools happy whilst we're at it.

#### There are loads of build failures! I can't merge this!

The failures are coming from Scrutinizer, not Travis, so the build is passing all the unit tests. The failures are because 'new' elements are becoming visible to the analysis as they are moved into classes, or moved to different classes. Since some of these fall below our entirely arbitrary 'new code' quality threshold (currently a 'D' in Scrutinizer) the build is being failed.

Improving these elements is a Good Thing™, and should be done eventually, but is outside the scope of this refactoring. The work moving code into classes will, however, make future analysis more useful in determining which code is in need of improvement.


www/docs/video/next.php
```diff
@@ -50,7 +49,7 @@
                 order by hpos desc limit 1");
             $adate = $q->field(0, 'adate');
             $atime = $q->field(0, 'atime');
-            $videodb = video_db_connect();
+            $videodb = \MySociety\TheyWorkForYou\Utility\Video::dbConnect();
             if ($videodb) {
                 $video = video_from_timestamp($videodb, $adate, $atime);
```
dracos: This call has not been fixed.

classes/Utility/Alert.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Alert Utilities
+ *
+ * Utility functions related to alerts
+ */
+
+class Alert
+{
+
+    public static function confirmationAdvert($details) {
```
dracos: Doesn't look like this is used anywhere any more.

classes/Utility/Calendar.php
```diff
@@ -0,0 +1,116 @@
...
+        $db = new \ParlDB();
+
+        $q = $db->query("SELECT * FROM future
+            LEFT JOIN future_people ON future.id = future_people.calendar_id AND witness = 0
+            WHERE event_date = '$date'
+            AND deleted = 0
+            ORDER BY chamber, pos");
+
+        if (!$q->rows()) {
+            if ($date >= date('Y-m-d'))
```
dracos: Good to get some brackets here.

www/docs/mp/index.php
```diff
@@ -556,9 +555,9 @@ function get_mp_by_constituency($constituency) {
 function get_regional_by_user_postcode($pc, $page) {
     global $this_page;
     $this_page = "your$page";
-    if ($page == 'msp' && postcode_is_scottish($pc)) {
+    if ($page == 'msp' && \MySociety\TheyWorkForYou\Utility\Postcode::postcodeIsScottish($pc)) {
         regional_list($pc, 'SPC', $page);
-    } elseif ($page == 'mla' && postcode_is_ni($pc)) {
+    } elseif ($page == 'mla' && \MySociety\TheyWorkForYou\Utility\Postcode::postcodeIsScottish($pc)) {
```
dracos: This is the wrong function.

www/docs/api/api_getConstituency.php
```diff
@@ -29,7 +29,7 @@ function api_getConstituency_postcode($pc) {
         return;
     }
 
-    $constituency = postcode_to_constituency($pc, true);
+    $constituency = MySociety\TheyWorkForYou\Utility\Postcode::postcodeToConstituency($pc);
```
dracos: This function has lost its second argument.

www/docs/api/api_getPerson.php
```diff
@@ -110,11 +110,11 @@ function api_getPerson_postcode($pc, $house) {
         $types = array('SPC', 'SPE');
     }
     if (validate_postcode($pc)) {
-        $constituencies = postcode_to_constituencies($pc, true);
+        $constituencies = MySociety\TheyWorkForYou\Utility\Postcode::postcodeToConstituencies($pc);
```
dracos: Ditto loss of second argument.

dracos:
A little bit concerning that it appears the function name changes weren't done with global search and replace (as I can't see how the Scottish/NI thing could have happened if it had :) ), so I hope I haven't missed any more changes like that. A few minor things to fix, otherwise okay :)

PR #725, opened 2014-12-11
Ignore additional structural tags in regmem XML

andylolz:
This update will parse the regmem XML structure currently output by parlparse, as well as the modified structure proposed in mysociety/parlparse#10.

### DIFF
--- a/scripts/mpinfoin.pl
+++ b/scripts/mpinfoin.pl
@@ -383,8 +383,7 @@ sub loadregmeminfo
                 $htmlcontent .= '<div class="regmemcategory">';
                 $htmlcontent .= $category->att("type") . ". " . $category->att("name");
                 $htmlcontent .= "</div>\n";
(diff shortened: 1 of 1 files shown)


dracos:
I would be happy for this to actually do something with the new structure! :) Put them in a wrapper div, that can then be styled?

andylolz:
Hi Matthew! I have a question about this one (sorry!) Do you think you’l retrospectively update the old XML that drives TheyWorkForYou, so it all has the new structure? (That would be awesome.) If so, it probably makes adding divs here a bit easier.

dracos:
Hmm, it seems doubtful anyone is relying on them, so I don't see why not. They can always regenerate them old style if need be anyway.

struan:
Closed in a925b88f14e106395cc90cdcf53996d30c59bf2a

PR #743, opened 2015-02-04
Fix missing bracket in MP URL SQL

### LINKED ISSUE #739: Can't edit MP's links
As a superuser I have the authorisation to add homepage links for MPs, but I've just tried changing Mark Tami's URL to http://www.marktami.co.uk/ (at his request) and it has had no effect.

struan:
End bracket from INSERT statement was missing causing it to fail.

Fixes #739


dracos:
:+1:

Issue #81, opened 2011-06-06
More sharing options


dracos:
"If you're reading a speech, a way of sharing that particular speech (or the whole debate) with Twitter, Facebook, etc would be good. Needs thought on how to do well."

MyfanwyNixon:
Let's prioritise this! Along with #1124, this feels like a way we could get massively more shares and visits to the site, thus expanding the number of people who have heard of us and expanding out of the network of people who already know about TWFY.

It's 2016, more and more people have flocked to the welcoming arms of Facebook, politics has gone bananas, and everyone wants a quick way to share a debate and an MP's voting record.

dracos:
For debates, the "select some text and it pops up a share box" seems to be more common nowadays (e.g. Guardian, Daily Mail, Medium), that could be a way to help with debate pages? I know how to use my browser's share button for a whole page, I don't know how common that is.

MyfanwyNixon:
My browser has a share button?

(I know Safari on mobile does but I've not noticed one on Firefox). 

That aside I think the text select/share would be nice, if discoverable. It's like a helping hand for anyone who has started on the weary path of copying and pasting into FB, I guess. It wouldn't capture the people who might already have decided that C+P is too much effort...

PR #755, opened 2015-02-13
retain search term over multiple redirects

### LINKED ISSUE #754: Retain search highlighting through a redirect
@EdwardBetts: “This URL:
http://www.theyworkforyou.com/debates/?id=2015-02-10a.596.1&s=Facebook#g597.4
redirects to this URL:
http://www.theyworkforyou.com/debates/?id=2015-02-10c.596.1#g597.4
It should be possible to keep the highlight parameter from the first URL (s=Facebook) during the redirection.”

struan:
if the link from the search page, or more realistically an email goes
through more than one level of redirect then the search term is lost
from it. Check if it's present and add it back in to the redirect.

Note that we will always lose and anchor part of the URL in this process
as that's all dealt with purely client side and never supplied to the
server.

Fixes #754

### DIFF
--- a/classes/SectionView/SectionView.php
+++ b/classes/SectionView/SectionView.php
@@ -97,6 +97,11 @@ class SectionView {
             } else {
                 $URL->insert( array('id'=>$e->getMessage()) );
             }
+            # put the search term back in so highlighting works.
+            # NB: as we don't see the # part of the URL we lose this :(
+            if ( $args['s'] !== '' ) {
+                $URL->insert( array('s'=>$args['s']) );
+            }
             redirect($URL->generate('none'));
         }
 


classes/SectionView/SectionView.php
```diff
@@ -97,6 +97,11 @@ protected function display_section_or_speech($args = array()) {
             } else {
                 $URL->insert( array('id'=>$e->getMessage()) );
             }
+            # put the search term back in so highlighting works.
+            # NB: as we don't see the # part of the URL we lose this :(
+            if ( isset($args['s']) && $args['s'] !== '' ) {
```
dracos: I think going by line #81, `s` will always be set, so this can just be the second half statement.

classes/SectionView/SectionView.php
```diff
@@ -97,6 +97,11 @@ protected function display_section_or_speech($args = array()) {
...
+                $URL->insert( array('s'=>$args['s']) );
```
dracos: Bit more spacing around the `=>` maybe?  see it matches elsewhere.

dracos:
:+1:

PR #757, opened 2015-02-16
Hansard citations no longer hidden, closes #752

### LINKED ISSUE #752: CSS-hidden Hansard references in debates cause jitter
I note that in the redesigned debate pages the Hansard uses CSS to change on hover. 

(Personally I don't see why they should be hidden, as they're already de-empahsised in the styling, and the use of hover implies a desktop-centric design.)

Since the hidden block is on a new line, and not inline with the hover, you get an odd jitter effect if you scroll down the page with a cursor on the screen - as you pass over each hover hotspot the next debate entry slides down, and then back up, to accommodate it.

wrightmartin:


struan:
Looks good.

dracos:
Merged as 6b23ae8

PR #762, opened 2015-02-19
Import voter lists for Lords as well as Commons.

dracos:


struan:
Looks fine to me. Only small thing is when I run it you end up with 

    B. [Teller] Butler-Sloss (teller)

for the teller names in the Lords which might be nice to tidy up.

Issue #105, opened 2011-12-09
Easy way to get statistics


abimysoc:
We quite often need statistics to give to people like funders.  Struan built a kick-ass awesome stats feature in the admin interface for FMS which means it takes seconds for me to extract the numbers myself.  We should build one for TWFY as well.

struan:
There is a very basic version of this now in the admin.

PR #769, opened 2015-02-27
add a reporting stats page to admin pages

struan:
tracks monthly signups so devs don't have to rootle this out the DB.

Start of #105


jacksonj04:
Assuming Scrutinizer comes back clean, :+1:

Issue #773, opened 2015-03-03
Design and implement page for displaying policy votes


struan:
This is a page to display the individual votes that make up a policy.

The initial version will display all the voted associated with a policy in date order, most recent first, for an MP showing how they voted ( yes, no or absent ).
Later versions should initially only show strong votes by default with an option to show all votes.
If no policy is specified it will show all policies with the votes for each one.
It will be part of the MP page with a new tab and also linked to from the more details button of individual policies.

See #576 for some discussion on this.

TomSteinberg:
So the MVP for an MP page is

1 - Hero image
2 - Repetition of the policy name
3 - Repetition of this MPs position on the policy
4 - Description of what the policy means
5 - One row for each vote inside the policy, using Richard's hand written 'aye' and 'nay' text
6 - Each vote links to debate where vote happened

zarino:
Pull request for this, so far today, here: https://github.com/mysociety/theyworkforyou/pull/781

It doesn't yet have the "Do you agree" button. And there are a few unanswered questions, which I've outlined in the pull request.

struan:
Things stil to decide/do:

How to handle tellers.
How to display strong/not strong votes.
Images etc for all the policies.
Default image for a policy.

PR #781, opened 2015-03-04
Example markup and styling for voting policy page

zarino:
This one’s for @struan.

I’ve used some hard-coded data at the top of `www/includes/templates/html/mp/divisions.php` as you’ll see.

Other things to note:

* We should handle the case where the policy (eg: "Hunting Ban") has no hero image, or no summary description, etc. Apart from assigning a generic hero image, this PR does nothing to handle this situation.
* We should handle the case where we have no "human" summary for an individual vote. Maybe something like "[MP name] voted “[aye/no]” in [vote title]"?
* We should handle the case where the MP was absent. Perhaps something like "[MP name] was absent for the vote on [vote title]"?

I'm also undecided on:

* Whether we should differentiate between what PublicWhip called "strong" votes and all the other votes. The screenshot below shows them smaller and greyed out, but I wonder whether a) it looks a bit weird when they're different with no explanation why, and b) people really won't care. Thoughts @wrightmartin @struan @dracos?
* The links to "Show full debate" – from what I can tell, the full debate is often [a page like this](http://www.theyworkforyou.com/debates/?id=2002-03-18.136.2), which isn't very useful for the casual browser. Am I wrong, and stub pages like that are the exception?

Here's how it looks:

hunting-ban

Part of #773.


dracos:
On the last point, I imagine it should be possible to link to the whole debate page (e.g. http://www.theyworkforyou.com/debates/?id=2002-03-18.136.0#g136.2 in this case, with or without the hash reference) which is a bit better (or if we could do the previous speech to have the what they're voting on bit). Modern imports will have a full list of voters there instead of just that line, but lots of them will be that line, yes.

On weak/strong, I think the strong ones should have some sort of label, perhaps just with a key at the bottom saying these votes have been given more weight in the policy calculation?

wrightmartin:
Regarding the weights of votes I was going to say similar to @dracos, it's more interesting that a vote is strong, rather than weak, so the label (icon, indicator, whatever) should be on the strong votes. 

In it's current design the weak votes could be mistaken children of the strong votes, which isn't the case.

PR #783, opened 2015-03-09
Stop regular expression segfault.

dracos:
The previous regular expression could easily run out of stack space when
run on something with a large number of HTML elements. The new one is a)
simpler, b) explained, and c) possessive, so hopefully won't use any.

### DIFF
--- a/www/includes/wikipedia.php
+++ b/www/includes/wikipedia.php
@@ -90,11 +90,19 @@ function wikipedize($source) {
         continue 2;
     }
 
-    # Go ahead
     twfy_debug("WIKIPEDIA", "Matched '$phrase'");
     # 1 means only replace one match for phrase per paragraph
-    # The regex here ensures that the phrase is only matched if it's not already within <a> tags, preventing double-linking. Kudos to http://stackoverflow.com/questions/7798829/php-regular-expression-to-match-keyword-outside-html-tag-a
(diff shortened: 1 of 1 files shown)


www/includes/wikipedia.php
```diff
@@ -90,11 +90,19 @@ function wikipedize($source) {
...
 
-    # Go ahead
     twfy_debug("WIKIPEDIA", "Matched '$phrase'");
     # 1 means only replace one match for phrase per paragraph
-    # The regex here ensures that the phrase is only matched if it's not already within <a> tags, preventing double-linking. Kudos to http://stackoverflow.com/questions/7798829/php-regular-expression-to-match-keyword-outside-html-tag-a
-    $source = preg_replace ('/\b(' . $phrase . ')\b(?!(?>[^<]*(?:<(?!\/?a\b)[^<]*)*)<\/a>)/', "<a href=\"http://en.wikipedia.org/wiki/$wikistring\">\\1</a>", $source, 1);
+    $source = preg_replace ('{
+    \b(' . $phrase . ')\b    # Match the phrase itself
+    (?!                      # Match as long as the following does *not* apply:
+        (                    #   Match, possessively, as many strings of:
```
struan: not sure if it makes any difference but I might make this (?: just to be clear we're not capturing it.

dracos: Ooh yes, I thought it was (I think it's because it used to be a `?>` and I removed that in preference of possessiveness).

struan:
Other than small comment about capturing looks fine to me.

PR #786, opened 2015-03-11
Tidy member class file

jacksonj04:
As part of Nick's ongoing quest of teeny tiny bits of code embetterment:

* Pull supporting functions out of the Member class file which aren't actually part of the class
* Remove functionality which is no longer used.
* Tests!


dracos:
:+1: (as I was deploying, I rebased this in master as I was there)

PR #790, opened 2015-03-24
Add a page to display the individual votes that make up a Policy

struan:
Add a /divisions page for each MP that details how they voted on all of the individual votes that make up a policy.

Includes import of this data from PW generated JSON files.


db/schema.sql
```diff
@@ -108,6 +108,44 @@ CREATE TABLE `memberinfo` (
...
+  `image_license` varchar(200) default '',
+  `image_license_url` text,
+  `image_source` text,
+  UNIQUE KEY `policy_id` (`policy_id`)
+);
+
+CREATE TABLE `policydivisions` (
+  `division_id` varchar(100) NOT NULL,
+  `policy_id` varchar(100) NOT NULL default '',
+  `house` varchar(100),
```
dracos: Do we want a separate division (or divisions) table with the division only bits?

scripts/json2db.pl
```diff
@@ -0,0 +1,238 @@
...
+
+        my $sources = $motion->{motion}->{sources};
+        my $gid = '';
+        foreach my $source (@$sources) {
+            if ( defined $source->{gid} ) {
+                $gid = $source->{gid};
+            }
+        }
+
+        my $motion_id = $motion->{motion}->{id} . "-$house";
```
dracos: I don't think this is unique (not even by policy perhaps), as the ID repeats every day in the Lords, and every Parliament in the Commons. If the house is included in the ID (as I think it has to be, as otherwise it may not be unique), then you might want to use the ID as is?

scripts/json2db.pl
```diff
@@ -0,0 +1,238 @@
...
+
+my $votecheck = $dbh->prepare("SELECT member_id, vote FROM memberdivisionvotes WHERE division_id = ?");
+my $voteadd = $dbh->prepare("INSERT INTO memberdivisionvotes (member_id, division_id, vote) VALUES (?, ?, ?)");
+my $voteupdate= $dbh->prepare("UPDATE memberdivisionvotes SET vote = ? WHERE member_id = ? AND division_id = ?");
+
+my $motionsdir = $parldata . "scrapedjson/policy-motions/";
+
+$motion_count = $policy_count = $vote_count = 0;
+
+foreach my $dreamid (
```
dracos: If this is the same list as in mpinfoin.pl, could we share it somewhere? Perhaps then even in a CSV file that classes/Policies.php could read in too… (but just sharing between the two Perl files would be good :) )

www/docs/style/sass/pages/_mp.scss
```diff
@@ -461,6 +559,36 @@
...
+      padding-right: 7em;
+
+      @media (min-width: $medium-screen) {
+          padding-right: 10em; // make room for longer "Show full debate" button
+      }
+  }
+}
+
+.policy-vote--minor {
+    display: none;
```
dracos: It's impossible to see minor votes if JavaScript is disabled. This should be visible by default, with a `.js `.policy-vote--minor` to hide it, and then to show the button to see all.

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -0,0 +1,206 @@
...
+        <div class="full-page__unit">
+            <div class="person-navigation page-content__row">
+                <ul>
+                    <li><a href="<?= $member_url ?>">Overview</a></li>
+                    <li class="active"><a href="<?= $member_url ?>/votes">Voting Record</a></li>
+                </ul>
+            </div>
+            <div class="person-panels page-content__row">
+                <div class="sidebar__unit in-page-nav">
+                    <?php if ( isset($policyinformation) && isset($policies) && $policies && isset($policydivisions) && $policydivisions ) { ?>
```
dracos: Could this just be `if ( isset($policydivisions) && $policydivisions )` as further down? Don't think the others can ever be set if that's not.

db/schema.sql
```diff
@@ -108,6 +108,44 @@ CREATE TABLE `memberinfo` (
...
+  `division_number` int(11),
+  `lastupdate` timestamp NOT NULL default CURRENT_TIMESTAMP on update CURRENT_TIMESTAMP,
+  UNIQUE KEY `policy_division` (`division_id`, `policy_id`),
+  KEY `division_id` (`division_id`)
+);
+
+CREATE TABLE `memberdivisionvotes` (
+  `member_id` int(11) NOT NULL default '0',
+  `division_id` varchar(100) NOT NULL default '',
+  `vote` text NOT NULL,
```
dracos: Should this be an enum if this is a restricted set of values?

classes/Divisions.php
```diff
@@ -0,0 +1,179 @@
...
+
+    private function divisionUrlFromGid($gid) {
+        $q = $this->db->query("SELECT gid FROM hansard WHERE epobject_id = ( SELECT subsection_id FROM hansard WHERE gid = :gid )", array( ':gid' => $gid ));
+        $parent_gid = $q->field(0, 'gid');
+        if ( !$parent_gid ) {
+            return '';
+        }
+        $page_type = preg_replace('#.*/(.*)/.*#', '\1', $parent_gid);
+        $parent_gid = fix_gid_from_db($parent_gid);
+        $url = new \URL($page_type);
```
dracos: I think a (slightly) better way of getting the URL would be to fetch the `major` in the query too, and then use `$hansardmajors[major]['page']`.

classes/Policies.php
```diff
@@ -303,4 +313,37 @@ public function limitToSet($set) {
...
+            'image_attribution' => $q->field(0, 'image_attrib'),
+            'image_source' => $q->field(0, 'image_source'),
+            'image_license_url' => $q->field(0, 'image_license_url')
+        );
+
+        $image = $q->field(0, 'image');
+
+        if ( $image && file_exists(BASEDIR . '/' . $image)) {
+            $props['image'] = $image;
+            $props['image_license'] = $q->field(0, 'image_license');
```
dracos: These four lines repeat what is done above; not sure if these should be removed, or if the four above should be changed to default data?

scripts/json2db.pl
```diff
@@ -0,0 +1,238 @@
...
+
+        my $motion_id = $motion->{motion}->{id} . "-$house";
+
+        # JSON is UTF-8, the database and TWFY are not
+        my $text = Encode::encode( 'iso-8859-1', $motion->{motion}->{text} );
+        my $curr_motion = $dbh->selectrow_hashref($motioncheck, {}, $motion_id, $dreamid);
+
+        if ( $curr_motion ) {
+            $curr_motion->{direction} ||= '';
+            $motion->{direction} ||= '';
```
dracos: Can motion->direction ever be unset? Won't the enum die in this case?

scripts/json2db.pl
```diff
@@ -0,0 +1,238 @@
...
+        # just skip them
+        if ( $motions_seen{$motion_id} ) {
+            #print "seen $motion_id already, skipping\n";
+            next;
+        } else {
+            $motions_seen{$motion_id} = 1;
+        }
+
+        my $curr_votes = $dbh->selectall_hashref($votecheck, 'member_id', {}, $motion_id);
+        unless ($curr_votes) {
```
dracos: This will always be true, as `{}` is true, I think?

scripts/populate-policy-images.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Doesn't fetch the images but assumes they've been
+ * added in www/docs/images/ as $policy_id.jpg
+ *
+ * Take the path to the CSV file with the details
+ * Columns should be:
+ * 1: policy id
+ * 5: source: url of the original image
+ * 6: image attribution - name of photographer
+ * 7: license URL - url of the CC licence of photo
```
dracos: The code has columns 1,6,7,8 :)

tests/DivisionsTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function testPolicyDirection() {
+        $page = $this->fetch_division_page();
+        $this->assertContains('voted strongly against introducing <b>foundation', $page);
+    }
+
+    public function testVotedAgainst() {
+        $page = $this->fetch_division_page();
+        $this->assertContains('Test Current-MP voted Do not agree', $page);
```
dracos: I guess it might be nice to check the Do note agree and Agreed are on the right lines :)

tests/FetchPageTestCase.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+/**
+ * Provides acceptance(ish) tests for API functions.
+ */
+abstract class FetchPageTestCase extends PHPUnit_Extensions_Database_TestCase
```
dracos: Woo

www/docs/admin/policies.php
```diff
@@ -0,0 +1,142 @@
+<?php
+
+include_once '../../includes/easyparliament/init.php';
+#include_once INCLUDESPATH . 'easyparliament/commentreportlist.php';
+#include_once INCLUDESPATH . 'easyparliament/searchengine.php';
+include_once INCLUDESPATH . 'easyparliament/member.php';
+#include_once INCLUDESPATH . 'easyparliament/people.php';
+
+$this_page = 'admin_policies';
```
dracos: I think this page needs to be present in metadata.php with a parent of admin, otherwise no admin protection will occur on this page.

www/docs/admin/policies.php
```diff
@@ -0,0 +1,142 @@
...
+    $out = '';
+
+    for ($row = 0; $row < $q->rows(); $row++) {
+
+        $out = "<h3>Edit policy " . $q->field($row, 'title') . "</h3>\n";
+
+        $out .= '<form action="policies.php?editpolicy=' . $q->field($row, 'policy_id') . '" method="post">';
+        $out .= '<input name="action" type="hidden" value="Save">';
+        $out .= '<label for="title">Title</label>';
+        $out .= '<span class="formw"><input id="title" name="title" type="text"  size="60" value="' . $q->field($row, 'title') . '"></span>' . "<br>\n";
```
dracos: None of these are being HTML escaped, I don't know if title/description could have " in them...

www/docs/mp/index.php
```diff
@@ -144,14 +146,23 @@
...
+                    if (str_replace('/mp/', '/' . $this_page . '/', get_http_var('url')) !== urldecode($MEMBER->url(FALSE)) . '/divisions')
+                    {
+                        member_redirect($MEMBER, 301, 'divisions');
+                    }
+
+                    break;
+
                 case 'profile':
                 default:
+                    break;
```
dracos: Is this meant to be here? If so, should the code underneath it be removed?

www/docs/mp/index.php
```diff
@@ -606,6 +617,37 @@
...
+                $policiesList = new MySociety\TheyWorkForYou\Policies;
+                $positions = new MySociety\TheyWorkForYou\PolicyPositions( $policiesList, $MEMBER );
+
+                $data['user_code'] = bin2hex(urandom_bytes(16));
+                $data['auth_signature'] = auth_sign_with_shared_secret($data['user_code'], OPTION_SURVEY_SECRET);
+
+                $data['page_url'] = "http://" . DOMAIN . $_SERVER['REQUEST_URI'] . ( $policyID ? '&' : '?' ) . 'answered=1';
+                $data['answered_q'] = $answered_q;
+
+                $data['positions'] = $positions->positionsById;
```
dracos: If the `$policiesList` was restricted if `$policyID` is present, and then the one policy was passed to the `PolicyPositions` constructor, using `positionsById` wouldn't be needed, I think.

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -0,0 +1,206 @@
+    <div class="<?= $current_assembly ?>">
```
dracos: Should this be shared with profile and votes? (Definitely yes, because that would prevent #795 happening again :) )

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -0,0 +1,206 @@
...
+                    <ul>
+                        <li><a href="<?= $member_url ?>/votes">Back to all topics</a></li>
+                    </ul>
+                </div>
+
+                <div class="primary-content__unit">
+
+                    <?php if ($party == 'Sinn Fein' && in_array(HOUSE_TYPE_COMMONS, $houses)): ?>
+                    <div class="panel">
+                        <p>Sinn F&eacute;in MPs do not take their seats in Parliament.</p>
```
dracos: Do we want the SNP line about only voting on Scotland things here too? (Will that apply after the election, fact fans?) In fact, why is the SNP line treated differently to this line in the code, hmm.

dracos:
And phew! Looks good :) A few things, but only the admin checking, JS switching, and ID uniqueness are actually something that might cause breakage I think.

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -0,0 +1,206 @@
...
+                                        </h3>
+
+                                        <?php if ( DEVSITE ) { ?>
+                                        <p class="policy-vote-agree-disagree">
+                                            <button class="button">I agree with this MP</button>
+                                            <button class="button button--negative">I disagree with this MP</button>
+                                        </p>
+                                        <?php } ?>
+
+                                        <h3 class="policy-votes-list-header">Key votes about <?= $policies[$policy['policy_id']] ?>:</h3>
```
dracos: This should be All votes by default (see JS comment elsewhere), change to Key votes if JS is working, then change back to All when someone clicks the see all votes link :)

www/docs/mp/index.php
struan: I quite like having a page with all the policies and votes. It's handy for checking things in development if nothing else :)

dracos: Sorry, I probably wasn't clear. I'm not suggesting getting rid of that page, just that the code seems to have two ways of doing the same thing - restricting Policies before passing to PolicyPositions, or not restricting and using this new positionsById.

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -0,0 +1,163 @@
...
+                                            <p class="policy-votes__byline">Vote information from <a href="http://www.publicwhip.org.uk">PublicWhip</a></p>
+                                            <?php if ( $policy['weak_count'] > 0 ) { ?>
+                                            <p><button class="button secondary-button small js-show-all-votes">Show all votes, including <?= $policy['weak_count'] ?> less important <?= $policy['weak_count'] == 1 ? 'vote' : 'votes' ?></button></p>
+                                            <?php } ?>
+                                        </div>
+
+                                        <script type="text/javascript">
+                                        $(function(){
+                                            $('.policy-vote--minor').addClass('js');
+                                            $('.js-show-all-votes').show();
```
dracos: TWFY adds a `js` class to `body`, so you could not have these two lines, if in the CSS instead of `.policy-vote--minor.js` you had `.js .policy-vote--minor`, and add `.js .js-show-all-votes { display: block; }`, then it would work directly (yay title change :) ).

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -1,77 +1,6 @@
-    <div class="regional-header regional-header--<?= $current_assembly ?>">
-        <div class="regional-header__overlay"></div>
-        <div class="person-header <?= $this_page ?> <?= (isset($data['photo_attribution_text'])?'has-data-attribution':'') ?>">
```
dracos: A number of things have been lost by the merger of the headers – the photo attribution code here and below is one example, others include the fact party might not be present, and the "royal" check for `current_member_anywhere`. I think that's it, but should be checked.

classes/Policies.php
```diff
@@ -234,12 +250,19 @@ class Policies {
...
     public function getArray() {
         $out = array();
-        foreach ($this->policies as $policy_id => $policy_text)
-        {
+        if ( $this->policy_id ) {
             $out[] = array(
-                $policy_id,
-                $policy_text
+                $this->policy_id,
+                $this->policies[$this->policy_id]
```
dracos: Should this throw an exception if policy_id is invalid? (Or should it do that in the constructor?)

classes/Policies.php
```diff
@@ -234,12 +250,19 @@ class Policies {
      */
     public function getArray() {
         $out = array();
-        foreach ($this->policies as $policy_id => $policy_text)
-        {
+        if ( $this->policy_id ) {
```
dracos: If the constructor altered `$this->policies` if policy_id was set, the code here and in limitToSet would be unnecessary, I think?

classes/Policies.php
```diff
@@ -225,6 +225,22 @@ class Policies {
...
+    private $db;
+
+    private $policy_id;
+
+    public function __construct($policy_id = NULL) {
+        $this->db = new \ParlDB;
+
+        if ( $policy_id ) {
+            $this->policy_id = $policy_id;
+        }
```
dracos: It seems odd that if you pass in a policy_id, `getPolicies` still returns a list of all the policies; only `getArray` and `limitToSet` limit themselves. Should it be renamed `getAllPolicies`, or should this be fixed in the constructor as per previous comment.

dracos:
Looking good; the comment on `populate-policy-images.php` is only outstanding thing from first time around looks like (and maybe my test comment, not sure); a few new things to do with the Policies class and the header merger.

dracos:
:rabbit:

PR #792, opened 2015-03-25
Issues/445 regional header backgrounds

wrightmartin:
On 'section' pages and MP pages we now use regional specific backgrounds.

Involved some slight refactoring to both the MP and section pages.


struan:
Woo! This all looks good. Should probably squash it all down into a single commit for merging.

PR #793, opened 2015-03-26
Use person ID rather than member ID in database fields

dracos:


dracos:
(The failure in Travis is due to GitHub's current DDOS.)

struan:
This looks good.

PR #808, opened 2015-04-09
Banner for new style pages plus move content into database

struan:
This adds a banner to pages with the new design.

It also adds an editorial table, which is a key/value thing at the moment in anticipation of it being used for e.g. controlling things on the front page, and fetches the banner text from that. This allows the banner to now be edited in the admin interface rather than having to edit files and deploy the site.

### DIFF
--- a/classes/Memcache.php
+++ b/classes/Memcache.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * Memcache wrapper
+ *
+ * @package TheyWorkForYou
+ */
+
+namespace MySociety\TheyWorkForYou;
+
+class Memcache {
+
+    static $memcache;
+
+    public function __construct() {
+        if (!self::$memcache) {
+            self::$memcache = new \Memcache;
+            self::$memcache->connect('localhost', 11211);
+        }
+    }
+
+    public function set($key, $value, $timeout = 3600) {
+        self::$memcache->set(OPTION_TWFY_DB_NAME . ':' . $key, $value, MEMCACHE_COMPRESSED, $timeout);
+    }
+
+    public function get($key) {
+        // see http://php.net/manual/en/memcache.get.php#112056 for explanation of this
+        $was_found = false;
+        $value = self::$memcache->get(OPTION_TWFY_DB_NAME . ':' . $key, $was_found);
+        if ( $was_found === false ) {
+            return false; // mmmmm
+        } else {
+            return $value;
+        }
+    }
+}
--- a/classes/Renderer.php
+++ b/classes/Renderer.php
@@ -422,6 +422,10 @@ class Renderer
         $data['footer_links']['international'] = self::get_menu_links(array ('newzealand', 'australia', 'ireland', 'mzalendo'));
         $data['footer_links']['tech'] = self::get_menu_links(array ('code', 'api', 'data', 'pombola', 'devmailinglist', 'irc'));
 
+        # banner text
+        $b = new Model\Banner;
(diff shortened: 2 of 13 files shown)


www/docs/admin/banner.php
```diff
@@ -0,0 +1,110 @@
...
+if (get_http_var('action') === 'Save') {
+    $out = update_banner();
+}
+
+$out .= edit_banner_form();
+
+print '<div id="adminbody">';
+print $out;
+?>
+<script language="javascript" type="text/javascript">
```
dracos: Don't really need either the language or type attributes nowadays.

www/docs/admin/banner.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+include_once '../../includes/easyparliament/init.php';
+
+$this_page = 'admin_banner';
+
+$db = new ParlDB;
+
+$scriptpath = '../../../scripts';
```
dracos: Is this variable used?

tests/PageTest.php
```diff
@@ -83,4 +86,12 @@ public function testSpeaker()
         $this->assertContains('<span class="party SPK">Speaker</span>', $page);
     }
 
+    public function testBanner() {
+        $page = $this->fetch_page( array( 'url' => '/' ) );
+        $this->assertNotContains('This is a banner', $page);
+        $this->db->exec("INSERT INTO editorial ( item, value ) VALUES ( 'banner', 'This is a banner' )");
+        $page = $this->fetch_page( array( 'url' => '/' ) );
+        $this->assertContains('This is a banner', $page);
```
dracos: Could test removing the row and checking it is again not present?

classes/Renderer.php
```diff
@@ -443,6 +446,33 @@ public static function output($template, $data = array())
         require_once INCLUDESPATH . 'easyparliament/templates/html/footer.php';
     }
 
+    private static function get_banner_text() {
+        $text = NULL;
+
+        global $memcache;
```
dracos: As this appears a few times here (and elsewhere), perhaps worth a small class to replace this global, that performs the connection when necessary and stores if needed?

dracos:
Presumably needs rebasing on to current master for the tests, too; looks good, think it might as well have a memcache class to reduce the code duplication wherever that's used.

classes/Memcache.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @package TheyWorkForYou
+ */
+
+namespace MySociety\TheyWorkForYou;
+
+class Memcache {
+
+    public function __construct() {
+        global $memcache;
```
dracos: Now this is a class, this should presumably be a private class attribute, not a global? :)

dracos: Sorry, I meant a static attribute.

www/docs/mp/index.php
```diff
@@ -822,12 +822,8 @@ function person_recent_appearances($member) {
...
     $person_id= $member->person_id();
 
-    global $memcache;
-    if (!$memcache) {
-        $memcache = new Memcache;
-        $memcache->connect('localhost', 11211);
-    }
-    //$recent = $memcache->get(OPTION_TWFY_DB_NAME . ':recent_appear:' . $person_id);
+    $memcache = new MySociety\TheyWorkForYou\Memcache;
+    //$recent = $memcache->get('recent_appear:' . $person_id);
```
dracos: As discussed, we think this can be uncommented.

PR #810, opened 2015-04-09
Add link to specific vote on a divisions page

struan:
In case someone wants to highlight a specific vote on a page add a link
to the end of each description. Also, add some JS tweaks to highlight the vote and make sure it's visible.

I'm not sure if this is possibly adding too much visual clutter:

screen shot 2015-04-09 at 17 34 34

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/divisions.php
+++ b/www/includes/easyparliament/templates/html/mp/divisions.php
@@ -107,6 +107,7 @@ include_once INCLUDESPATH . "easyparliament/templates/html/mp/header.php";
                                     ?>
                                     <?php foreach ($policy['divisions'] as $division) { ?>
                                         <li id="<?= $division['division_id'] ?>" class="<?= $division['strong'] || $show_all ? 'policy-vote--major' : 'policy-vote--minor' ?>">
+                                            <a class="policy-vote__permalink" href="<?= $page_url ?>#<?= $division['division_id'] ?>" title="Link to this vote">#</a>
                                             <span class="policy-vote__date">On <?= strftime('%e %b %Y', strtotime($division['date'])) ?>:</span>
                                             <span class="policy-vote__text"><?= $full_name ?><?= $division['text'] ?></span>
                                             <?php if ( $division['url'] ) { ?>
--- a/www/docs/style/sass/pages/_mp.scss
+++ b/www/docs/style/sass/pages/_mp.scss
@@ -590,6 +590,13 @@
     }
 }
 
+.policy-vote--major,
+.policy-vote--minor {
+    &:target {
+        background-color: #ffc;
+    }
+}
+
 .policy-vote--minor {
     padding-left: 1.5em;
     line-height: 1.3em !important; // override .vote-descriptions .policy-votes > li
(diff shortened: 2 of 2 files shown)


www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -106,8 +106,10 @@
                                         ?>
                                         <?php foreach ($policy['divisions'] as $division) { ?>
                                             <li id="<?= $division['division_id'] ?>" class="<?= $division['strong'] || $show_all ? 'policy-vote--major' : 'policy-vote--minor' ?>">
+                                                <a name="<?= $division['division_id'] ?>"></a>
```
dracos: I think the ID on the `li` should be enough for the internal link to work as is.

dracos: That would also mean you could use `::target` for style changes, no need for a selected class.

dracos:
In terms of clutter, I would perhaps suggest having the permalink floated right (guess it'd have to be at the start then, not the end), in the same colour as the date?

struan:
Looking at this again and I realise that with the link to a full debate this still looks cluttered.

Possibly @wrightmartin or @zarino can offer some suggestion to improve it.

screen shot 2015-05-26 at 10 23 39

zarino:
@struan – I agree it looks a bit cluttered. The right-floating vote button style was only really meant to handle a single link per vote.

If we want two links for each vote, maybe we could take a cue from the right-hand links on debate pages…?

screen shot 2015-06-01 at 12 46 26

Still looks a bit busy.

I wonder whether maybe turning the *name of the vote* into a link (to show full debate) might be more obvious. Then the only button on the right would be the "Link to this" button.

I guess it partly boils down to: why is this new "link to this" link here? I can see why it makes sense on long pages like the debate pages (link to a specific speech in a debate) but the list of a single MP's votes on a specific topic will usually be quite short (especially when we automatically filter out "minor" votes). Has somebody asked for this feature, or have you found yourself wanting it?

struan:
It's there as I thought it might be a useful feature for things like twitter to be able to point to a specific vote, and could imagine using, and it looked to be quick to add.

Making the title of the debate a link to the debate is quite a nice idea.

zarino:
Here's how it'd look with the vote names turned into links…

screen shot 2015-06-01 at 14 52 33

Initial reaction is that the links distract too much from the *real* focus of each list item – the MP's stance (for / against / absent). You could mute the link colour, but then it starts looking less like a link, which defeats the purpose…

screen shot 2015-06-01 at 14 58 10

Any thoughts @wrightmartin?

wrightmartin:
I have no problem with the muted link colour at all, in terms of usability. It's underlined, it's a different colour - it's a link, not sure if it's a link? Hover on it or tap it and see.

However, it's a new pattern for the MP pages. Everywhere else, links in the content (amongst the black text) are blue and underlined, I'd go with that, it's consistent.

screen shot 2015-06-02 at 08 51 42

I'm slightly worried about the point you made 
> Initial reaction is that the links distract too much from the real focus of each list item
but I think I'd need to see some evidence of this before we can justify introducing a new link style to counteract it.

wrightmartin:
Actually, thinking about this further - the use case for this PR isn't a common one, should we bend and break the usability of the page to this degree for a minority use-case?

struan:
I'm closing this as there's no user demand for it and we don't seem to have a solution that doesn't impair the usability of the page.

PR #811, opened 2015-04-10
extract test GetConnection out to abstract class and inherit

struan:
rather than having the same getConnection method in all the database
tests extract it out to an abstract class and inherit from that.
Also move the includes for it and the FetchPage test class to the
bootstrapper so they don't need to be included in every file.


dracos:
Looks good.

PR #814, opened 2015-04-13
Add an API pricing page

### LINKED ISSUE #152: Add a new page detailing TWFY API pricing, current customers and so on

stevenday:
Adds a basic API pricing page with a table of prices and links to it from the
API homepage.

@struan, @jacksonj04, @dracos - When you have a mo, would you mind sanity
checking the way I've added this page to make sure it fits in with the way
TWFY does things?

@BenJam, @MyfanwyNixon, @steiny, @abimysoc - I'm guessing the copy and pricing
details here are going to need some finesse - consider this a starter for ten!

I'll add a screenshot in a mo..

Closes #152

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -450,6 +450,7 @@ CREATE TABLE `api_key` (
   `created` datetime NOT NULL,
   `disabled` datetime default NULL,
   `reason` text NOT NULL,
+  `estimated_usage` int(9) NOT NULL default 0,
   PRIMARY KEY (`id`),
   UNIQUE KEY `api_key` (`api_key`)
 );
--- a/www/docs/api/index.php
+++ b/www/docs/api/index.php
@@ -142,27 +142,76 @@ function api_front_page($error = '') {
     if ($error) {
         print "<p style='color: #cc0000'>$error</p>";
     }
+
+$em = join('&#64;', array('enquiries', 'mysociety.org'));
+
 ?>
-<p>Welcome to TheyWorkForYou's API section, where you can learn how to query our database for information.</p>
 
 <h3>Overview</h3>
+<p>
+    Welcome to TheyWorkForYou's API section. The API (Application Programming
+    Interface) is a way of querying our database for information.
+</p>
+<p>
+    To use the API you need <a href="key">an API key</a>, and you may need a
+    license from us.
+</p>
+<p>
+    The documentation for each individual API function is linked from this
+    page: you can read what each function does, and test it out, without
+    needing an API key or to write any code.
+</p>
 
-<ol style="font-size:130%">
-<li>
-<?php if ($THEUSER->loggedin()) { ?>
-<a href="key">Get an API key (or view stats of existing keys)</a>.
-<?php } else { ?>
-<a href="key">Get an API key</a>.
-<?php } ?>
-<li>All requests are made by GETting a particular URL with a number of parameters. <em>key</em> is required;
-<em>output</em> is optional, and defaults to <kbd>js</kbd>.
-</ol>
+<h3>Terms of usage</h3>
+<p>
+    Low volume, charitable use of the API is free. This means direct use by
+    registered charities, or individuals pursuing a non-profit project on an
+    unpaid basis, with a volume of up to 50,000 calls per year. All other use
+    requires a licence, as a contribution towards the costs of providing this
+    service. Please email us at <a href="mailto:<?=$em?>"><?=$em?></a> if your
+    usage is likely to fall into the non-free bracket: <strong>unlicensed users
+    may be blocked without warning</strong>.
+</p>
 
-<p align="center"><strong>http://www.theyworkforyou.com/api/<em>function</em>?key=<em>key</em>&amp;output=<em>output</em>&amp;<em>other_variables</em></strong></p>
+<h3>Pricing</h3>
+<p>
+    &pound;500 per year for up to 50,000 calls (free for charitable usage)<br/>
+    &pound;1,000 per year for up to 100,000 calls<br/>
+    &pound;2,000 per year for up to 200,000 calls<br/>
+    &pound;2,500 per year for up to 300,000 calls<br/>
+    &pound;3,000 per year for up to 500,000 calls<br/>
+</p>
+<p>In addition, we offer a 50% discount on the above rates for charitable usage.</p>
 
-<p>The current version of the API is <em>1.0.0</em>. If we make changes to the
-API functions, we'll increase the version number and make it an argument so you
-can still use the old version.</p>
+<h3>Credits</h3>
+<p>Parliamentary material (that's data returned from getDebates, getWrans, and
+getWMS) may be reused under the terms of the
+<a href="http://www.parliament.uk/site-information/copyright/">Open Parliament Licence</a>.
+Our own data &ndash; lists of MPs, Lords, constituencies and so on &ndash; is
+available under the
+<a href="http://creativecommons.org/licenses/by-sa/2.5/">Creative Commons
+Attribution-ShareAlike license version 2.5</a>.
+
+<p>Please credit us by linking to <a href="http://www.theyworkforyou.com/">TheyWorkForYou</a>
+with wording such as "Data service provided by TheyWorkForYou" on the page
+where the data is used. This attribution is optional if you've paid for use of
+the service.
+
+
+<h3>Technical documentation</h3>
+<p>
+    All requests are made by GETting a particular URL with a number of
+    parameters. <em>key</em> is required; <em>output</em> is optional, and
+    defaults to <kbd>js</kbd>.
+</p>
+<p align="center">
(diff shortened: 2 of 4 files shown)


stevenday:
New page:
screen shot 2015-04-13 at 15 22 56

Wording that links to it:
screen shot 2015-04-13 at 15 23 10

www/docs/api/index.php
```diff
@@ -197,7 +197,7 @@ function with the <em>callback</em> variable, and then that function will be
 <p>Low volume, charitable use of the API service itself is free. This means
 direct use by registered charities, or individuals pursuing a non-profit
 project on an unpaid basis, with a volume of up to 50,000 calls per year.
-Please email us at <a href="mailto:<?=$em?>"><?=$em?></a> for all other use. If you are using
+Please see our <a href="pricing">pricing page</a> for all other use. If you are using
```
jacksonj04: This should either be `pricing.php`, or there should be a `pricing` subfolder containing an `index.php`.

dracos: TheyWorkForYou pages work with or without a `.php`.

abimysoc:
I wrote some text for this purpose a few months ago - now shared here

https://docs.google.com/a/mysociety.org/document/d/1PZ-6cyti_hsyVdYrgYrW_sNs3OD6W15GjBQDLqV-z-4/edit

stevenday:
aha, thanks @abimysoc - so the preference is not to have a separate pricing page at all?

abimysoc:
Oh, I don't really mind whether it's same page or separate page. I just want the pricing to be easy to find.

stevenday:
OK - @abimysoc, I've used your new text, and since it's all so nicely laid out, I don't think a separate page is really necessary:

screen shot 2015-04-13 at 16 24 17

stevenday:
I've also added the requested changes to the new api key form now. The updated form looks like:

screen shot 2015-04-13 at 17 41 30

and if you pick an option that's over 50,000 calls or commercial, you see this message:

screen shot 2015-04-13 at 17 41 51

@abimysoc - is this how you imagined it?

Now this definitely needs a review from someone more TWFY-aware, as my PHP skills are a bit rusty (though I've tried to copy the conventions).

abimysoc:
:+1:

MyfanwyNixon:
Sorry to be late to this thread, but :+1: from me too :)

www/docs/api/key.php
```diff
@@ -10,29 +10,39 @@
...
+<p>TheyWorkForYou API calls require a key, so that we can monitor usage
 of the service, and provide usage stats to you. Please see <a href="/api">the
-API overview</a> for how to use your key.';
+API overview</a> for how to use your key.<br/><br/>
+
+<?php
 if ($THEUSER->loggedin()) {
     if (get_http_var('create_key') && get_http_var('reason')) {
-        create_key(get_http_var('commercial'), get_http_var('reason'));
+        $estimated_usage = intval(get_http_var('estimated_usage'));
```
jacksonj04: The recommendation is that if you don't need bases other than 10 you should do this with `(int) get_http_var('estimated_usage')` instead of `intval()`.

db/schema.sql
```diff
@@ -449,6 +449,7 @@ CREATE TABLE `api_key` (
   `created` datetime NOT NULL,
   `disabled` datetime default NULL,
   `reason` text NOT NULL,
+  `estimated_usage` mediumint(9) NOT NULL,
```
dracos: The schema change doesn't match the migration; if the migration is correct, this should also be `int(9)` and have a `DEFAULT 0`.

www/docs/api/index.php
```diff
@@ -142,27 +142,75 @@ function api_front_page($error = '') {
...
+</p>
+
+<h3>Terms of usage</h3>
+<p>
+    Low volume, charitable use of the API is free. This means direct use by
+    registered charities, or individuals pursuing a non-profit project on an
+    unpaid basis, with a volume of up to 50,000 calls per year. All other use
+    requires a licence, as a contribution towards the costs of providing this
+    service. Please email us at <a href="mailto:<?=$em?>"><?=$em?></a> if your
+    usage is likely to fall into the non-free bracket: <u>unlicensed users
```
dracos: Could this be strong or em instead of u?

www/docs/api/key.php
```diff
@@ -62,31 +72,44 @@
...
+<input id="non_comm" type="radio" name="commercial" value="0"> <label for="non_comm">an individual pursuing a non-profit project on an unpaid basis</label><br/>
+<input id="non_comm_2" type="radio" name="commercial" value="0"> <label for="non_comm_2">a registered charity</label><br/>
+<input id="comm" type="radio" name="commercial" value="1"> <label for="comm">neither of the above</label>
+</p>
 <p><label for="reason">Please describe what you're going to use the key for:</label>
 <br>
 <textarea id="reasons" name="reason" rows=7 cols=40></textarea>
+</p>
+<p>What's your estimated annual API call volume?<br/>
+    <input id="estimated_usage_50k" type="radio" name="estimated_usage" value="50000"> <label for="estimated_usage_50k">50,000</label><br/>
```
dracos: Should this perhaps say "up to 50,000", or the question say 'maximum', otherwise if you know yours will be low usage you might not be sure which to pick.

www/docs/api/key.php
```diff
@@ -62,31 +72,44 @@
...
 <br>
-<h3>Apply for a new key</h3>
+<h3>Get a new key</h3>
 <form action="/api/key" method="post">
-<p>Is your application for:
-<input id="non_comm" type="radio" name="commercial" value="0"> <label for="non_comm">Non-commercial use</label>
-<input id="comm" type="radio" name="commercial" value="1"> <label for="comm">Commercial use</label>
+<p>About you:<br/>
+<input id="non_comm" type="radio" name="commercial" value="0"> <label for="non_comm">an individual pursuing a non-profit project on an unpaid basis</label><br/>
+<input id="non_comm_2" type="radio" name="commercial" value="0"> <label for="non_comm_2">a registered charity</label><br/>
```
dracos: If we'd like to store these two options differently, I think you can just use a new value (presumably "2") to record it?

www/docs/api/index.php
```diff
@@ -142,27 +142,75 @@ function api_front_page($error = '') {
...
-<?php } else { ?>
-<a href="key">Get an API key</a>.
-<?php } ?>
-<li>All requests are made by GETting a particular URL with a number of parameters. <em>key</em> is required;
-<em>output</em> is optional, and defaults to <kbd>js</kbd>.
-</ol>
+<h3>Pricing</h3>
+<p>
+    &pound;500 per year for up to 50,000 calls<br/>
+    &pound;1000 per year for up to 100,000 calls<br/>
```
jacksonj04: Very nitpicky, but given all other values are formatted with commas, should these be `1,000`, `2,000`, `2,500` and `3,000`?

www/docs/api/key.php
```diff
@@ -10,29 +10,39 @@
...
-API overview</a> for how to use your key.';
+API overview</a> for how to use your key.<br/><br/>
+
+<?php
 if ($THEUSER->loggedin()) {
     if (get_http_var('create_key') && get_http_var('reason')) {
-        create_key(get_http_var('commercial'), get_http_var('reason'));
+        $estimated_usage = intval(get_http_var('estimated_usage'));
+        $commercial = get_http_var('commercial');
+        create_key(get_http_var('commercial'), get_http_var('reason'),  get_http_var('estimated_usage'));
```
dracos: This call could pass in the variables just created above :)

dracos:
Looks good.

stevenday:
@dracos, @jacksonj04 - thanks for the comments. I think I've addressed everything except https://github.com/mysociety/theyworkforyou/pull/814#discussion_r28311648 - I guess this is more a question for @abimysoc. Would you like to record whether people say "non-profit" or "registered charity" differently?

abimysoc:
It would be useful to know the difference between the three types, yes. Individuals working unpaid on a non-profit project are not subject to licensing charges, and then the charging schemes are different for charities/commercial usage. So good to be able to separate or exclude categories.

stevenday:
OK - I've changed that @abimysoc, so non-profit, charity and commercial are all recorded separately.

I've pushed my changes as FIXUP commits to keep them clear, could someone confirm when they're happy for me to squash and merge?

jacksonj04:
:+1:

abimysoc:
I've just noticed the original version (before my text crept in) included charity rates as well as the standard ones. I think it would be a good idea to reinstate that actually (free for sub 50k, half price for the other bands). Or at least a sentence saying "we offer a 50% discount for charitable usage".

stevenday:
I've made @abimysoc's suggested change, and rebased everything up - assuming the tests etc pass, am I good to merge this?

jacksonj04:
Looks :+1: to me.

PR #816, opened 2015-04-14
Factor out people loading script, and use new data

dracos:


struan:
One teeny stylistic thing but otherwise looks good.

PR #817, opened 2015-04-15
Parameterize a few missing db queries.

dracos:
And escape a few variables.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -18,8 +18,8 @@ build_failure_conditions:
     # No new element rated worse than a D exists
     - 'elements.rating(<= D).new.exists'
     # No new major or higher severity issues exist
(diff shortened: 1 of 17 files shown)


struan:
Looks good.

jacksonj04:
I know I'm coming to this a bit late in the game, but is there a reason that the Scrutinizer fail conditions were removed as well as the problems causing the automated review to fail being fixed?

dracos:
As the commit message says, Scrutinizer was failing to filter out false positives. It listed 8 critical issues, all of which I checked and none of which actually was in any way, they were all ignored within the interface (and still are) and yet would continually be listed as present and thus causing github to show a failure when none was occurring. I am happy for Scrutinizer to be run and examined for possible improvements to the code, but I am not having pull requests continually marked as failures when they are not.

Issue #800, opened 2015-03-30
Make new streamlined MP alert sign up page


struan:
Very simplified enter postcode and signup button page for use in email campaigns encouraging people to sign up to their MP post election.

@zarino to do HTML layout and styling (using the same base template as the new MP and Debates pages) then @struan or @dracos to implement it in the TWFY code.

TomSteinberg:
The goal here is to have a 'sign up to your new MP for alerts' page that can be put in the email newsletter, into twitter and so on

TomSteinberg:
@MyfanwyNixon for interest and preparation - its in the grid now, maybe consider a newsletter for the day after the election?

MyfanwyNixon:
Nice!

jacksonj04:
Could someone ping me once this is almost ready to go, so I can get tracking set up?

dracos:
Design wise, this is a page that performs the same functionality as http://www.theyworkforyou.com/alert/ but solely asking for postcode (and email address if not logged in), and with presumably a lot less gumph, and also presumably in the new design. Technologically, it will then take you directly to be signed up/email confirmation, rather than the annoying intermediary step that http://www.theyworkforyou.com/alert/ does when given a postcode (and if this page can use the same code as http://www.theyworkforyou.com/alert/ and we can fix that at the same time, great). It would be good if the design could in some way be usable for the more general http://www.theyworkforyou.com/alert/ in future too.

zarino:
Pull request for the design work here: https://github.com/mysociety/theyworkforyou/pull/818

MyfanwyNixon:
Please also ping me when this is live so that I can put the link into the newsletter & blog post.

PR #818, opened 2015-04-15
Design for streamlined MP/postcode email signup

zarino:
This is the design aspect of #800. A developer needs to take these templates and add the backend logic to make them work.

I've kept the pages very minimal. There are some simple error handling styles for the form, and two different "success" states, which can be viewed by: submitting the form (to see the "check your emails" message) or appending `?confirmed` to the URL (to see what a user would see after clicking the confirmation link in their email).

The longer term vision is that this layout and styling will later be used for *all* of the `/alert` pages. By removing the `alerts-header--jumbo` class, and adding some more content below the header (like the "How your money helps" section on [mysociety.org/donate](https://www.mysociety.org/donate/), or the "How it works" section on [the Collideoscope homepage](http://collideosco.pe/)) we could encompass much of the functionality of the existing `/alert` pages, while hopefully making the information easier to scan.

**Screenshots:**

screen shot 2015-04-15 at 16 34 35

screen shot 2015-04-15 at 16 43 48

screen shot 2015-04-15 at 16 34 45

screen shot 2015-04-15 at 16 34 29

iphone5c_green_portrait


zarino:
PS. The background image of a person sitting with their back facing us is [Creative Commons Zero licensed from unsplash.com](https://unsplash.com/license), so no attribution is required.

man sitting with his back to us

www/docs/alert/by-postcode/index.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$new_style_template = TRUE;
+
+include_once '../../../includes/easyparliament/init.php';
+include_once INCLUDESPATH . 'easyparliament/member.php';
+
+global $THEUSER;
```
dracos: Don't need a global at the top level :)

www/includes/easyparliament/templates/html/alert/postcode.php
```diff
@@ -0,0 +1,84 @@
...
+            </div>
+
+          <?php else: ?>
+
+            <p class="lead">Enter your postcode, and we&rsquo;ll email you every time your MP speaks or receives a written answer.</p>
+
+          <?php if (isset($data['invalid-postcode-or-email'])): ?>
+            <div class="alerts-message alerts-message--error">
+                <h2>Oops!</h2>
+                <p>Please supply a valid UK postcode and email address.</p>
```
dracos: Could this wording be changed just so it puts the onus more on us for e.g. missing new build postcodes and doesn't annoy the user (e.g. "not a known postcode", "we didn't recognise the postcode" or similar, as it may be our fault).

dracos:
Nice :)

dracos:
(Should do some squashing)

PR #819, opened 2015-04-17
Load people data from JSON not XML.

dracos:


scripts/load-people
```diff
@@ -150,59 +135,64 @@ sub loadconstituency {
...
 
-    my $fromdate = $member->att('fromdate');
-    $fromdate .= '-00-00' if length($fromdate) == 4;
-    $fromdate = '0000-00-00' unless $fromdate;
-    my $todate = $member->att('todate');
-    $todate .= '-00-00' if length($todate) == 4;
+    my $start_date = $member->{start_date};
+    $start_date .= '-00-00' if length($start_date) == 4;
+    $start_date = '0000-00-00' unless $start_date;
+    my $end_date = $member->{end_date} || '9999-12-31';
```
struan: is there a reason not to follow this pattern for start_date?

struan:
Tiny stylistic thing aside this looks good.

scripts/load-people
dracos: I don't think there actually is anything without a start_date, so it's a bit odd; but yes, might as well :)

PR #828, opened 2015-04-20
Dot dev dns fix

zarino:
One for @jacksonj04.

No matter what we tried, `theyworkforyou.dev` wouldn't resolve on my Mac (OS X 10.10.3) so he suggested changing TWFY to not use `.dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ If you have questions, the best place to ask is the mySociety TheyWorkForYou ema
 
 You will need the latest versions of VirtualBox and Vagrant, then:
 
-* Stick an entry in your `hosts` file to point `theyworkforyou.dev` at `192.168.88.10` (Apache doesn't like IP addresses).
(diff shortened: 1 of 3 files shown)


dracos:
puphpet is currently broken with supervisor failing to start (looks to have been fixed 2 days ago). While I'm looking at this, is there a reason it is installing node? And does the puphpet config file format always change so quickly? Seems to make it hard to compare changes.

jacksonj04:
@dracos: Unfortunately yes, the PuPHPet config file tends to move quite quickly. It's a lot more stable than it was (which is saying something). It also includes a few random strings which change each time it's generated to ensure that any modifications are properly applied.

I can't see a reason for node to be there, given it's only installing grunt and bower (neither of which we use in a TWFY workflow).

dracos:
I've pushed this to master as 52b17f5.

Issue #455, opened 2014-03-06
New design: medium width page margins & related


dracos:
Fine at large and small, but there's a range within which the content buts right up against the browser edge.

dracos:
Just to add another between-816-and-1140 issue: the sidebar on a debates page moves underneath the speech content, but just not in as nice a way as <816.

dracos:
And it doesn't move underneath without potentially some overlap, e.g. http://www.theyworkforyou.com/pbc/2014-15/National_Health_Service_%28Amended_Duties_and_Powers%29_Bill/06-0_2015-03-03a.2.0
![screen shot 2015-04-21 at 09 43 41](https://cloud.githubusercontent.com/assets/154364/7248488/edcb785e-e80a-11e4-8af3-d314a5e34afc.png)

dracos:
Yay (the media width is my normal browser size :) ).

PR #829, opened 2015-04-20
Issues/745 spwrans meta

### LINKED ISSUE #745: Decide what to do about SP WRANS mentions
They either need a bit of design love or, as it's old data that's not going to be updated now possibly just removed.

See http://www.theyworkforyou.com/spwrans/?id=2011-03-17.S3W-40354.h for an example.

zarino:
~~(Sorry, this branch is on top of https://github.com/mysociety/theyworkforyou/pull/828 – but the `twfy.mysociety` change isn't actually required for the stuff this branch does. Can some Git genius separate it out?)~~

Fixes #745. Tidies up the display of written question metadata:

screen shot 2015-04-20 at 17 33 14

The `get_question_mentions_html()` function in `SpwransView.php` looked pretty hairy, so I tried not to touch it at all. So I've done what I can in CSS alone. It's less of an eyesore than it was before, but if we wanted to spend more time on it, I'd recommend drastically shortening the output of `get_question_mentions_html()`.

### DIFF
--- a/classes/SectionView/SpwransView.php
+++ b/classes/SectionView/SpwransView.php
@@ -123,7 +123,7 @@ class SpwransView extends WransView {
             if( $reference ) {
                 $references[] = "\n<li>$inner.";
             } else {
-                $result .= "\n<li class=\"link-to-hansard\">$description$inner.</span>";
+                $result .= "\n<li class=\"link-to-hansard\">$description$inner</span>";
                 $last_date = $row["date"];
             }
         }
--- a/www/docs/style/sass/pages/_section.scss
+++ b/www/docs/style/sass/pages/_section.scss
@@ -374,28 +374,42 @@ span.hi {
   color: lighten(#9d9195, 10%);
 }
 
-.link-to-hansard span {
-  display: inline;
+.link-to-hansard {
+  margin-bottom: 5px;
   font-size: 14px;
   color: lighten(#9d9195, 10%);
-  @media (min-width: $medium-screen) {
-    display: block;
+
+  span, .debate-speech__meta__link {
+    display: inline;
+
+    @media (min-width: $medium-screen) {
+      display: block;
+    }
+  }
+
+  .question-mention-gap {
+    display: inline; // This should always be inline, no matter what the width
   }
 }
 
 .debate-speech__links {
   margin: 0;
+
   li {
     display: block;
   }
+
   @media (min-width: $medium-screen) {
     padding-left: 70px;
   }
+
   @media (min-width: $large-screen) {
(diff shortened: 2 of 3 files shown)


zarino:
It's worth pointing out that the metadata `<ul>` is still absolutely positioned, so if you have a very short written question that's mentioned lots and lots of times, the list of mentions will extend beyond the bottom of the question wrapper and overlay on top of the next section of the page.

If you think this will be a common enough occurrence to warrant more investment to fix, let me know.

dracos:
Thanks :) I don't know if this is too tied in to #455 to fix for here, but at a 'medium' width (816-1140 on that ticket), the "Mentioned in" and "After 8 days, Given a" still appear on separate lines, but as the sidebar is now in flow with the main content, they could display all on one line, as they do at a smaller width?

"If you think this will be a common enough occurrence to warrant more investment to fix, let me know." – I haven't quickly found a written answer for which it happens, but it does happen for PBCs e.g. http://www.theyworkforyou.com/pbc/2014-15/National_Health_Service_%28Amended_Duties_and_Powers%29_Bill/06-0_2015-03-03a.2.0 so if it were a simple thing to fix, that would be good.

zarino:
> if it were a simple thing to fix, that would be good.

Hmmmm. We could wrap `.debate-speech__speaker` and `.debate-speech__content` in *another* `div` that’s floated left, and then set `.debate-speech__links` to `{ float: right }` on wide screens rather than setting it to `{ position: absolute }`. That way, we can clearfix the parent (`.debate-speech`) and no matter how long or short the speech or the meta links are, they'll stay contained in the row they're meant to.

> at a 'medium' width […] they could display all on one line, as they do at a smaller width?

Yes, that sounds possible.

zarino:
Turns out there's already been an attempt to fix the "context links overlapping into next speech" bug – https://github.com/mysociety/theyworkforyou/commit/485d363d3d3de6a76947db7711e6489bc0b8a74b – assuming this has since been deployed, looking at the page linked to in the original ticket, it doesn't seem to have helped.

dracos:
Was it meant to be min-height, not min-width? (though I think some might have a greater height than that, whilst it might leave gaps elsewhere)

zarino:
Yes, I think min-height was a likely candidate. Anyway, I'm almost finished with a version that uses floats instead, and it's looking promising so far. Will test it out a bit more locally, and then push for your delectation @dracos.

dracos:
Rebased after spotting typo in issue number in commit message.

dracos:
Merged and deployed; great to have fix for the overlapping too :)

PR #837, opened 2015-04-23
Improve load people

dracos:
The parlparse JSON no longer has sequential duplicates, so we don't need that slow code any more. Add a date so we don't have to consider old things if we don't want to, and store office ID to save deleting the entire table each run.


struan:
This looks good to me.

PR #840, opened 2015-04-24
Move names to separate database table

### LINKED ISSUE #825: Support names in separate table
Rather than one name per membership (other end of mysociety/parlparse#25).

dracos:
I've done this in batched commits by functionality.
The code assumes that a person ID will have only one name for a particular date, whereas the schema update copies all member names across, leading to duplicates. This only errors if you try and index or view a page with such a person on it (e.g. Gerry Adams). A script still needs to be written to consolidate overlapping names into one – most people will only need the one name anyway.
I'm not very happy with the load-people changes which slow it down again after I'd just sped it up; I can't think of a better way of doing it, though (even when the names are moved in the JSON to `persons` it'll be the same, though there will be fewer of them).
A few small random bugs are fixed along the way. Fixes #825.


struan:
Mostly looks ok. Some bugs in there and a few time where it fell over due to multiple names which is a know issue. ( Campbell in the search was the one I could identify )

On the other hand, manually changing a name for the previous Parliament did the right thing so woo!

dracos:
Merged as 509b38a57eb1f5086ea0bc5d06d708b265599805

PR #843, opened 2015-04-28
Implement "Where Next?" research tracking

jacksonj04:
* [x] Implement event logging for research actions.
* [x] Implement Magical JavaScript™ for performing aforementioned logging, as well as research bucket allocation.
* [x] Implement on Person pages.
* [x] Implement on Speech pages.

### What?

1. Randomly assigns a user visiting either a person or speech page to either a treatment or control group.
2. Records a page view.
3. Shows any necessary treatment.
4. Records any interactions with the treatment.

### DIFF
--- a/www/docs/js/qual2.js
+++ b/www/docs/js/qual2.js
@@ -76,8 +76,7 @@ $(document).ready(function() {
           'page': document.URL,
           'bucket': researchUser.bucket,
           'data': link.data('research-qual2-bucket1-linkname'),
-          'timer': timer,
-          'timeout': 300
+          'timer': timer
         },
         dataType: 'json'
       });
--- a/www/includes/easyparliament/templates/html/mp/profile.php
+++ b/www/includes/easyparliament/templates/html/mp/profile.php
@@ -307,17 +307,4 @@ you can find it.</p>
     </div>
 </div>
 
-<div id="whereNextModal" class="reveal-modal" data-reveal aria-labelledby="modalTitle" aria-hidden="true" role="dialog">
-    <h2 id="modalTitle">Where to next?</h2>
-    <p class="lead">Do you know what would you like to do with the information on this page?</p>
-    <ul>
-        <li><a id="research-qual2-bucket1-wherenext" href="/action/where-next">I&rsquo;m not sure &mdash; Click here for suggestions on what to do next</a></li>
-        <li><a data-research-qual2-bucket1-linkname="write-mp">I&rsquo;m going to write to an MP</a></li>
-        <li><a data-research-qual2-bucket1-linkname="social-share">I&rsquo;m going to share this information on social media</a></li>
-        <li><a data-research-qual2-bucket1-linkname="work">I&rsquo;m going to use this information for work purposes</a></li>
-        <li><a data-research-qual2-bucket1-linkname="browsing">I don&rsquo;t need to do anything with this information &mdash; I was just browsing</a></li>
-    </ul>
-</div>
-
-<script src="<?= cache_version('js/jquery.storageapi.min.js') ?>"></script>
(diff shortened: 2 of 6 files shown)


struan:
Should this not update schema.js for the new table?

jacksonj04:
It should, and that has now been fixed (along with including a migration).

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -313,3 +306,18 @@
...
+        <li><a id="research-qual2-bucket1-wherenext" href="/action/where-next">I&rsquo;m not sure &mdash; Click here for suggestions on what to do next</a></li>
+        <li><a data-research-qual2-bucket1-linkname="write-mp">I&rsquo;m going to write to an MP</a></li>
+        <li><a data-research-qual2-bucket1-linkname="social-share">I&rsquo;m going to share this information on social media</a></li>
+        <li><a data-research-qual2-bucket1-linkname="work">I&rsquo;m going to use this information for work purposes</a></li>
+        <li><a data-research-qual2-bucket1-linkname="browsing">I don&rsquo;t need to do anything with this information &mdash; I was just browsing</a></li>
+    </ul>
+</div>
+
+<script src="<?= cache_version('js/jquery.storageapi.min.js') ?>"></script>
+<script src="<?= cache_version('js/qual2.js') ?>"></script>
```
struan: Given this is repeated in two places could it not be pulled out into an include?

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -313,3 +306,18 @@
...
+<div id="whereNextModal" class="reveal-modal" data-reveal aria-labelledby="modalTitle" aria-hidden="true" role="dialog">
+    <h2 id="modalTitle">Where to next?</h2>
+    <p class="lead">Do you know what would you like to do with the information on this page?</p>
+    <ul>
+        <li><a id="research-qual2-bucket1-wherenext" href="/action/where-next">I&rsquo;m not sure &mdash; Click here for suggestions on what to do next</a></li>
+        <li><a data-research-qual2-bucket1-linkname="write-mp">I&rsquo;m going to write to an MP</a></li>
+        <li><a data-research-qual2-bucket1-linkname="social-share">I&rsquo;m going to share this information on social media</a></li>
+        <li><a data-research-qual2-bucket1-linkname="work">I&rsquo;m going to use this information for work purposes</a></li>
+        <li><a data-research-qual2-bucket1-linkname="browsing">I don&rsquo;t need to do anything with this information &mdash; I was just browsing</a></li>
+    </ul>
```
struan: There's no obvious close option here for people who don't want to participate. I realise you can click off the the popup to dismiss it but an explicit link to do so would be nice.

struan:
There's nothing on the popup to say who is collecting the data and what it's going to be used for which seems like something it should do. As it is my reaction is very "er, what is this, why am I being asked this?" which doesn't really fit very well with how we usually do things.

www/docs/js/qual2.js
```diff
@@ -0,0 +1,165 @@
...
+          'data': link.data('research-qual2-bucket1-linkname'),
+          'timer': timer,
+          'timeout': 300
+        },
+        dataType: 'json'
+      });
+
+      $('#whereNextModal').foundation('reveal', 'close');
+
+    });
```
struan: I wonder if these two .click functions could be factored out into one given only the data is different between them?

jacksonj04: Possible, but they also have different timeout behaviour and differ in how they execute the next action, so I'm inclined to say it's actually clearer to leave them as two separate bits of code since they conceptually do two different things (one reports back and closes the dialogue, one blocks the page being changed until the report is complete or has timed out).

PR #844, opened 2015-04-28
Basic styling for actions page elements

zarino:
Spent basically no time on this at all, but main differences are:

* Clearer distinction between the introduction and the options in the main body
* More legible line lengths
* White background for most of the text

screen shot 2015-04-28 at 15 33 02

Folds up nicely on narrow screens, as you'd expect.


jacksonj04:
👍🏼🎉

PR #846, opened 2015-04-30
Email users with alerts for now former MPs

struan:
Script and page to inform users who are signed up to alerts MPs who have lost/resigned their seat at the election and to easily create an alert for the new MP.

Built on top of the new streamlined alert page branch.


struan:
mmm, I've got something wrong setting this up, trying again

PR #848, opened 2015-05-01
New homepage design

struan:
Implements new homepage for UK, Scotland and NI parliaments


www/docs/admin/featured.php
```diff
@@ -0,0 +1,198 @@
...
+
+$this_page = 'admin_featured';
+
+$db = new ParlDB;
+
+$PAGE->page_start();
+$PAGE->stripe_start();
+
+$out = '';
+$preview = 0;
```
dracos: Is this variable used?

www/docs/admin/featured.php
```diff
@@ -0,0 +1,198 @@
...
+<?php
+}
+
+function get_gid_from_url($url) {
+    $gid = NULL;
+    $parts = parse_url($url);
+    parse_str($parts['query'], $query);
+
+    if ( $query['id'] ) {
+        $gid = "uk.org.publicwhip/debate/" . $query['id'];
```
dracos: Should the admin say somewhere it's commons debates only?

www/docs/admin/featured.php
```diff
@@ -0,0 +1,198 @@
...
+    if ( get_http_var('url') ) {
+        $url = get_http_var('url');
+        $title = get_http_var('title');
+        $context = get_http_var('context');
+        $related1 = get_http_var('related1');
+        $related2 = get_http_var('related2');
+        $related3 = get_http_var('related3');
+    } else {
+        $url = gid_to_url($gid);
+        $related1 = gid_to_url($related[0]);
```
dracos: Is the first related optional too? Looks like `gid_to_url` returns empty string if not present, so should the following two lines not bother with the `isset` either?

www/docs/admin/featured.php
```diff
@@ -0,0 +1,198 @@
...
+    $related2 = get_http_var('related_gid2');
+    $related3 = get_http_var('related_gid3');
+
+    $featured->set_title($title);
+    $featured->set_context($context);
+    $featured->set_gid($gid);
+    $featured->set_related(array($related1, $related2, $related3));
+
+    $out = "<h4>update successful</h4>";
+    $out .= "<p>Title set to $title and gid to $gid</p>";
```
dracos: HTML injection here, for what it's worth.

www/docs/admin/featured.php
```diff
@@ -0,0 +1,198 @@
+<style>
```
dracos: I'd move this style block down to before the confirm-panel it refers to.

classes/SectionView/NiView.php
```diff
@@ -20,59 +21,52 @@ protected function front_content() {
...
+        $urls['alert'] = $alert->generate();
+
+        $data['urls'] = $urls;
+
         $DEBATELIST = new \NILIST;
-        $DEBATELIST->display('recent_debates', array('days' => 30, 'num' => 5));
+
+        $random_debate = $DEBATELIST->display('recent_debates', array('days' => 30, 'num' => 1), 'none');
+
+        $debates = $DEBATELIST->display('recent_debates', array('days' => 30, 'num' => 5), 'none');
```
dracos: If you called this once with 6, and popped the first result as `$random_debate`, that would prevent the same thing showing up in both sections.

www/includes/easyparliament/templates/html/index.php
```diff
@@ -0,0 +1,162 @@
...
+                            <input type="submit" value="Find out &rarr;" class="button homepage-search__button" />
+                        </div>
+                    </form>
+                </div>
+                <?php } ?>
+            </div>
+        </div>
+        <div class="hero__site-intro">
+            <div class="hero__site-intro__wrap">
+                <h2>Democracy: it's for everyone</h2>
```
dracos: Could we have nice apostrophes everywhere? :)

www/includes/easyparliament/templates/html/index.php
```diff
@@ -0,0 +1,162 @@
...
+                <p>You shouldn't have to be an expert to understand what goes on in Parliament. Your politicians represent you&hellip; but what exactly do they do in your name?</p>
+                <p>TheyWorkForYou takes open data from the UK Parliament, and presents it in a way that's easy to follow &ndash; for everyone. So now you can check, with just a few clicks: are They Working For You?</p>
```
dracos: Ha ha ha :)

www/includes/easyparliament/templates/html/ni/index.php
```diff
@@ -0,0 +1,111 @@
+    <div class="hero">
+        <div class="row">
+        <div class="hero__mp-search">
+            <div class="hero__mp-search__wrap">
+                <h1>Do your MLAs represent you?</h1>
```
dracos: This presumably matches the old behaviour, but I see this is less fully functional than UK/Scotland in terms of showing you existing etc. – is that easier to fix now these things are more overlapping?

www/includes/easyparliament/templates/html/index.php
```diff
@@ -0,0 +1,162 @@
...
+                        <a class="homepage-search__change-postcode" href="<?= $mp_data['change_url'] ?>">Change postcode</a>
+                    </div>
+                </div>
+                <?php } else { ?>
+                <h1>Does your MP represent you?</h1>
+                <div class="row collapse">
+                    <form action="/postcode/" class="mp-search__form">
+                        <label for="postcode">Your postcode</label>
+                        <div class="medium-9 columns">
+                            <input id="postcode" class="homepage-search__input" type="text" placeholder="TF1 8GH" />
```
dracos: This is missing a name, so doesn't work.

www/includes/easyparliament/templates/html/index.php
```diff
@@ -0,0 +1,162 @@
+    <div class="hero">
+        <div class="row">
+        <div class="hero__mp-search">
+            <div class="hero__mp-search__wrap">
+                <?php if (count($mp_data)) { ?>
+                <h1>Does <?= $mp_data['name'] ?> represent you?</h1>
```
dracos: I note that the "former" was lost in 4eb267b5fdbf1dba3faa719dab40502fe91c03ca and so this question doesn't make much sense during an electoral period; something needs to be done when former is present.

www/includes/easyparliament/templates/html/index.php
```diff
@@ -0,0 +1,162 @@
...
+                </div>
+                <?php } ?>
+            </div>
+        </div>
+        <div class="hero__site-intro">
+            <div class="hero__site-intro__wrap">
+                <h2>Democracy: it's for everyone</h2>
+                <p>You shouldn't have to be an expert to understand what goes on in Parliament. Your politicians represent you&hellip; but what exactly do they do in your name?</p>
+                <p>TheyWorkForYou takes open data from the UK Parliament, and presents it in a way that's easy to follow &ndash; for everyone. So now you can check, with just a few clicks: are They Working For You?</p>
+                <a href="#" class="site-intro__more-link">Find out more about TheyWorkForYou <i>&rarr;</i></a>
```
dracos: Where is this link going?

www/includes/easyparliament/hansardlist.php
```diff
@@ -2757,6 +2757,126 @@ public function biggest_debates($args=array()) {
         return $this->_get_data_by_biggest_debates($args);
     }
 
+    public function _get_data_by_featured_gid($args=array()) {
```
dracos: Could this be factored in with recent debates, it seems to be quite similar to it?

dracos:
A few odds and sods, looks good :)

www/includes/easyparliament/templates/html/ni/index.php
struan: Well, the Scottish one only shows you your non list MSP. I guess we could list all of them but I think I'd want some design time for that. I might open a ticket about it.

dracos: http://www.theyworkforyou.com/postcode/?pc=eh11bb shows me list and non-list MSPs? Sorry if I've misunderstood.

www/includes/easyparliament/hansardlist.php
struan: I did think about that but there's enough differences that factoring it out would require lots of conditionals which I think would make the code harder rather than easier to read.

www/docs/admin/featured.php
```diff
@@ -0,0 +1,197 @@
...
+$out = '';
+if (get_http_var('preview')) {
+    preview_featured();
+} else if ( get_http_var('confirm') ) {
+    $out = update_featured();
+}
+
+print '<div id="adminbody">';
+print $out;
+print '<p><b>This currently only works for debates in the UK Parliament</b></p>';
```
dracos: As it uses `/debate/` I think it's only Commons debates, not Lords?

classes/SectionView/NiView.php
```diff
@@ -20,59 +21,69 @@ protected function front_content() {
...
+        $MOREURL->insert( array( 'more' => 1 ) );
+
+        // this makes sure that we don't repeat this debate in the list below
+        $random_debate = NULL;
+        if ( isset($debates['data']) && count($debates['data']) ) {
+            $random_debate = $debates['data'][0];
+        }
+
+        $recent = array();
+        if ( isset($debates['data']) && count($debates['data']) ) {
```
dracos: Should this check the count is high enough for the for loop coming? Or have that truncate to the count if smaller?

www/includes/postcode.inc
```diff
@@ -29,7 +29,7 @@ function postcode_to_constituencies($postcode, $api = false) {
     global $last_postcode, $last_postcode_value;
     $postcode = canonicalise_postcode($postcode);
     if ($last_postcode == $postcode) {
-        twfy_debug ("TIME", "Postcode $postcode looked up last time, is $last_postcode_value .");
+        twfy_debug ("TIME", "Postcode $postcode looked up last time, is " . $last_postcode_value['WMC']);
```
dracos: I think this is fixed on master to be better anyway, so won't be needed here.

classes/Homepage.php
```diff
@@ -0,0 +1,205 @@
...
+        if ($THEUSER->isloggedin() && $THEUSER->postcode() != '' || $THEUSER->postcode_is_set()) {
+            // User is logged in and has a postcode, or not logged in with a cookied postcode.
+
+            // (We don't allow the user to search for a postcode if they
+            // already have one set in their prefs.)
+
+            // this is for people who have e.g. an English postcode looking at the
+            // Scottish homepage
+            try {
+                $MEMBER = new Member(array('postcode'=>$THEUSER->postcode(), 'house'=> $this->mp_house));
```
dracos: As with the other PR, I think this won't work for Scotland page with a postcode in a place that has differing SP/UK constituency names.

dracos:
Thanks for all the fixes :) Unfortunately, this has the constituency name issue too; but apart from that looks fine.

classes/Homepage.php
```diff
@@ -0,0 +1,208 @@
...
+                return $mp_data;
+            }
+
+            if ($MEMBER->valid) {
+                if ($THEUSER->isloggedin()) {
+                    $CHANGEURL = new \URL('useredit');
+                } else {
+                    $CHANGEURL = new \URL('userchangepc');
+                }
+                $mp_data['name'] = $MEMBER->given_name() . ' ' . $MEMBER->family_name();
```
dracos: Better if this were MEMBER->full_name().

classes/Member.php
```diff
@@ -376,4 +376,54 @@ public function getEnterLeaveStrings() {
...
+                    array(
+                        ':house' => $house,
+                        ':cons_name' => $cons_name,
+                        ':dissolution_date' => $dissolution_dates[$house]
+                    )
+                );
+            }
+
+            for ($i = 0; $i < $q->rows; $i++) {
+                $mreg[] = $q->row($i);
```
dracos: To match e.g. people.php, if you could call member_full_name here with the right fields, and then just use name rather than given/family in the regional listing templates, to avoid the templates hopefully having to think about names.

www/includes/easyparliament/templates/html/ni/index.php
```diff
@@ -0,0 +1,120 @@
...
+                <div class="row collapse">
+                    <?php if ( count($data['regional']) > 0 ) { ?>
+                        <ul class="homepage-rep-list">
+                            <li>Your MLAs: </li>
+                        <?php foreach ( $data['regional'] as $mla ) { ?>
+                            <li class="homepage-rep-list__rep"><a href="/mla/?p=<?= $mla['person_id'] ?>"><?= $mla['first_name'] ?> <?= $mla['last_name'] ?></a></li>
```
dracos: Still first/last name (but note other comment that using $mla['name'] here instead would be nice).

classes/Member.php
```diff
@@ -376,4 +376,54 @@ public function getEnterLeaveStrings() {
...
+                AND entered_house >= start_date
+                AND left_house <= end_date
+                AND left_reason = 'still_in_office' AND house = :house",
+                array(
+                    ':house' => $house,
+                    ':cons_name' => $cons_name
+                )
+            );
+            if ( !$q->rows() ) {
+                $q = $db->query("SELECT member.person_id, first_name, last_name, constituency, house
```
dracos: Still using first/last name here.

www/includes/postcode.inc
```diff
@@ -17,7 +17,7 @@ function _postcode_to_constituencies($postcode, $mp_only) {
     $postcode = canonicalise_postcode($postcode);
 
     if ($last_postcode == $postcode) {
-        $return_value = $mp_only ? $last_postcode_value['WMC'] : $last_postcode_value;
+        $return_value = is_array($last_postcode_value) ? $last_postcode_value['WMC'] : $last_postcode_value;
```
dracos: last_postcode_value is always an array, so this won't be returning the right value when mp_only is false, I think? Not sure what the actual issue is here, sorry; is something calling this function with an incorrect second argument, perhaps?

dracos:
Sorry for the name-related issues :/

dracos:
Looks good :)

Issue #799, opened 2015-03-30
Send email to people who are signed up to alerts for an MP who has lost


struan:
Tell them that they won't get emails about them any more as they're not an MP.
Possibly include a "sign up to alerts on your new MP" link as we should know who that is.

PR #849, opened 2015-05-05
Script to send email to users with alerts for MPs who are no longer in Parliament

struan:
Send an email with a link to allow them to update the alert to the new MP in that constituency. For #799

Also a page to allow them to confirm changing the alert.

( NB: this is the diff to #818 as it uses the design from that )

### DIFF
--- a/classes/AlertView.php
+++ b/classes/AlertView.php
@@ -31,6 +31,12 @@ class AlertView {
 
             $result = $this->createAlertForPostCode($data['email'], $data['postcode']);
             $data = array_merge( $data, $result );
+        } elseif (get_http_var('update')) {
+            $result = $this->getNewMP(get_http_var('update'));
+            $data = array_merge( $data, $result );
+        } elseif (get_http_var('update-alert')) {
+            $success = $this->replaceAlert( get_http_var('confirmation') );
+            $data['confirmation_received'] = $success;
         } elseif (get_http_var('confirmed')) {
             $success = $this->confirmAlert( get_http_var('confirmed') );
             $data['confirmation_received'] = $success;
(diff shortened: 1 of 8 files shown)


www/includes/easyparliament/templates/html/alert/update-mp.php
```diff
@@ -0,0 +1,72 @@
...
+                <?php endif ?>
+            </div>
+
+          <?php elseif (isset($data['error'])): ?>
+            <div class="alerts-message alerts-message--error">
+                <h2>Something went wrong</h2>
+                <p>Sorry, we were unable to create this alert. Please <a href="mailto:<?= str_replace('@', '&#64;', CONTACTEMAIL) ?>">let us know</a>. Thanks.</p>
+            </div>
+
+          <?php else: ?>
```
dracos: Do we want this else condition, should it just redirect to the alert/by-postcode page?

www/docs/alert/update-mp/index.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$new_style_template = TRUE;
+
+include_once '../../../includes/easyparliament/init.php';
+include_once INCLUDESPATH . 'easyparliament/member.php';
+
+global $THEUSER;
```
dracos: Ditto other PR

scripts/alertmpchanged.php
```diff
@@ -0,0 +1,186 @@
...
+        $queries++;
+        $members[$person_id] = new MEMBER(array('person_id' => $person_id));
+    }
+    $member = $members[$person_id];
+
+    // if they're still elected then don't send the email
+    if ($member->current_member_anywhere()) continue;
+
+    // skip if they didn't lose their westminster seat in the most recent election
+    if ($member->left_house[1]['date'] != '2015-03-30') continue;
```
dracos: This could use DISSOLUTION_DATE for if it's still around in 5 years (or a few months, who knows).

classes/AlertView.php
```diff
@@ -123,6 +129,26 @@ private function confirmAlert($token) {
         return $this->alert->confirm($token);
     }
 
+    private function replaceAlert($confirmation) {
+        $existing = $this->alert->fetch_by_token($confirmation);
```
dracos: The alert script will, if I've understood it right, have found any alert containing `speaker:ID` (e.g. "horse speaker:ID"), not just exactly `speaker:ID`. As well as dealing with that, this script needs to replace all such alerts the user has, not just that particular one whose token it put in the email.

classes/AlertView.php
```diff
@@ -136,4 +162,47 @@ private function isEmailSignedUpForPostCode($email, $postcode) {
...
+
+        $existing = $this->alert->fetch_by_token($confirmation);
+        $criteria = str_replace('speaker:', '', $existing['criteria']);
+        $data = array();
+
+        $old_mp = new Member(array( 'person_id' => $criteria ) );
+        if ( $old_mp->current_member_anywhere() ) {
+            $data = array(
+                'old_mp' => $old_mp->full_name(),
+                'still_in_post' => True,
```
dracos: Not sure what `still_in_post` is for – is it needed, does the script not make sure this doesn't happen?

classes/AlertView.php
```diff
@@ -136,4 +162,47 @@ private function isEmailSignedUpForPostCode($email, $postcode) {
         }
         return $is_signed_up;
     }
+
+    private function getNewMP($confirmation) {
+        if (!$confirmation) {
```
dracos: Ditto the issue with replaceAlert

dracos:
Looks good, apart from I think the front end needs to deal with more alerts than the pure speaker ones.

struan:
Merged in d688e18dc2706e37eb584e84b60fdda4cdb72f87

PR #850, opened 2015-05-06
Add section to new MP's pages saying they are new

### LINKED ISSUE #703: Design and implement better page for brand new MPs with almost no history
"I’ve just looked at the Voting Summary for Liz McInnes, the recently elected Labour MP for Heywood and Middleton.
 
While the relevant page does make it clear that Liz McInnes was only elected on 10 Oct 2014, this information is rather buried in the text lower down the page.
 
Particularly for MP’s who enter Parliament through a by-election, would it not give a more balanced picture if their date of entry into Parliament was shown more prominently at the top of the voting Summary page?"

struan:
Note to say they were recently elected which is why there's not much
content on the page. Dissapears either after 6 months or if they've got
recent activity.

Similar for policy page.

Fixes #703

### DIFF
--- a/classes/Member.php
+++ b/classes/Member.php
@@ -65,6 +65,51 @@ class Member extends \MEMBER {
 
     }
 
+    /*
+     * Determine if the member is a new member of a house where new is
+     * within the last 6 months.
+     *
+     * @param int house - identifier for the house, defaults to 1 for westminster
+     *
+     * @return boolean
+     */
+
+    public function isNew($house = 1) {
+        $date_entered = $this->getEntryDate($house);
+
+        if ($date_entered) {
+            $date_entered = new \DateTime($date_entered);
+            $now = new \DateTime();
+
+            $diff = $date_entered->diff($now);
+            if ( $diff->y == 0 && $diff->m <= 6 ) {
+                return TRUE;
+            }
+        }
+
+        return FALSE;
+    }
+
+    /*
+     * Get the date the person first entered a house. May not be for their current seat.
+     *
+     * @param int house - identifier for the house, defaults to 1 for westminster
+     *
+     * @return string - blank if no entry date for that house otherwise in YYYY-MM-DD format
+     */
+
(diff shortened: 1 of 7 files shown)


classes/Member.php
```diff
@@ -65,6 +65,50 @@ public function isDead() {
...
+     *
+     * @param int house - identifier for the house, defaults to 1 for westminster
+     *
+     * @return string - blank if no entry date for that house otherwise in YYYY-MM-DD format
+     */
+
+    public function getEntryDate($house = 1) {
+        $date_entered = '';
+
+        $entered_house = $this->entered_house();
```
dracos: This could be `$this->entered_house($house)` which then returns either $entered_house[$house] or null straight off.

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -39,6 +39,16 @@
                     <div class="panel">
                         <p>Sinn F&eacute;in MPs do not take their seats in Parliament.</p>
                     </div>
+                    <?php elseif (isset($is_new_mp) && $is_new_mp && count($recent_appearances['appearances']) == 0): ?>
+                    <div class="panel--secondary">
+                        <h3><?= $full_name ?> is a recently elected MP - elected on <?= format_date($entry_date, LONGDATEFORMAT) ?></h3>
```
dracos: en dash please :)

classes/Member.php
```diff
@@ -65,6 +65,50 @@ public function isDead() {
...
+     * @param int house - identifier for the house, defaults to 1 for westminster
+     *
+     * @return boolean
+     */
+
+    public function isNew($house = 1) {
+        $entered_house = $this->entered_house();
+
+        if ( $entered_house && isset($entered_house[$house]) ) {
+            $date_entered = new \DateTime($entered_house[$house]['date']);
```
dracos: Can this call getEntryDate perhaps? :)

dracos:
Looks fine

PR #857, opened 2015-05-11
ignore case differences when loading constituencies

### LINKED ISSUE #856: load-names unsets main_name for some constituencies
If the same name appears twice then the second time main_name is unset and hence the constituency does not appear in search results.

struan:
Because mysql's default collation is case insensitive the replace into
is triggered by case differences in names becuase the index sees
Berwick-upon-Tweed and Berwick-Upon-Tweed as the same. This means the
for the second one we update main_name to 0 which means no results are
found by the constituency name search.

Fixes #856


dracos:
This looks fine; an alternative solution (which may be what I intended but clearly forgot to do) would be to say the names in constituencies.json are case-insensitive (looks like they always have been, really) and remove the ones that only differ in case - but could have this as well in case that was forgotten in future, of course.

Issue #593, opened 2014-07-18
Improve Email Alerts (high level ticket)


TomSteinberg:
This needs to cover:
- [ ] 1 - Improving the experience of unsubscribing - #737 
- [ ] 2 - Improving the experience of subscribing to an MP's updates from an MP page
- [ ] 3 - Improving the experience of subscribing for keyword alerts (get designers to suggest whole journey here)
- [x] 4 - Stop people subscribing to alerts based on postcodes actually being mentioned in parliament
- [x] 5 - Stop people subscribing to alerts based on when a person's name is mentioned in parliament
- [ ] 6 - Improve the experience of modifying a user's current email alerts (plural)
- [x] 7 - Improve the layout of the emails themselves #277 #251
- [ ] 8 - Improve the experience of signing people up to multiple keyword alerts #24

struan:
Related: #737

PR #863, opened 2015-05-18
detect signups for postcodes with a keyword and advise user

struan:
If someone is signing up for and alert that is of the form

    ZZ1 1AA elephants

then it's likely they actually want to sign up for the MP for that
postcode mentioning elephants so put in a notice to this effect and also
a button to enable them to subscribe to an alert for that.

part of #593

### DIFF
--- a/tests/AlertsPageTest.php
+++ b/tests/AlertsPageTest.php
@@ -1,0 +1,58 @@
+<?php
+
+/**
+ * Provides test methods to ensure pages contain what we want them to in various ways.
+ */
+class AlertsPageTest extends FetchPageTestCase
+{
+
+    /**
+     * Loads the member testing fixture.
+     */
+    public function getDataSet()
+    {
+        return $this->createMySQLXMLDataSet(dirname(__FILE__).'/_fixtures/alertspage.xml');
+    }
+
+    private function fetch_page($vars)
+    {
+        return $this->base_fetch_page('', $vars, 'www/docs/alert');
+    }
+
+    public function testFetchPage()
+    {
+        $page = $this->fetch_page( array() );
+        $this->assertContains('TheyWorkForYou Email Alerts', $page);
+    }
+
+    public function testKeywordOnly() {
+        $page = $this->fetch_page( array( 'alertsearch' => 'elephant') );
+        $this->assertContains('Mentions of [elephant]', $page);
+    }
+
+    public function testPostCodeOnly() {
+        $page = $this->fetch_page( array( 'alertsearch' => 'SE17 3HE') );
+        $this->assertContains('things by Mrs Test Current-MP', $page);
+    }
+
+    public function testPostCodeWithKeyWord()
+    {
+        $page = $this->fetch_page( array( 'alertsearch' => 'SE17 3HE elephant') );
+        $this->assertContains('You have used a postcode and something else', $page);
+        $this->assertContains('[elephant] by Mrs Test Current-MP', $page);
+        $this->assertNotContains('by your MSP', $page);
+    }
+
+    public function testScottishPostcodeWithKeyword() {
+        $page = $this->fetch_page( array( 'alertsearch' => 'PH6 2DB elephant') );
+        $this->assertContains('You have used a postcode and something else', $page);
+        $this->assertContains('[elephant] by your MP, Mr Test2 Current-MP', $page);
+        $this->assertContains('[elephant] by your MSP, Mrs Test Current-MSP', $page);
+    }
+
+    public function testPostcodeAndKeywordWithNoSittingMP() {
+        $page = $this->fetch_page( array( 'alertsearch' => 'OX1 4LF elephant') );
+        $this->assertContains('You have used a postcode and something else', $page);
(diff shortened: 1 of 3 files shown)


www/docs/alert/index.php
```diff
@@ -341,6 +341,38 @@ function display_search_form ( $alert, $details = array(), $errors = array() ) {
             echo '<em class="error">You have used a comma in your search term &ndash; are you sure this is what you want?
 You cannot sign up to multiple search terms using a comma &ndash; either use OR, or fill in this form multiple times.</em>';
         }
+        if ( preg_match('#([A-Z]{1,2}\d+ ?\d[A-Z]{2})#i', $alertsearch, $m) && strlen($alertsearch) > strlen($m[1]) && validate_postcode($m[1]) ) {
```
dracos: This won't cover all postcodes; could we alter `validate_postcode` to optionally allow within string as well as whole string as it currently is?

www/docs/alert/index.php
```diff
@@ -341,6 +341,38 @@ function display_search_form ( $alert, $details = array(), $errors = array() ) {
...
+            $scottish_text = '';
+            $mp_display_text = '';
+            if (postcode_is_scottish($m[1])) {
+                $mp_display_text = 'your MP, ';
+                $scottish_text = ' or MSP';
+            }
```
dracos: There's a postcode_is_ni function too :)

www/docs/alert/index.php
```diff
@@ -341,6 +341,38 @@ function display_search_form ( $alert, $details = array(), $errors = array() ) {
...
+            $tidy_alertsearch= $m[1] . " " . trim(str_replace("$m[1]", "", $alertsearch));
+            $alertsearch_display = str_replace("$m[1] ", "", $tidy_alertsearch);
+            $alertsearch = str_replace("$m[1]", "speaker:" . $MEMBER->person_id, $tidy_alertsearch);
+            echo "<li>";
+            echo $form_start . '<input type="hidden" name="keyword" value="' . _htmlspecialchars($alertsearch) . '">';
+            echo 'Mentions of [';
+            echo _htmlspecialchars($alertsearch_display) . '] by ' . $mp_display_text . $MEMBER->full_name();
+            echo ' <input type="submit" value="Subscribe"></form>';
+            if ( $scottish_text ) {
+                $MEMBER = new MEMBER(array('postcode'=>$m[1], 'house' => 4));
```
dracos: I think this won't work if there is one of MP/MSP but not both (e.g. during an election period of one of them), as one of the MEMBER calls will throw an exception.

dracos: Ah, actually it won't work even if both are present; the postcode lookup of MEMBER only appears to work for MPs (it only calls `postcode_to_constituency` which returns the MP constituency name, not the perhaps different MSP one, e.g. EH11BB East/Central). It looks like mp/index.php and postcode/index.php both do the same sort of thing (they have the same fix me comment also) for getting MSPs/MLAs for a postcode, and something different would have to be done here, I'm afraid.

www/docs/alert/index.php
dracos: Though of course you have multiple MLAs for a postcode, don't you. I guess it could sign up for any of them saying the thing?

dracos:
The MP part looks fine (some tests would be good); the MSP part will need reworking, or I guess this could just be MPs for now.

dracos:
Looks like this is missing the fixture file?

tests/AlertsPageTest.php
```diff
@@ -0,0 +1,58 @@
...
+        return $this->base_fetch_page('', $vars, 'www/docs/alert');
+    }
+
+    public function testFetchPage()
+    {
+        $page = $this->fetch_page( array() );
+        $this->assertContains('TheyWorkForYou Email Alerts', $page);
+    }
+
+    public function testPostCodeOnly() {
```
dracos: Think name of this function and next should be swapped :)

www/docs/alert/index.php
```diff
@@ -343,6 +343,45 @@ function display_search_form ( $alert, $details = array(), $errors = array() ) {
...
+                $alertsearch_display = str_replace("$m[1] ", "", $tidy_alertsearch);
+                $alertsearch = str_replace("$m[1]", "speaker:" . $MEMBER->person_id, $tidy_alertsearch);
+                echo "<li>";
+                echo $form_start . '<input type="hidden" name="keyword" value="' . _htmlspecialchars($alertsearch) . '">';
+                echo 'Mentions of [';
+                echo _htmlspecialchars($alertsearch_display) . '] by ' . $mp_display_text . $MEMBER->full_name();
+                echo ' <input type="submit" value="Subscribe"></form>';
+                if ( $scottish_text ) {
+                    $constituencies = postcode_to_constituencies($m[1]);
+                    if ( $constituencies['SPC'] ) {
```
dracos: An `isset` or similar, just in case it doesn't?

dracos:
Couple of tiny things, assuming tests pass when the fixture is included :)

PR #865, opened 2015-05-18
Add postcode search to /mp page if redirect fails

struan:
If someone arrives at /mp without either a session with a postcode or a query string then present them with a postcode lookup dialog and also a link to the appropriate list of representatives for the page type.

### DIFF
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -671,13 +671,37 @@ try {
     }
 
 } catch (MySociety\TheyWorkForYou\MemberException $e){
+    global $this_page;
+
+    switch($this_page) {
+    case 'mla':
+        $rep = 'MLA';
+        $SEARCHURL = '/postcode/';
+        $MPSURL = new \URL('mlas');
+        break;
+    case 'msp':
+        $rep = 'MSP';
+        $SEARCHURL = '/postcode/';
+        $MPSURL = new \URL('msps');
+        break;
+    case 'peer':
+        $rep = 'Lord';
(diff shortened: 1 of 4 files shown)


www/docs/mp/index.php
```diff
@@ -671,13 +671,32 @@
...
+        break;
+    case 'msp':
+        $rep = 'MSP';
+        $SEARCHURL = '/postcode/';
+        $MPSURL = new \URL('msps');
+        break;
+    default:
+        $rep = 'MP';
+        $SEARCHURL = new \URL('mp');
+        $SEARCHURL = $SEARCHURL->generate();
```
dracos: I think this could use `/postcode/` in all cases? Might as well stick with the single redirector unless going straight to MP does something different.

www/docs/mp/index.php
```diff
@@ -671,13 +671,32 @@
...
+        $MPSURL = new \URL('mps');
+    }
```
dracos: Could this handle `peer` too? (Though not show postcode form then, of course.)

dracos:
Looks good.

www/docs/mp/index.php
struan: If you use /mp/ it redirects straight to the MP page whereas postcode shows you all the representatives for that postcode so /mp/ saves a click for the user.

dracos: Aha, makes sense.

dracos:
Great!

PR #866, opened 2015-05-18
Standardise page gutters on "new" pages

### LINKED ISSUE #695: Page-edge spacing in new design is broken and needs reviewing
All the outside gutters seem to be implemented in different ways across the new pages, and as a result is inconsistent or just broken in many places. It needs reviewing, this won't be a fun ticket.

zarino:
Fixes #695.

MP pages and debates pages were using a variety of different methods to set columns and gutters, especially on full-width elements, and each of these methods resulted in a lack of gutters at one breakpoint or another.

This PR standardises all the "new style" pages (MP pages, debates/wrans pages, topic pages) to use the `.full-page__unit` convention, and to avoid nested grid columns. The `/alert/by-postcode` page and [new homepage](https://github.com/mysociety/theyworkforyou/tree/issues/589-new-homepage-design) already use `.full-page__unit` where necessary, so didn't need any fixes.

The template changes *look* more drastic than they really are, because a lot of them are indentation changes as I've added or removed wrapping elements, to keep the same structure across all the pages.

I've kept the commits separate rather than squashing as the changes were pretty far-reaching and I'm worried, if they're compressed into a single change, Future Me will have no idea what happened. But if still want them squashed, @dracos, let me know.

From a design perspective, this is ready to go, but I'd appreciate a :+1: from a developer /cc @dracos @struan before merging.

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/profile.php
+++ b/www/includes/easyparliament/templates/html/mp/profile.php
@@ -6,312 +6,312 @@ include_once INCLUDESPATH . "easyparliament/templates/html/mp/header.php";
     <div class="full-page__row">
         <div class="full-page__unit">
             <?php if (count($policyPositions->positions) > 0): ?>
-            <div class="person-navigation page-content__row">
+            <div class="person-navigation">
                 <ul>
                     <li class="active"><a href="<?= $member_url ?>">Overview</a></li>
                     <li><a href="<?= $member_url ?>/votes">Voting Record</a></li>
                 </ul>
             </div>
             <?php endif; ?>
-            <div class="person-panels page-content__row">
-                <div class="sidebar__unit in-page-nav">
-                    <ul data-magellan-expedition="fixed">
-                      <?php if (count($policyPositions->positions) > 0): ?>
-                        <li data-magellan-arrival="votes"><a href="#votes">Votes</a></li>
-                      <?php endif; ?>
-                      <?php if (count($recent_appearances['appearances'])): ?>
-                        <li data-magellan-arrival="appearances"><a href="#appearances">Appearances</a></li>
-                      <?php endif; ?>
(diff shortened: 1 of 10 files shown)


www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -5,313 +5,313 @@
 <div class="full-page">
     <div class="full-page__row">
         <div class="full-page__unit">
-            <?php if (count($policyPositions->positions) > 0): ?>
-            <div class="person-navigation page-content__row">
+            <?php if (true || count($policyPositions->positions) > 0): ?>
```
dracos: Should this `true` still be present?

zarino: Oh! No! Good catch!

dracos:
"The template changes look more drastic than they really are" – note you can view the diff ignoring whitespace changes by appending `?w=1` to any GitHub URI, so for this PR e.g. https://github.com/mysociety/theyworkforyou/pull/866/files?w=1 (I don't think I'd have spotted that `true` without that ;) ).

One question, more for my understanding, than any issue – in fa069ae3e9de4a4f042b310af96680ed5a1c5492, why did you leave `topic-header__content` present? It isn't doing anything and its children are setting `grid-column` (so grandchildren of `full-page__row`), whereas on e.g. MP page it has full-page__unit setting grid-column directly under `full-page__row` with a `person-header__content` child. Should it be an extra class on the same div as `full-page__row`?

If you'd prefer the commits not to be squashed, then let's not squash the commits :) Whatever you think will be best for future looking back, as you say, is the best thing to do.

zarino:
> why did you leave `topic-header__content` present?

I just didn't notice that it wasn't being used by anything else! I've just pushed an amendment to that first commit, which also removes that div. You're right – there's no point having it in there if it's not performing a function.

Issue #838, opened 2015-04-24
Move people list pages to new design


struan:
/mps, /peers etc

zarino:
At the risk of being unoriginal, this could very easily adopt almost the exact same styling as the [new Search page](https://github.com/mysociety/theyworkforyou/pull/868), just with downloads listed in the sidebar, and a table of MPs in the main content area.

![screen shot 2015-05-19 at 16 57 30](https://cloud.githubusercontent.com/assets/739624/7707536/3174c83c-fe48-11e4-929d-50e167c8fede.png)

I would make *just one feature request* – that the "Find your MP" search box at the top performs a search both by postcode and MP name. (This is currently handled by two different search inputs on the existing page).

dracos:
The search by name box already accepts a postcode, so that's just copy change and reusing that form rather than the postcode one; and perhaps (given the postcode lookup redirects immediately, whilst name doesn't) having the search results page automatically redirect when given a postcode input (I don't see why it shouldn't)?

zarino:
@dracos: Ace! I tried a name in the postcode box, and it didn't work. I didn't think to try the other way round!

zarino:
Decided to timebox this to two hours, just to see how far I could get. The answer is: pretty far. Pull request here: https://github.com/mysociety/theyworkforyou/pull/870

![1](https://cloud.githubusercontent.com/assets/739624/7730546/efa63032-ff12-11e4-92fe-3ae85c56a74a.png)

![2](https://cloud.githubusercontent.com/assets/739624/7730547/f20b98e4-ff12-11e4-8507-a80837a0fffe.png)

PR #870, opened 2015-05-20
Example layouts for both Search and MP/MSP list pages

zarino:
The MP pages from #838 were so similar to the Search pages that I figured it'd be a quick job to repurpose the same layout. So here it is.

Take a look at http://twfy.mysociety/mps/new and http://twfy.mysociety/mps/new/with-postcode

I'm not 100% percent happy with the mess in the sidebar, but it's a good start for only one or two hours’ work.

Connects to #838


wrightmartin:
Looking great!

You could give the sidebar survey similar styling to the secondary box at the bottom of the MP pages? 
screen shot 2015-05-20 at 17 21 11

Some other feedback - I think there's a lack of contrast in the MP list that's making it feel busy. Have you tried setting the constituency and party type in a lighter shade? You could probably get away with a touch more vertical whitespace between the results too

I can't tell if the sorting controls should come before or after the A-Z pagination, it might be right as it is, but I'm not 100% sure

zarino:
Maybe we could clean up that sidebar by not using a `form`, and instead having something less obtrusive, like a link:

toggle

zarino:
@struan: I'm happy with how this looks now, after the tweaks in https://github.com/mysociety/theyworkforyou/commit/4c8b0eee923dcc7d9effe089bb809b82217d45c2.

As ever, let me know if there are any unhandled edge-cases that need styling, once you've implemented it.

struan:
Connects to #838

struan:
Closing in favour of #932 due to merge awfulness with this.

PR #872, opened 2015-05-22
Improvements to Quant2 recording

jacksonj04:
* Moves recording to use central Logbook, along with other sites.
* Adds outbound link recording to "where next" page.
* Fixes "qual2/quant2" typo.

**NB.** Removing the `record.php` event collector and the DB table will happen in a subsequent tidyup, as some end users may still be recording events to there for a few minutes following deployment.


dracos:
Looks okay to me.

PR #878, opened 2015-05-27
Add submission tracking to homepage postcode form

jacksonj04:
Adds tracking of submit events to the postcode form on the home page (similar to the search form).


struan:
You'll need to add this to the front pages for NI and Scotland too I think.

scotland/index.php and ni/index.php in the template directory.

struan:
Otherwise looks fine.

PR #881, opened 2015-05-27
Improve URL/search name lookup.

dracos:
Use any name, not just most recent per membership; throw an exception for multiple results; refactor mp/index.php to be tidier. https://github.com/mysociety/theyworkforyou/pull/881/files?w=1


struan:
Aside from minor SQL thing this looks good.

dracos:
Merged as 91d121c9823fe3dee81b76c15aca2fb4285d381a

PR #883, opened 2015-05-28
Fix changing name on search results

dracos:
Fix e.g. http://www.theyworkforyou.com/search/?s=speaker:25177&p=2&pop=1 (which should switch to "Bishop of Durham" halfway down.


struan:
Looks good.

Issue #871, opened 2015-05-21
New front page interacts badly with User menu


struan:
The big green panel at the right of the screen and the green user menu mean that the latter is very difficult to see on the new front page:

![screen shot 2015-05-21 at 17 30 53](https://cloud.githubusercontent.com/assets/5310/7753747/7415dbe8-ffdf-11e4-9e02-e3a1eb72d91c.png)

zarino:
Pull request here: https://github.com/mysociety/theyworkforyou/pull/891

dracos:
New PR #895

struan:
Fixed in 8c494d096b6d7c294896e164d7aca16fb2011c8d

PR #895, opened 2015-06-12
Improved nav bar for new-style pages

### LINKED ISSUE #711: Improve navigation bar on new style pages
The widescreen version is on the designs for #589:

398b7128-66a0-11e4-9937-d4f97b5c6e9a

And the mobile version:

screen shot 2014-11-06 at 16 26 25

zarino:
Fixes #711 and #871.

Adds some hierarchy to the nav drawer on mobiles. Greatly improves nav bar display at medium sizes, and improves dropdown menu styling and behaviour on wide screens.

Also upgrades TWFY to use the new grey mySociety header bar in place of the old grey tab.

Degrades sensibly in IE7 and IE8 (which get the mobile menu styling, to save repeating all of our desktop styling in a special stylesheet just for them). In fact, because of this, the new header actually avoids a number of layout bugs suffered by the current version on the live site.

The dropdown menus all work when JavaScript is unavailable too. Which they didn't before.

Here's a selection of screenshots:

narrow with menu open

medium screen with go to menu open

wide screen with assembly menu open

wide screen with search active

wide screen with user menu open


www/includes/easyparliament/templates/html/header.php
```diff
@@ -131,63 +131,70 @@ function trackLinkClick(link, category, name, value) {
...
+                        </a>
+                    </li>
                 </ul>
             </nav>
+
         </div>
     </header>
-    <?php if ( isset($banner_text) ) { ?>
+
+    <?php if ( false && isset($banner_text) ) { ?>
```
dracos: I guess this `false` is why the test fails :)

struan:
Looks good to me.

Issue #910, opened 2015-07-01
Display issues where MP differs from party on profile page


struan:
Use the work from #905 to replace the votes panel on the MP profile page with a list of issues where the MP has differed from their party. If there are no issues where they've differed then display the current list of randomly selected issues.

MyfanwyNixon:
I am just wondering why the wording is ordered as it is, eg:

> Labour's anti-terrorism laws. Stephen Phillips consistently voted for. Most Conservative MPs generally voted against. 

Is there any reason this is not worded:

> Stephen Phillips consistently voted for Labour's anti-terrorism laws, while most Conservative MPs generally voted against. 

To my mind that is more natural to read.

MyfanwyNixon:
Also @struan, if this is a feasible change please let me know, and I'll hold off from blogging until it's in place. :)

struan:
Text updated in 6543a8031128c2bde54dc2e11a14c37858bf1b2f.

PR #911, opened 2015-07-02
Display any votes where MP differs from party on profile page

struan:
This imports the policy vote from PublicWhip and then uses all the votes of current party members to calculate the party's position on a policy. It then compares that with an MP's position and if it is in opposition then it displays that policy on the MP's profile page.

Connects to #910

### DIFF
--- a/classes/Party.php
+++ b/classes/Party.php
@@ -1,0 +1,176 @@
+<?php
+/**
+ * Party Class
+ *
+ * @package TheyWorkForYou
+ */
+
+namespace MySociety\TheyWorkForYou;
+
+/**
+ * Party
+ */
+
+class Party {
+
+    public $name;
+
+    private $db;
+
+    public function __construct($name) {
+        $this->name = $name;
+        $this->db = new \ParlDB;
+    }
+
+    public function calculateAllPolicyPositions($policies) {
+        $positions = array();
+
+        foreach ( $policies->getPolicies() as $policy_id => $policy_text ) {
+            list( $position, $score ) = $this->calculate_policy_position($policy_id, true);
+            if ( $position === null ) {
+                continue;
+            }
+
+            $positions[$policy_id] = array(
+                'policy_id' => $policy_id,
+                'position' => $position,
+                'score' => $score,
+                'desc' => $policy_text
+            );
+        }
+
+        return $positions;
+    }
+
+    public function calculate_policy_position($policy_id, $want_score = false) {
+
+        // This could be done as a join but it turns out to be
+        // much slower to do that
+        $divisions = $this->db->query(
+            "SELECT division_id, policy_vote, division_date
+            FROM policydivisions
+            WHERE policy_id = :policy_id
+            AND house = 'commons'",
+            array(
+                ':policy_id' => $policy_id
+            )
+        );
+
+        $score = 0;
+        $max_score = 0;
+        $num_divs = $divisions->rows;
+        $total_votes = 0;
+
+        for ( $i = 0; $i < $num_divs; $i++ ) {
+            $division_id = $divisions->field($i, 'division_id');
+            $weights = $this->get_vote_scores($divisions->field($i, 'policy_vote'));
+
+            $votes = $this->db->query(
+                "SELECT count(*) as num_votes, vote
+                FROM persondivisionvotes
+                JOIN member ON persondivisionvotes.person_id = member.person_id
+                WHERE
+                    party = :party
+                    AND member.house = 1
+                    AND division_id = :division_id
+                    AND left_house = '9999-12-31'
+                GROUP BY vote",
+                array(
+                    ':party' => $this->name,
+                    ':division_id' => $division_id
+                )
+            );
+
+            $num_votes = 0;
+            for ( $j = 0; $j < $votes->rows(); $j++ ) {
+                $vote_dir = $votes->field($j, 'vote');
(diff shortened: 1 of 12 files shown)


classes/Party.php
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        return $scores;
+    }
+
+    public static function getParties() {
+        $db = new \ParlDB;
+
+        $party_list = $db->query(
+            "SELECT DISTINCT party FROM member"
```
dracos: Ignore empty party?

scripts/generate_party_positions.php
```diff
@@ -0,0 +1,26 @@
...
+
+include_once dirname(__FILE__) . '/../www/includes/easyparliament/init.php';
+
+$parties = MySociety\TheyWorkForYou\Party::getParties();
+$policies = new MySociety\TheyWorkForYou\Policies;
+
+$party_count = 0;
+foreach ( $parties as $party ) {
+    if ( !$party ) {
+        continue;
```
dracos: I think this check should be in getParties. However, I think this check is needed for: Independents, potentially Crossbench, and I guess it needs to treat Labour+Labour/Co-operative as one.

classes/Party.php
```diff
@@ -0,0 +1,218 @@
...
+                    ':party' => $this->name,
+                    ':division_id' => $division_id
+                )
+            );
+
+            $num_votes = 0;
+            for ( $j = 0; $j < $votes->rows(); $j++ ) {
+                $vote_dir = $votes->field($j, 'vote');
+                if ( $vote_dir == 'tellno' ) $vote_dir = 'no';
+                if ( $vote_dir == 'tellaye' ) $vote_dir = 'aye';
```
dracos: Are we including teller votes in someone's summary of their position? Presumably we are (and always have?), just wondering about votes such as http://www.publicwhip.org.uk/division.php?date=2015-07-07&number=41 ...

classes/Party.php
```diff
@@ -0,0 +1,218 @@
...
+
+            $votes = $this->db->query(
+                "SELECT count(*) as num_votes, vote
+                FROM persondivisionvotes
+                JOIN member ON persondivisionvotes.person_id = member.person_id
+                WHERE
+                    party = :party
+                    AND member.house = 1
+                    AND division_id = :division_id
+                    AND left_house = '9999-12-31'
```
dracos: Just wondering if this fix to only current membership causes any issues.

classes/Party.php
```diff
@@ -0,0 +1,218 @@
...
+
+            $total_votes += $num_votes;
+            $max_score += $num_votes * max( array_values( $weights ) );
+        }
+
+        if ( $total_votes == 0 ) {
+            return null;
+        }
+
+        if ( $max_score == 0 ) {
```
dracos: If this were true, I guess score would have to be 0 too, then weight would be 1, which doesn't sound right? Is this always caught by the above total_votes == 0?

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -55,11 +55,39 @@
...
+                        </p>
+
+                        <p>
+                        However, <?= $full_name ?> sometimes <b>differs</b> from their party collegues, such as:
+                        </p>
+
+                        <ul class="vote-descriptions">
+                          <?php foreach ($sorted_diffs as $policy_id => $score): ?>
+                            <li>
+                                <?= ucfirst(strip_tags($policies[$policy_id])) ?>. <?= $full_name ?> voted <b><?= $positions[$policy_id]['voted'] ?></b>. Most <?= $party ?> MPs voted <b><?= $party_positions[$policy_id]['position'] ?></b>.
```
dracos: Just to note #913 changes contents of positions and will affect this.

dracos:
Some questions, but not really on the code which is fine :)

struan:
`$motion->direction` is the result of the vote ( Majority indicates the vote went with the policy and minority against ) and `$motion->motion->policy_vote` is which direction an MP would vote if they agreed with the policy.

classes/Party.php
struan: Yeah, teller votes are included which will give some odd results but I figure we have to do the calculations the same for it to be a valid comparison.

classes/Party.php
struan: It does a little in that you're not comparing someone with how other members of their party voted at the time but I've taken the view that it's more useful to say how the MP compares to their current colleagues.

My hypothetical justification for this is if at an election the new intake causes a party to shift position on something and your MP doesn't agree with that shift then that's more useful to know than the fact that they voted in line with the previous position of the party.

dracos: Sounds good, as long as it's explained somewhere.

www/docs/mp/index.php
```diff
@@ -464,6 +464,22 @@
         // Generate limited voting record list
         $data['policyPositions'] = new MySociety\TheyWorkForYou\PolicyPositions($policies, $MEMBER, 6);
 
+        // generate party policy diffs
+        $party = new MySociety\TheyWorkForYou\Party($MEMBER->party());
+        $positions = new MySociety\TheyWorkForYou\PolicyPositions( $policiesList, $MEMBER );
+        if ( $party ) {
```
dracos: Is this ever false?

dracos:
One tiny query, otherwise :+1:

Issue #558, opened 2014-06-30
Explain use of "strongly against" etc. wording used for policy positions


RichardTaylor:
I don't think there's any published explanation (unless you count the TheyWorkForYou code)  of how wording such as "very strongly against" or " strongly against" etc. is selected for use when describing individuals' voting records on TheyWorkForYou. 

This omission came up via usertesting.

The answer isn't straightforward; here's my go at an explanation:

>Each statement such as "Voted moderately for more EU integration" on TheyWorkForYou is a summary of an individual's record based on a number of votes. 

>Each vote which provides an opportunity to vote for or against a policy such as "more EU integration" is assigned a number of points. Important votes are assigned 50 points. Less important votes are assigned 10 points if the individual in question took part in the vote and 2 points if they did not. 

>The total number of points allocated to each vote included within a policy is the number of points the individual would have been allocated if they had voted in-line with the policy at every vote included in it. 

>In important votes an individual gets awarded the full 50 points for voting in line with the policy, no points for voting against the policy, and 25 points for not voting. In less important votes, the MP gets 10 points for voting with the policy, no points for voting against, and 1 if absent. 

>An individual scoring over 95% possible points is described as voting "very strongly for" the policy, and one scoring less than 5% of the possible points is described as voting "very strongly against" the policy. 

>The full table of thresholds for using the different wordings to describe an individual's position in relation to a policy position is as follows:

 
| % Possible Points      | Wording  |
| ------------- |:-------------:|
|  < 5      | very strongly against | 
| 5 - < 15     | strongly against     |  
| 15 - <  40  | moderately against      | 
| 40 - <  60      | a mixture of for and against     |  
| 60  - <  85      | moderately for     |  
| 85 - < 95      | strongly for     |  
|  > 95      | very strongly for     |  

I don't know if the following is clearer or more cluttered:

| % Possible Points (p)      | Wording  |
| ------------- |:-------------:|
| 0 ≤ p < 5      | very strongly against | 
| 5 ≤ p < 15     | strongly against     |  
| 15 ≤ p <  40  | moderately against      | 
| 40 ≤ p <  60      | a mixture of for and against     |  
| 60 ≤ p <  85      | moderately for     |  
| 85 ≤ p < 95      | strongly for     |  
| 95 ≤ p ≤ 100      | very strongly for     |  

---

The above detail, or similar, could go on a help page. The key information to present next to the statements (along with a link to the full explanation) might be something like:

[...]

PR #913, opened 2015-07-03
Improvements to vote summary wording and buttons

zarino:
Connects to #558.

* "Generally", "Almost always", and "Consistently" rather than "Moderately", "Strongly", "Very strongly".
* Explanation of how votes are summarised on topic vote summary page.
* Chunkier buttons next to vote summaries, also reworded from "Details" to "Show votes".

screen shot 2015-07-03 at 09 56 01

### DIFF
--- a/classes/PolicyPositions.php
+++ b/classes/PolicyPositions.php
@@ -108,11 +108,13 @@ class PolicyPositions {
 
                 // Make sure the dream actually exists
                 if (!empty($dream_info)) {
-                    $this->positions[] = array( 'policy_id' => $policy[0], 'desc' => $dream_info['full_sentence'] );
+                    $this->positions[] = array(
+                        'policy_id' => $policy[0],
+                        'desc' => $dream_info['full_sentence']
+                    );
                     $this->positionsById[$policy[0]] = array(
                         'policy_id' => $policy[0],
-                        'desc' => $dream_info['full_sentence'],
-                        'voted' => $dream_info['voted']
+                        'desc' => $dream_info['full_sentence']
                     );
                     $i++;
                 }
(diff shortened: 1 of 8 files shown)


www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -86,9 +61,11 @@
                             <?php if ( isset($policy['position']) ) { ?>
                                 <div class="panel">
                                     <h3 class="policy-vote-overall-stance">
-                                        <?= $full_name ?> <?= $policy['position']['voted'] == 'never voted' ? $policy['position']['voted'] . ' on' : 'voted ' .$policy['position']['voted'] ?> <?= $policy['desc'] ?>
+                                        <?= $full_name . ' ' . $policy['position']['full_description'] . ' ' . $policy['desc'] ?>
                                     </h3>
 
+                                    <p>TheyWorkForYou has automatically calculated this MP&rsquo;s stance based on all of their votes on the topic. <a href="http://www.publicwhip.org.uk/mp.php?mpid=<?= $member_id ?>&dmp=<?= $policy['policy_id'] ?>">You can browse the source data on PublicWhip.org.uk</a>.</p>
```
dracos: Unescaped ampersand.

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -86,9 +61,11 @@
                             <?php if ( isset($policy['position']) ) { ?>
                                 <div class="panel">
                                     <h3 class="policy-vote-overall-stance">
-                                        <?= $full_name ?> <?= $policy['position']['voted'] == 'never voted' ? $policy['position']['voted'] . ' on' : 'voted ' .$policy['position']['voted'] ?> <?= $policy['desc'] ?>
+                                        <?= $full_name . ' ' . $policy['position']['full_description'] . ' ' . $policy['desc'] ?>
```
dracos: full_description is undefined anywhere, presumably meant to be full_sentence?

zarino: Yes!

www/includes/utility.php
```diff
@@ -1083,19 +1083,19 @@ function _major_summary_title($major, $data, $LISTURL, $daytext) {
 function score_to_strongly($dmpscore) {
     $dmpdesc = "unknown about";
     if ($dmpscore > 0.95 && $dmpscore <= 1.0)
-        $dmpdesc = "very strongly against";
+        $dmpdesc = "consistently voted against";
     elseif ($dmpscore > 0.85)
-        $dmpdesc = "strongly against";
+        $dmpdesc = "almost always voted against";
```
dracos: Interestingly, I personally consider "almost always" stronger than "consistently voted" (which to me means most of the time, but less than almost always would imply).

classes/PolicyPositions.php
```diff
@@ -111,8 +111,7 @@ private function getMemberPolicyPositions ($limit = NULL) {
                     $this->positions[] = array( 'policy_id' => $policy[0], 'desc' => $dream_info['full_sentence'] );
                     $this->positionsById[$policy[0]] = array(
                         'policy_id' => $policy[0],
-                        'desc' => $dream_info['full_sentence'],
-                        'voted' => $dream_info['voted']
+                        'desc' => $dream_info['full_sentence']
```
dracos: Probably not wanting to be done here, but just to note I'm not sure why here we have both `$positions` and `$positionsById` which (after this PS) contain exactly the same information.

www/includes/easyparliament/templates/html/mp/divisions.php
dracos: And could this be line-wrapped, so it isn't more luck that I spot that ampersand? :)

dracos:
Hopefully the test failure is just due to the variable misname. One query about the wording, as personally I'm not sure "consistently voted" gives the vibe of "always", even though I do realise that is the literal meaning of consistently &lt;shrug>.

zarino:
@dracos: Consistently / almost always / generally is my and @MyfanwyNixon’s best guess at new wording for now. I think we're hoping to try it and see whether user questions increase or decrease.

zarino:
One of the tests is failing because it checks for `voted strongly against introducing <b>foundation` on the summary page: https://travis-ci.org/mysociety/theyworkforyou/jobs/70755810#L931

zarino:
@dracos – Tests pass now.

dracos:
Thanks, looks good :-) (In general, it's preferable to leave squashing until after the final review of any changes, as that makes it easier for the reviewer to only review the things changed to fix their comments, and any other additions along the way.)

PR #916, opened 2015-07-08
Redirect gid for featured debate if required

### LINKED ISSUE #873: Front page featured debate sometimes blank
I think it's not picking up GID redirects.

struan:
Make sure that if the gid for the featured debate has been redirected we
follow it rather than displaying nothing.

Fixes #873


dracos:
Looks fine, but doesn't it need to do the same thing for the `related_gid`s too?

PR #917, opened 2015-07-08
swap search term and glossary highlighting order

### LINKED ISSUE #912: Class="glossary" html gone awry
Eg look at "maiden speech": http://www.theyworkforyou.com/debates/?id=2015-07-02a.1623.1&s=%22maiden+speech%22#g1623.5

struan:
Because of the way the search engine highlighting works it will add
highlighting to text that is inside tags, e.g. in the title tag of a
glossary link. The glossary highlighting doesn't do this so swap the
order we do these in.

Fixes #912

### DIFF
--- a/classes/SectionView/SectionView.php
+++ b/classes/SectionView/SectionView.php
@@ -159,6 +159,12 @@ class SectionView {
             $bodies[] = $body;
         }
 
+        if ($SEARCHENGINE) {
+            // We have some search terms to highlight.
+            twfy_debug_timestamp('Before highlight');
+            $bodies = $SEARCHENGINE->highlight($bodies);
+            twfy_debug_timestamp('After highlight');
+        }
         // Do all this unless the glossary is turned off in the URL
         if (get_http_var('ug') != 1) {
             // And glossary phrases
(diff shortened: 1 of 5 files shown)


dracos:
Highlighting things in tags I thought had been fixed by a299e717d8ee951de9dd8d8195ab08b905f94477 (which even included tests!), but looks like it didn't handle the case of a search *phrase* which is used here. This needs fixing in search's `highlight` regardless because the HTML might have been included prior, e.g. the bug is shown on http://www.theyworkforyou.com/debates/?id=2014-07-02b.893.0&s=%22Shabana%22 which doesn't involve the glossary. I guess this explains why the order hadn't been swapped previously, though that would fix this particular case easily enough.

classes/SectionView/SectionView.php
```diff
@@ -160,6 +160,12 @@ protected function display_section_or_speech($args = array()) {
         }
 
         // Do all this unless the glossary is turned off in the URL
+        if ($SEARCHENGINE) {
```
dracos: The comment line above is referring to the line after this added block (so this block needs moving up a line or two, or put back where it was if the other commit now fixes it a different way).

www/includes/easyparliament/searchengine.php
```diff
@@ -532,7 +532,7 @@ public function highlight_internal($body, $stemmed_words) {
 
         foreach ($this->phrases as $phrase) {
             $phrasematch = join($phrase, '[^'.$this->wordchars.']+');
-            array_push($findwords, "/\b($phrasematch)\b/i");
+            array_push($findwords, "/\b($phrasematch)\b(?!(?>[^<>]*>))/i");
```
dracos: Nice :)

dracos:
Looks good.

PR #919, opened 2015-07-08
Move static pages to new style templates

### LINKED ISSUE #914: Move static pages to new design

struan:
Minimal changes other than moving to new templates and wrapping in the minimum of structure divs. 

@zarino or @wrightmartin should probably have a look.

Fixes #914


zarino:
screen shot 2015-07-08 at 17 07 19

The page content needs to be in some sort of white-backgrounded container. Like the `.business-section` and `.business-section__*` classes I added to https://github.com/mysociety/theyworkforyou/pull/920, or the `.panel` classes used on the homepage and MP profile page.

zarino:
Ok @struan I've made the bare minimum change, to put `.panel` around these pages so they fit in.

I notice the line lengths are quite unwieldy, so I've opened a separate ticket for that: https://github.com/mysociety/theyworkforyou/issues/921

Issue #772, opened 2015-03-02
Move day/year/bill/section summary pages to new design


dracos:
These are all powered by the same bits of code. e.g.
Day page http://www.theyworkforyou.com/debates/?d=2015-02-10
Year page http://www.theyworkforyou.com/debates/?y=2015
Summary pages http://www.theyworkforyou.com/debates/
http://www.theyworkforyou.com/written-answers-and-statements/
http://www.theyworkforyou.com/pbc/
Bill page http://www.theyworkforyou.com/pbc/2014-15/National_Health_Service_%28Amended_Duties_and_Powers%29_Bill/

zarino:
The homepage debate listing styles could provide a good basis for this. Here's how it looks without the homepage hero, and with a calendar:

![screen shot 2015-06-01 at 15 29 11](https://cloud.githubusercontent.com/assets/739624/7915137/74acc6c4-0873-11e5-8da3-d03442f44d3d.png)

Aside from that, feels like it needs a clear header to explain what the page is about. We should consider how much priority the search should be given (gut instinct is that the inverted search box, as per the homepage, is too heavy).

Worth noting that there are slight differences between how debates, written answers, and bills are presented. Written answers are the simplest. Debates are mostly like written answers, except sometimes they are nested (eg: sections inside of "Oral answers to questions" at the start of a day's business). Bills committees are so infrequent that it doesn't make sense to show them one day at a time – so there's no calendar on the Bills page.

zarino:
I shall also leave these here, as they were the original wireframes for how a reimagined debates browsing experience could look.

Key things I think we should keep:

* All the stuff Parliament generates, presented on a single page (rather than separate silos for debates, written answers, and bill committees).
* Explanations of what jargon like "Written answers" and "Public Bill Committees" mean.
* Simplified navigation: **Who** is in Parliament, **what happens** in Parliament, and **what’s happening right now?**

![3 debates statements](https://cloud.githubusercontent.com/assets/739624/7916049/669f7fda-0879-11e5-9ddd-3bce81d91f22.jpg)

![8 today](https://cloud.githubusercontent.com/assets/739624/7916055/697a5450-0879-11e5-8557-b999418ce993.jpg)

zarino:
[...]

PR #920, opened 2015-07-08
Styling and static markup for "Business" overview pages

zarino:
Designs for #772. (Connected to #772.) Ready for @struan to wire up with real data.

Try the pages out at [/new/debates](http://twfy.mysociety/new/debates), [/new/debates?d=YYYY-MM-DD](http://twfy.mysociety/new/debates?d=YYYY-MM-DD), and [/new/debates?y=YYYY](http://twfy.mysociety/new/debates?y=YYYY).

screen shot 2015-07-08 at 16 53 42

screen shot 2015-07-08 at 16 54 19

screen shot 2015-07-08 at 16 54 09


www/docs/style/sass/pages/_business.scss
```diff
@@ -0,0 +1,378 @@
...
+        padding: 0;
+        border-bottom: 1px solid $colour_off_white;
+
+        & > * {
+            display: block;
+            padding: 0.5em 0.55556em;
+        }
+
+        &.no {
+            color: $light-text;
```
dracos: Don't know if this is my monitor, but I can't really tell the difference at all between `.no` days and other days in the calendar unless I squint.

www/includes/easyparliament/templates/html/section/year.php
```diff
@@ -0,0 +1,39 @@
...
+                <a href="<?= $next['url'] ?>" class="calendar__controls__next"><?= $next['title'] ?> &rarr;</a>
+                <?php } else { ?>
+                <span class="calendar__controls__next"></span>
+                <?php } ?>
+            </div>
+            <?php if ( isset($years) ) {
+                foreach ( $years as $year => $months ) { ?>
+            <div class="calendar-year">
+                <?php foreach ($months as $month => $dates) {
+                  include '_calendar.php';
```
dracos: The months don't fit correctly always, e.g. screen shot 2015-08-11 at 16 32 29

www/includes/easyparliament/templates/html/section/_sp_desc.php
```diff
@@ -0,0 +1,7 @@
+                <p>
+                The Scottish Parliament is the national legislature of Scotland,
+                located in Edinburgh. The Parliament consists of 129 members known
+                as <a href="/msps/">MSPs</a>, elected for four-year terms; 73 represent
```
dracos: This link appears as blue on green in the header on /sp

www/includes/easyparliament/templates/html/section/_calendar.php
```diff
@@ -0,0 +1,166 @@
...
+                print "<td colspan=\"$remainingDays\">&nbsp;</td>";
+            }
+?>
+            </tr>
+        </tbody>
+    </table>
+  <?php if(!isset($years)){ ?>
+    <div class="calendar__footer">
+    <?php
+        $y = new URL($this_page);
```
dracos: This goes wrong on /written-answers-and-statements, and on /debates, not linking to the right URL in most cases.

www/includes/easyparliament/templates/html/section/day.php
```diff
@@ -0,0 +1,89 @@
...
+                            </div>
+                        <?php } else { ?>
+                            <a href="<?= $row['listurl'] ?>" class="business-list__title">
+                                <h3>
+                                    <?= $row['body'] ?>
+                                </h3>
+                            </a>
+                        <?php }
+                        if ( isset($row['contentcount']) && $row['contentcount'] > 0 ) { ?>
+                            <p class="business-list__meta">
```
dracos: Any chance the N speeches could perhaps be on the same row as the heading? The day page is very much more spaced out than previously, and I think that might help a bit.

www/docs/style/sass/pages/_business.scss
```diff
@@ -0,0 +1,378 @@
...
+    margin-bottom: 0.8em;
+    margin-top: -0.4em;
+    line-height: 1.2em;
+}
+
+.business-section__header__description {
+    p {
+        margin: 0.5em 0 0 0;
+        max-width: 36em;
+        color: mix($primary-color, #fff, 25%);
```
dracos: This fails a contrast check (plus I can't read it :) ).

www/includes/easyparliament/templates/html/section/day.php
```diff
@@ -0,0 +1,89 @@
...
+    <div class="business-section">
+        <div class="business-section__header">
+            <h1 class="business-section__header__title">
+                <?= $parent_title ?>
+            </h1>
+            <p class="business-section__header__date">
+                <?= format_date( $info['date'], LONGERDATEFORMAT ) ?>
+            </p>
+            <div class="business-section__header__description">
+                <?php include '_' . $section . '_desc.php'; ?>
```
dracos: Even at quite a large browser size (I'm currently on 1471x800), half of this page is heading; I see 3 entries on e.g. `/debates/?d=2013-07-10` as opposed to 10 or so in the old design. Could the description be relegated to the side where it used to be?

www/includes/easyparliament/templates/html/section/wrans_index.php
```diff
@@ -0,0 +1,7 @@
+<div class="full-page__row">
+
+    <?php $content = $wrans; $title = "Written Answers"; include '_business_section.php'; ?>
+    <?php $search_title = 'Search Written Answers'; include '_search.php'; ?>
```
dracos: This template is also being used for written-answers-and-statements, in which case it's searching both, if I understand it right. Perhaps this could be made clearer in general in this template?

www/includes/easyparliament/templates/html/section/ni_index.php
```diff
@@ -0,0 +1,24 @@
+<div class="full-page__row">
```
dracos: This page is showing the header twice, and calendar twice.

www/includes/easyparliament/templates/html/section/index.php
```diff
@@ -0,0 +1,19 @@
+<div class="full-page__row">
```
dracos: I think something more needs to be done for /debates and /written-answers-and-statements. As it is, it's not clear to me that there is more equivalent data underneath the search box or that the page has anything more than the first thing (Commons debates or Written answers) on it.

classes/SectionView/SectionView.php
```diff
@@ -21,23 +22,116 @@ public function __construct() {
...
+
+    protected function addCommonData($data) {
+        global $DATA, $this_page;
+
+        $common = new \MySociety\TheyWorkForYou\Common;
+        $data['urls'] = $this->getURLs($data);
+        $data['popular_searches'] = $common->getPopularSearches();
+        $data['recess_major'] = $this->getRecessMajor($data);
+
+        $nextprev = $DATA->page_metadata($this_page, 'nextprev');
```
dracos: This is duplicated in getURLs (which puts next/prev URLs only in $urls) - are both needed?

www/includes/easyparliament/templates/html/section/column.php
```diff
@@ -0,0 +1,431 @@
+<?php
```
dracos: This file overlaps a lot with section.php, it seems - any chance it could use section.php instead?

www/includes/easyparliament/templates/html/section/day.php
```diff
@@ -0,0 +1,89 @@
...
+                                <?= $row['contentcount'] == 1 ? '1 speech' : $row['contentcount'] . ' speeches' ?>
+                            </p>
+                        <?php }
+                        if ( isset($row['excerpt']) ) { ?>
+                            <p class="business-list__excerpt">
+                                <?= trim_characters($row['excerpt'], 0, 200 ) ?>
+                            </p>
+                        <?php }
+                    } else { ?>
+                    <ul>
```
dracos: This is creating a sub list by creating a new 1-item list for each entry in the sublist, which doesn't seem right.

dracos:
A number of design queries here as well as some minor code ones.

www/docs/new/debates.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
dracos: Presumably not needed :) (And I think some other files can be removed, e.g. hansard_calendar?)

zarino:
So, a quick rundown of the design issues I'm aware of (add to the list if I've missed any):

- [x] `td.no` in calendars isn’t distinctive enough
- [x] Month calendars in year view need clearing
- [x] Should we move header "description" into sidebar, to bring page content further above the fold?
- [x] <del>…If not, links inside green headers need to be white not blue</del>
- [x] <del>…and `.business-section__header__description` is too low contrast</del>
- [ ] Dark grey search boxes between business-sections look like the end of the page.
- [x] Wasted space next to “X speeches” – put it next to debate title on line above?
- [x] A new `<ul>` is created for each subelement in section/day.php:51–69

I've fixed all the ones I know about, apart from the dark grey search box. Since we already track scroll depth on TWFY, I wonder whether we could deploy the search box as-is, and revisit in a month to see whether it impacted scroll depth on these pages? If it did, we can remove it, and we've learnt a lesson for similar designs in future. If not, win-horse.

dracos:
Happy to deploy without any search box change; I was more wondering if that page needed some way at the top to signify it had different content within it, but I imagine these are little used pages (hopefully). Looks good, thanks, the only thing I can see is that the day.php list is now one ul, whereas I think the sub lists should still be sublists, there should just only be one of them per group of sublist things :)

zarino:
> the only thing I can see is that the day.php list is now one ul, whereas I think the sub lists should still be sublists

I spent a little while trying to work out an elegant way of keeping them as nested lists, but couldn't work out how best to close the child `<ul>` once we reach the end of the child list items. If you can work out how to make the PHP/HTML do what you want, then I can revert the CSS to match.

PR #924, opened 2015-07-16
improve vagrant config to install xapian

### LINKED ISSUE #586: Get Xapian running in default vagrant config
What it says on the tin. Vagrant setup should automatically configure the dev site to use Xapian and the PHP Xapian bindings.

@jacksonj04 and I tried doing this by hand this morning but after 45 minutes of #fail, we figured it should be estimated as a proper ticket. Estimators beware: this isn't as easy as it sounds. But it is totally worth doing, since non-functioning search was [a blocker on the debate page redesign](https://github.com/mysociety/theyworkforyou/pull/587), and will no doubt block [further page redesigns](https://github.com/mysociety/theyworkforyou/issues/555) to come.

### LINKED ISSUE #845: Vagrant box doesn't work without manually running install.sh
Had to run ```/data/twfy/puphpet/files/exec-once/install.sh``` manually to get everything set up after first ```vagrant up```

struan:
update puphpet install script to compile and install the xapian package
and create the search index.
Also add new modules for changed people code and update the upcoming
entries

Fixes #586  and Fixes #845


dracos:
All worked fine, apart from compass at the end which died due to wrong version of multi_json being installed. This is because compass is being installed in rum as part of the server set up (and getting multi_json 1.11.2 as part of that), but when you run compass inside the repo it sees the Gemfile and Gemfile.lock and looks for 1.11.0 and dies as it can't find it. I've added another call to bundle install in the repo just to make sure the right version is installed in the rvm. So as long as that looks okay to you, should be good to merge.

jacksonj04:
Just used this to spin up a new (as far as I can tell fully working) box, so :+1: from me.

dracos:
Merged in b9d0fbb

PR #925, opened 2015-07-16
only set banner_text page variable if banner content

### LINKED ISSUE #915: Banner sometimes displayed when there is no banner text set
@zarino explains:

1. [The template](https://github.com/mysociety/theyworkforyou/blob/8c494d096b6d7c294896e164d7aca16fb2011c8d/www/includes/easyparliament/templates/html/header.php#L197) checks whether `(isset($banner_text))`
2. [The view](https://github.com/mysociety/theyworkforyou/blob/5a6821687889e724c19f717e7be1bfcfa7363baf/www/includes/easyparliament/page.php#L284-L285) gets `$banner_text` from the Banner model’s `get_text()` method
3. [The model](https://github.com/mysociety/theyworkforyou/blob/17dd26d0828dc9ee2b2a4fbdae4ea11441cccdc8/classes/Model/Banner.php#L28-L44) returns a string, or `NULL`, or `False`, based on a) whether `memcache->get()` finds a row in its table, and, if it doesn't, b) whether there’s a row in the `editorial` MySQL table instead

On @zarino's VM, [the memcache check](https://github.com/mysociety/theyworkforyou/blob/fda8a69bc70ec39305fa262d932209644159cfea/classes/Memcache.php#L29-L33) returns False, meaning the (empty) `editorial` table is queried, returning no results, and leaving `$banner_text` as `False` which passes the `isset($banner_text)` check in the template.

We have a feeling, on the live site, the memcache finds a value of `NULL`, and that gets returned instead, meaning `$banner_text` is `NULL`, which gets caught by the `isset($banner_text)` check in the template.

Should the template check for `!empty($banner_text)` instead of `isset($banner_text)`?

struan:
Previously if the banner_text page variable was set to anything, even an
empty string, then the banner would be displayed with no content. This
makes sure that we only set the banner_text variable if there is some
banner content.

Fixes #915

### DIFF
--- a/tests/PageTest.php
+++ b/tests/PageTest.php
@@ -72,8 +72,8 @@ class PageTest extends FetchPageTestCase
     public function testBanner() {
         $banner = new MySociety\TheyWorkForYou\Model\Banner;
 
-        $banner->set_text('');
         $page = $this->fetch_page( array( 'url' => '/' ) );
+        $this->assertNotContains('<div class="banner">', $page);
         $this->assertNotContains('This is a banner', $page);
 
         $banner->set_text('This is a banner');
@@ -82,6 +82,7 @@ class PageTest extends FetchPageTestCase
 
         $banner->set_text('');
         $page = $this->fetch_page( array( 'url' => '/' ) );
+        $this->assertNotContains('<div class="banner">', $page);
         $this->assertNotContains('This is a banner', $page);
     }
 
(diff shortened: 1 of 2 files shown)


classes/Renderer.php
```diff
@@ -420,7 +420,10 @@ public static function output($template, $data = array())
 
         # banner text
         $b = new Model\Banner;
-        $data['banner_text'] = $b->get_text();
+        $text = $b->get_text();
+        if ( $text ) {
+            $data['banner_text'] = $text;
+        }
```
dracos: This is fine, but would it be simpler to leave this as it was and changing the template to `if ($banner_text)` instead of an `isset`? Would it ever be unset?

PR #935, opened 2015-07-28
Remove all Quant2 tracking and related bits and pieces.

jacksonj04:
This piece of research has concluded on TWFY.

Removes changes introduced in #843 and #872.


dracos:
Looks good to me.

PR #943, opened 2015-08-05
Move Alert page to the new design

### LINKED ISSUE #929: Clean up styling of alerts pages
theyworkforyou email alerts - theyworkforyou 20150803

zarino:
Fixes #929.

Selection of screenshots:

screen shot 2015-08-05 at 17 53 58

screen shot 2015-08-05 at 17 58 37

screen shot 2015-08-05 at 17 53 43

screen shot 2015-08-05 at 17 50 38


dracos:
Not sure why composer.lock is changing in this.

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+              <?php } else { ?>
+                <h3>Request a new TheyWorkForYou email alert</h3>
+              <?php } ?>
+
+                <form action="<?= $actionurl ?>" method="post" class="alert-page-main-inputs">
+                  <?php if (!$email_verified) { ?>
+                    <p>
+                        <input type="email" class="form-control" placeholder="Your email address" name="email" id="email" value="<?= _htmlentities($email) ?>">
+                        <?php if (isset($errors["email"]) && $submitted) { ?>
+                          <p><?= $errors["email"] ?></p>
```
dracos: This isn't highlighted as an error at all, I left my email address blank and missed the message.

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+                        <?php } ?>
+                    </p>
+                  <?php } ?>
+
+                    <p>
+                      <?php if ($pid) { ?>
+                        <input type="text" class="form-control" name="alertsearch" id="alertsearch" disabled="disabled" value="<?= $pid_member->full_name() ?>
+                        (<?= _htmlspecialchars($pid_member->constituency()) ?>)">
+                      <?php } else if ($keyword) { ?>
```
dracos: Should be elseif.

zarino:
> Not sure why composer.lock is changing in this.

Nor am I. I just assumed it was one of those things that randomly gets updated, and that I needed to check in.

classes/AlertView/Standard.php
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+        if ( $this->data['alertsearch'] ) {
+            $this->data['search_text'] = $this->data['alertsearch'];
+        } else {
+            $this->data['search_text'] = $this->data['keyword'];
+        }
+    }
+
+    private function prettifyCriteria($alert_criteria) {
```
dracos: The current list of your alerts spots multiple speaker: in it, this function doesn't (though I realise criteria_pretty doesn't either, just the current list of alerts didn't use that, wonder if it's way is better...).

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+                  ?>
+                    <li>
+                        <form action="<?= $actionurl ?>" method="post">
+                            <input type="hidden" name="t" value="<?= _htmlspecialchars($token) ?>">
+                            <input type="hidden" name="email" value="<?= _htmlspecialchars($email) ?>">
+                            <input type="hidden" name="pid" value="<?= $q->field($n, 'person_id') ?>">
+                            When
+                            <?= member_full_name($q->field($n, 'house'), $q->field($n, 'title'), $q->field($n, 'given_name'), $q->field($n, 'family_name'), $q->field($n, 'lordofname') ) ?>
+                            <?php if ($q->field($n, 'constituency')) { ?>
+                                ( <?= $q->field($n, 'constituency') ?> )
```
dracos: Don't think we want spaces inside brackets.

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+        <div class="alert-section">
+            <div class="alert-section__primary">
+
+              <?php if ($pid) { ?>
+                <h3>
+                    Sign up for an alert when <?= $pid_member->full_name() ?>
+                    (<?= _htmlspecialchars($pid_member->constituency()) ?>)
+                    speaks.
+                </h3>
+              <?php } else if ($keyword) { ?>
```
dracos: elseif

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+                        <form action="<?= $actionurl ?>" method="post">
+                            <input type="hidden" name="t" value="<?= _htmlspecialchars($token) ?>">
+                            <input type="hidden" name="email" value="<?= _htmlspecialchars($email) ?>">
+                            <input type="hidden" name="pid" value="<?= $q->field($n, 'person_id') ?>">
+                            When
+                            <?= member_full_name($q->field($n, 'house'), $q->field($n, 'title'), $q->field($n, 'given_name'), $q->field($n, 'family_name'), $q->field($n, 'lordofname') ) ?>
+                            <?php if ($q->field($n, 'constituency')) { ?>
+                                ( <?= $q->field($n, 'constituency') ?> )
+                            <?php } ?>
+                            speaks.
```
dracos: Can we think of something like 'speaks' that covers the fact lots of it isn't spoken? :)

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+                            <?= $member->full_name() ?>
+                            ( <?= _htmlspecialchars($constituency) ?> )
+                            <input type="submit" class="button small" value="Subscribe"></form>
+                        </li>
+                <?php } ?>
+                </ul>
+              <?php } ?>
+
+              <?php if ($alertsearch) { ?>
+                <h3>Great! Can you just confirm what you mean?</h3>
```
dracos: If there have been name matches, but the person wants to sign up for mentions of that name, this heading is confusing. You get a list of name matches, then this heading.

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+              <?php if (isset($constituencies) && count($constituencies) > 0) { ?>
+                <h3>Sign up for alerts when MPs for constituencies matching <i><?= _htmlspecialchars($alertsearch) ?></i> speak</h3>
+                <ul>
+                <?php foreach ($constituencies as $constituency => $member) { ?>
+                    <li>
+                        <form action="<?= $actionurl ?>" method="post">
+                            <input type="hidden" name="t" value="<?= _htmlspecialchars($token) ?>">
+                            <input type="hidden" name="email" value="<?= _htmlspecialchars($email) ?>">
+                            <input type="hidden" name="pid" value="<?= $member->person_id() ?>">
+                            <?php if ($valid_postcode) { ?>
```
dracos: I'm not sure we can get here with more than one constituency and valid_postcode being set due to the recently added shortcutting of a postcode.

classes/AlertView/Standard.php
```diff
@@ -0,0 +1,391 @@
...
+                $MEMBER = new \MEMBER( array('postcode' => $postcode) );
+                // move the postcode to the front just to be tidy
+                $tidy_alertsearch = $postcode . " " . trim(str_replace("$postcode", "", $this->data['alertsearch']));
+                $alertsearch_display = str_replace("$postcode ", "", $tidy_alertsearch);
+
+                $this->data['member_alertsearch'] = str_replace("$postcode", "speaker:" . $MEMBER->person_id, $tidy_alertsearch);
+                $this->data['member_displaysearch'] = $alertsearch_display;
+                $this->data['member'] = $MEMBER;
+
+                if ( $this->data['scottish_text'] ) {
```
dracos: Hmm, the postcode shortcutting means this is no longer reached if you enter a Scottish postcode, it jumps to assuming you want MP. Currently this doesn't matter given we don't have Scottish updating, but just to note as and when that's fixed we might want to not shortcut Scottish postcodes so that this bit of code runs and lets you pick.

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -0,0 +1,371 @@
...
+                        <?php if (isset($errors["email"]) && $submitted) { ?>
+                          <p><?= $errors["email"] ?></p>
+                        <?php } ?>
+                    </p>
+                  <?php } ?>
+
+                    <p>
+                      <?php if ($pid) { ?>
+                        <input type="text" class="form-control" name="alertsearch" id="alertsearch" disabled="disabled" value="<?= $pid_member->full_name() ?>
+                        (<?= _htmlspecialchars($pid_member->constituency()) ?>)">
```
dracos: The large spacing between name/constituency is shown in the disabled input, at least in Safari.

classes/AlertView/Standard.php
dracos: Probably good to have just one function that does this speaker:NN replacement, anyway, given I see keyword/alertsearch elsewhere also doing it.

dracos:
Looks good :) The "Great!" heading when there are also name results and the lack of error highlighting are the main display things, prettifying a search string the main code thing, the rest is v minor.

www/includes/easyparliament/templates/html/alert/index.php
struan: "speaks or writes"?

www/includes/easyparliament/templates/html/alert/index.php
struan: If you type in something like "Fife" that matches more than one constituency you get this.

dracos: Yes, but won't `$valid_postcode` always be false?

struan: Ah right yes, I see what you mean.

struan:
Outstanding issues:

- [ ] blank email error message

zarino:
@struan I've added styling for input errors:

screen shot 2015-08-18 at 11 45 18

While I was there, I noticed the heading above them is a bit weird – **"[[foobar] is mentioned] is mentioned"** – is this the only time it'll be shown like that (in which case, I'm inclined to let it go) or is this a wider problem?

PR #945, opened 2015-08-06
Output new header/footer from old page class

### LINKED ISSUE #938: Spit out new headers and footer from page class
Update the page class to output the new headers and footers in order to top and tail the remaining pages on the site with the new design.

struan:
This is to move all the remaining pages over to the new design by at least wrapping them in the new header and footer, plus a bit of fixup CSS and HTML to correct the worst issues.

Fixes #938

### DIFF
--- a/classes/Renderer/User.php
+++ b/classes/Renderer/User.php
@@ -1,0 +1,176 @@
+<?php
+
+namespace MySociety\TheyWorkForYou\Renderer;
+
+/**
+ * User data for headers
+ */
+
+class User
+{
+    public $data;
+
+    private $user;
+    private $pagedata;
+    private $page;
+    private $returl;
+
+    public function __construct() {
+        global $THEUSER, $DATA, $this_page;
+
+        $this->user = $THEUSER;
+        $this->pagedata = $DATA;
+        $this->page = $this_page;
+        $this->data = array();
+        $this->setupNavLinks();
+        $this->addRepLinks();
+    }
+
+    private function setupNavLinks() {
+
+        $this->data['user_nav_links'] = array();
+
+        // We may want to send the user back to this current page after they've
+        // joined, logged out or logged in. So we put the URL in $returl.
+        $URL = new \URL($this->page);
+        $this->returl = $URL->generate('none');
+
+        //user logged in
+        if ($this->user->isloggedin()) {
+            $this->addLoggedInLinks();
+        } else {
+            $this->addLoggedOutLinks();
+        }
+    }
+
+    private function addLoggedInLinks() {
+        // The 'Edit details' link.
+        $menudata   = $this->pagedata->page_metadata('userviewself', 'menu');
+        $edittext   = $menudata['text'];
+        $edittitle  = $menudata['title'];
+        $EDITURL    = new \URL('userviewself');
+        if ($this->page == 'userviewself' || $this->page == 'useredit' ) {
+            $editclass = 'on';
+        } else {
+            $editclass = '';
+        }
+
+        // The 'Log out' link.
+        $menudata   = $this->pagedata->page_metadata('userlogout', 'menu');
+        $logouttext = $menudata['text'];
+        $logouttitle= $menudata['title'];
+
+        $LOGOUTURL  = new \URL('userlogout');
+        if ($this->page != 'userlogout') {
+            $LOGOUTURL->insert(array("ret"=>$this->returl));
+            $logoutclass = '';
+        } else {
+            $logoutclass = 'on';
+        }
+
+        $username = $this->user->firstname() . ' ' . $this->user->lastname();
+
+        $this->data['user_nav_links'][] = array(
+            'href'    => $LOGOUTURL->generate(),
+            'title'   => $logouttitle,
+            'classes' => $logoutclass,
+            'text'    => $logouttext
+        );
+        $this->data['user_nav_links'][] = array(
+            'href'    => $EDITURL->generate(),
+            'title'   => $edittitle,
+            'classes' => $editclass,
+            'text'    => $edittext
+        );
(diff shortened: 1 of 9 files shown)


classes/Renderer/User.php
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    private function addLoggedInLinks() {
+        // The 'Edit details' link.
+        $menudata   = $this->pagedata->page_metadata('userviewself', 'menu');
+        $edittext   = $menudata['text'];
+        $edittitle  = $menudata['title'];
+        $EDITURL    = new \URL('userviewself');
+        if ($this->page == 'userviewself' || $this->page == 'useredit' || $this->header->top_highlight == 'userviewself') {
```
dracos: Interestingly, I don't think the top_highlight can ever be userviewself (the only thing with it as a parent is useralerts which doesn't appear to be used anywhere…? Might have missed something) so this (and the passing of header into this) could hopefully be got rid of entirely.

classes/Renderer/User.php
```diff
@@ -0,0 +1,178 @@
...
+                // And it's also silly if we're sent back to Change Password.
+                // And the join page.
+                $LOGINURL->insert(array("ret"=>$this->returl));
+            }
+            $loginclass = '';
+        } else {
+            $loginclass = 'on';
+        }
+
+            $this->data['user_nav_links'][] = array(
```
dracos: Spacing goes a bit wrong here (perhaps as this is a new file also get rid of the extra spaces either side of `=` and so on).

www/includes/easyparliament/page.php
```diff
@@ -19,59 +19,56 @@ class PAGE {
     public $within_stripe_sidebar = false;
 
     public function page_start() {
+        $this->checkForAdmin();
```
dracos: The old code was wrapped in a if !page_started - are we happy that can't happen with the rewrite?

www/docs/user/index.php
```diff
@@ -668,7 +667,6 @@ function display_form ( $details = array(), $errors = array() ) {
     }
 ?>
                 <div class="row">
-                <span class="label">&nbsp;</span>
```
dracos: There's a couple more further down that could be removed too, for the email updates Yes/No, and the terms/submit button itself.

dracos:
Great to have those header/footers finally gone from PHP code. Looks good!

classes/Renderer/User.php
struan: And in fact it doesn't really look like the classes are used for anything anyway.

dracos:
:jack_o_lantern:

www/docs/style/sass/app.scss
```diff
@@ -176,6 +176,8 @@ form {
 @import "pages/static";
 @import "pages/search";
 @import "pages/business";
+@import "pages/legacy";
+>>>>>>> add some styles to help legacy pages
```
dracos: diff line left in here

www/includes/easyparliament/page.php
```diff
@@ -19,59 +19,65 @@ class PAGE {
...
+            $this->page_start();
+        }
 
-    public function page_end($extra = null) {
-        $this->content_end();
-        $this->page_footer($extra);
+        echo '</div></div></div>';
+        $footer = new MySociety\TheyWorkForYou\Renderer\Footer();
+        $data['footer_links'] = $footer->data;
+        extract($data);
```
dracos: `$data` isn't initialized, and then it's extracted, so I guess this could just be `$footer_links = $footer->data;`.

classes/Renderer/Header.php
```diff
@@ -229,6 +229,8 @@ private function get_menu_highlights() {
             'section' => $section,
         );
 
+        $this->data['top_highlight'] = $top_highlight;
```
dracos: Not sure what this is added for, I can't see it being used anywhere.

www/includes/easyparliament/page.php
```diff
@@ -1287,13 +389,33 @@ public function error_message($message, $fatal = false, $status = 500) {
...
+                $this->page_start();
+            }
+
+            $this->message($message, 'error');
+
             if ($this->within_stripe()) {
                 $this->stripe_end();
             }
             $this->page_end();
+            exit();
```
dracos: I don't think we want to exit here; all the callers seem to deal with it okay, and this would mean fatal error email wouldn't be sent.

www/includes/easyparliament/page.php
```diff
@@ -1287,13 +389,33 @@ public function error_message($message, $fatal = false, $status = 500) {
...
 
         if ($fatal) {
+            if (!$this->page_started()) {
+                if (!headers_sent()) {
+                    header("HTTP/1.0 $status Internal Server Error");
+                }
+                $this->page_start();
+            }
+
+            $this->message($message, 'error');
```
dracos: I think this line isn't needed, as if the page has been started, it was output above, and if it hasn't, it'll be output by `page_start`.

www/includes/easyparliament/page.php
```diff
@@ -1241,43 +376,10 @@ public function postcode_form() {
...
         // If $fatal is true, we exit the page right here.
         // $message is like the array used in $this->message()
-
-        if (!$this->page_started()) {
-            if (!headers_sent()) {
-                header("HTTP/1.0 $status Internal Server Error");
-            }
-            $this->page_start();
-        }
+        global $page_errors;
```
dracos: Before, this would mean it sent out a (probably) 500 error if error_message was called, fatal or not, which seems like the right thing to do as we wouldn't want e.g. Google to cache the page and something has gone wrong (even if it manages to show most of the page).

dracos:
Few tiny things.

PR #948, opened 2015-08-07
Only display MP policy position if voted in strong division

### LINKED ISSUE #947: Only display policy position if MP has voted on a strong vote
In order to avoid issues with making statements about MPs positions on policies when we've not got much information only display their policy position when they've voted on a division that is marked as strong. We should still display the votes they've taken part in, just not the direction.

struan:
When we import the division votes we look at MPs who've voted in strong
divisions and if so set a flag for the current policy accordingly. We
then use this flag to decide if we should say what an MP's position on
an issue is. If they've not voted in a strong division we fall back to
saying we don't have enough information to decide, but still allow users
to see the votes for that policy.

Also, don't include policies without strong votes on the front page or
the differs from party stats.

Fixes #947

### DIFF
--- a/scripts/json2db.pl
+++ b/scripts/json2db.pl
@@ -44,6 +44,11 @@ my $votecheck = $dbh->prepare("SELECT person_id, vote FROM persondivisionvotes W
 my $voteadd = $dbh->prepare("INSERT INTO persondivisionvotes (person_id, division_id, vote) VALUES (?, ?, ?)");
 my $voteupdate= $dbh->prepare("UPDATE persondivisionvotes SET vote = ? WHERE person_id = ? AND division_id = ?");
 
+my $strong_vote_check = $dbh->prepare("SELECT data_value from personinfo where data_key = ? and person_id = ?");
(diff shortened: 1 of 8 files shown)


scripts/json2db.pl
```diff
@@ -159,6 +163,13 @@
                          . DBI->errstr . "\n";
                  }
             }
+                if ( $motion->{motion}->{policy_vote} =~ /3/ ) {
+                    my $pw_id = "public_whip_dreammp" . $dreamid . "_has_strong_vote";
+                    my $has_strong = $strong_vote_check->execute( $pw_id, $mp_id_num );
+                    if ( $strong_vote_check->rows() < 1 ) {
+                        $strong_vote_add->execute( $pw_id, 1, $mp_id_num);
```
dracos: Does this need to use strong_vote_update at some point, for if there are now no longer any strong votes?

dracos:
Looks fine, apart from presumably rare-case of strong votes changing back to non-strong.

dracos:
Woo, squash away

PR #952, opened 2015-08-14
Fix Register post election.

dracos:
This only affected the "diff" pages, not the list on the MP page itself.
The XML no longer contains 'memberid', so make looking for that
optional. The categories changed at the election, so add capability for
spotting them being different in a comparison/history.

### DIFF
--- a/www/docs/regmem/index.php
+++ b/www/docs/regmem/index.php
@@ -6,8 +6,9 @@ $dir = RAWDATA . 'scrapedxml/regmem';
 $dh = opendir($dir);
 $files = array();
 while ($file = readdir($dh)) {
-    if (preg_match('#^regmem#', $file))
+    if (preg_match('#^regmem#', $file)) {
         $files[] = "$dir/$file";
+    }
 }
 rsort($files);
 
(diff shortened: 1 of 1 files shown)


www/docs/regmem/index.php
```diff
@@ -58,8 +58,9 @@ function person_history($p) {
     $earliest = $files[0];
     foreach ($files as $_) {
         $file = file_get_contents($_);
+        $iso = preg_replace("#$dir/regmem(.*?)\.xml#", '$1', $_);
```
struan: I think I'd prefer $date which makes it a bit more obvious what's being captured.

www/docs/regmem/index.php
```diff
@@ -332,12 +335,23 @@ function prettify($s) {
...
+function cat_heading($cat_type, $date_pre, $date_post, $table = true) {
     global $cats;
-    $row = "<h3>$cat_type. $cats[$cat_type]</h3>";
-    if ($table)
-        return span_row($row, true);
-    return $row;
+    $cat_pre = isset($cats[$date_pre][$cat_type]) ? $cats[$date_pre][$cat_type] : '';
+    $cat_post = isset($cats[$date_post][$cat_type]) ? $cats[$date_post][$cat_type] : '';
+    if ($cat_pre == $cat_post || !$cat_post || !$cat_pre) {
+        if (!$cat_pre) $cat_pre = $cat_post;
```
struan: Missing {} here and in a few places below.

struan:
This looks fine, minor quibbles with variable names and formatting aside.

PR #954, opened 2015-08-17
Standardise banner position and styling

### LINKED ISSUE #949: "You're from another country" banners could do with design
For people in [other countries](https://github.com/mysociety/theyworkforyou/blob/master/www/includes/easyparliament/templates/html/header.php#L117) we display a banner (or at least the code exists to display a banner, YMMV) at the very top of TWFY pages. This could do with a lick of paint.

twfy-country-banner-nope

zarino:
The `$country` and `$banner_text` banners are now both styled the same way, and put right at the top of the page.

screen shot 2015-08-17 at 14 37 02

Fixes #949.


jacksonj04:
:+1:, with the erroneous banner for the quiz removed.

PR #957, opened 2015-08-17
Add party colours for UKIP, Plaid Cymru, DUP, and Sinn Féin

### LINKED ISSUE #690: Update party colours with newer parties
The existing party colour list hasn't been updated for a while, and doesn't feature colours for current parties of interest (or likely to be of interest in the upcoming election) such as UKIP and Green.

<del>Colour lists are found in two places:</del>

* [_mp.scss:200](https://github.com/mysociety/theyworkforyou/blob/281fb424057adfcec83f1e33928fab6679f7147c/www/docs/style/sass/pages/_mp.scss#L200)
* [<del>member.php:750</del>](https://github.com/mysociety/theyworkforyou/blob/8bf90af677e08d592f48707d3e8c7a939aa02c48/www/includes/easyparliament/member.php#L750)

<del>which probably want bringing in step or re-evaluating to determine if we really need both.</del>

zarino:
Fixes #690.

The full complement now looks like:

out

### DIFF
--- a/www/docs/style/sass/pages/_mp.scss
+++ b/www/docs/style/sass/pages/_mp.scss
@@ -189,6 +189,12 @@
 
             .party {
                 &:before {
+                    display: inline-block;
+                    width: 1.4em;
+                    height: 1.4em;
+                    line-height: 1.4em;
+                    vertical-align: 0.2em;
+                    text-align: center;
                     font-size: em-calc(12);
                     content: 'O';
                     margin-right: 0.5em;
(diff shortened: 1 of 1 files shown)


jacksonj04:
:+1:

jacksonj04:
This won't compile on the TWFY vagrant box, and might not do on production. UTF-8 encoding errors, you may want to tack a `@charset "UTF-8";` at the top of the .scss file.

PR #959, opened 2015-08-18
Add banner linking 50% of users out to research survey

jacksonj04:


dracos:
Looks fine; could make some tiny optimisations to the JS, but nothing important (e.g. I don't think it needs the second argument to the calls to `substring`, or the `var` in the second setting of `t`, and so on).

PR #962, opened 2015-08-24
Update Dependencies

jacksonj04:
* Update PHPUnit dependency version to 4.8.
* Update versions of all Composer dependencies.

Addresses [Gemnasium](https://gemnasium.com/mysociety/theyworkforyou) warnings.


dracos:
Looks good, thanks!

PR #963, opened 2015-08-24
Replace legacy className() constructors with future-proof __construct()

jacksonj04:
Old-style constructors will throw warnings in PHP 7, and will not be supported in PHP 8. This is mostly to keep the build logs free of warnings so we can see what is actually broken.

[Before](https://travis-ci.org/mysociety/theyworkforyou/jobs/76981612) vs [after](https://travis-ci.org/mysociety/theyworkforyou/jobs/76988009).

**NB:** You probably also want to merge the changes in commonlib, since nothing should be running on PHP 4.

### DIFF
--- a/www/includes/data.php
+++ b/www/includes/data.php
@@ -73,7 +73,7 @@ class DATA {
 
 
 
-    public function Data() {
+    public function __construct() {
 
         include_once METADATAPATH;	// defined in config.php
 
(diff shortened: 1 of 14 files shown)


dracos:
Could you submit a PR for the commonlib changes, rebased against the current master? Then this will need updating to point commonlib at the right commit. Looks fine otherwise.

PR #967, opened 2015-09-04
Add Content-Disposition to person CSV downloads

### LINKED ISSUE #654: CSV downloads should suggest a .csv extension via Content-Disposition
We had some TWFY support mail from someone who was having problems using the CSV files we offer (e.g. http://www.theyworkforyou.com/mps/?f=csv ) - it turned out that the problem was because the saved file had no .csv extension - we can avoid this by suggesting that it's called `MPs.csv`, for example, via the `Content-Disposition` header, e.g.:

    Content-Disposition: attachment; filename=MPs.csv

jacksonj04:
So it saves as a file rather than spitting out csv to the browser

* Fixes #654

### DIFF
--- a/www/includes/easyparliament/templates/csv/people_mlas.php
+++ b/www/includes/easyparliament/templates/csv/people_mlas.php
@@ -2,6 +2,7 @@
 
 twfy_debug("TEMPLATE", "people_mlas.php");
 
+header('Content-Disposition: attachment; filename=MLAs.csv');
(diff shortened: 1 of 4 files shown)


dracos:
Thanks for this; however, these files aren't used any more (and could be removed, presumably missed in the move of those pages to the new design), I'm afraid, it is the `sendAsCSV` function in classes/People.php.

struan:
closing in favour of #974

PR #970, opened 2015-09-07
Advanced search terms lost on pagination links

### LINKED ISSUE #956: pid is lost on search page navigation
If you search for a things said by a person then the person restriction is lost in the pagination links

struan:
The advanced search terms weren't registered as session vars for the search page so the URL constructor was throwing them away.

Fixes #956

### DIFF
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -5,7 +5,7 @@ namespace MySociety\TheyWorkForYou;
 
 class Search {
 
-    private $searchstring;
+    protected $searchstring;
     private $searchkeyword;
 
     public function __construct() {
@@ -41,7 +41,7 @@ class Search {
                 $data['template'] = 'search/by-person';
             } else {
                 $search = new Search\Normal();
-                $data = $search->search($this->searchstring);
+                $data = $search->search();
                 $data['template'] = 'search/results';
             }
         }
--- a/classes/Search/Normal.php
+++ b/classes/Search/Normal.php
@@ -5,8 +5,6 @@ namespace MySociety\TheyWorkForYou\Search;
 
 class Normal extends \MySociety\TheyWorkForYou\Search {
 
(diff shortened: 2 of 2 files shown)


www/includes/easyparliament/metadata.php
```diff
@@ -732,7 +732,7 @@
         'url'			=> 'search/',
         'robots'		=> 'noindex, nofollow',
         'heading'		=> '',
-        'session_vars'	=> array ('q', 's', 'pid', 'o', 'pop')
+        'session_vars'	=> array ('q', 's', 'pid', 'o', 'pop', 'person', 'from', 'to', 'section', 'column')
```
dracos: This confuses the links. The "Sort by" links have already incorporated all the advanced search terms into their `q` parameter, and now have duplicate advanced search parameters on top. The pagination links, previously having nothing, now also duplicate the parameters in `q` (due to the previous commit change to searchstring) and their own params.

If you don't have the second commit of this PR, only the first, then that fixes it in that all the data is passed through in the pagination links, but within the main query string, not as advanced params (so they then appear in the main box, not in the advanced search boxes). That would fix the bug, at least. If we want the data to remain in the advanced search boxes if it's entered there, then we'll need to also change the code to not include those terms within the query string when it's going to be output along with them in a URL (so only have one compiled query string for the actual searching, not re-output).

dracos:
Easy fix is to just have the first commit, with a changed commit message. Best fix would involve some changing in how the query string is being handled, as currently this will introduce some other issues.

struan:
Removing the second commit and opening up #978 to deal with the handling of the query string separately.

dracos:
Looks fine apart from typo in commit message, and as above presumably the `$this->searchstring = $searchstring;` line from Normal.php can be dropped entirely.

PR #971, opened 2015-09-07
swap speaker:$id in search results for name

### LINKED ISSUE #944: Translate id number to a name on the search results page showing a rep's recent appearances
If I'm on my MP's page and click the link to "[More of Daniel Zeichner’s recent appearances](http://www.theyworkforyou.com/search/?pid=25386&pop=1#n4)"

The largest header on that page is:
>Results 1–20 of 40 for speaker:25386

Not translating the "speaker:25386" into a name isn't very pretty and is a bit surprising, raising the question of if the page is showing what I expected.

The page title says:
>Search for speaker:Daniel Zeichner
The sidebar says:
>Subscribe to an email alert for speaker:Daniel Zeichner 

So if it's practical to show the name rather than number in the main heading too I'd have thought that'd  be good.

struan:
also do the same in the spelling correction so it's a bit easier for
people to read.

Fixes #944

### DIFF
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -46,6 +46,10 @@ class Search {
             }
         }
 
+        if ( isset($data['info']['spelling_correction']) ) {
+            $data['info']['spelling_correction_display'] = $this->prettifySearchString($data['info']['spelling_correction']);
+        }
+
         $data['searchstring'] = $this->searchstring;
         $data['urls'] = $this->get_urls();
         $data['this_url'] = $this->get_search_url();
@@ -220,5 +224,9 @@ class Search {
         $DATA->set_page_metadata($this_page, 'title', $pagetitle);
     }
 
+    private function prettifySearchString($string) {
+        $string = Utility\Search::speakerIDsToNames($string);
 
+        return $string;
+    }
 }
--- a/www/includes/easyparliament/templates/html/search/results.php
+++ b/www/includes/easyparliament/templates/html/search/results.php
@@ -63,10 +63,10 @@
                     <?php } else { ?>
                     All <?= $info['total_results'] ?> results
                     <?php } ?>
-                    for <em class="current-search-term"><?= _htmlentities(_htmlentities($searchstring)) ?></em></h2>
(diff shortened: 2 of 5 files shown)


www/includes/easyparliament/templates/html/search/results.php
```diff
@@ -63,10 +63,10 @@
                     <?php } else { ?>
                     All <?= $info['total_results'] ?> results
                     <?php } ?>
-                    for <em class="current-search-term"><?= _htmlentities(_htmlentities($searchstring)) ?></em></h2>
+                    for <em class="current-search-term"><?= _htmlentities(_htmlentities($searchstring_display)) ?></em></h2>
```
dracos: Double htmlentities? When did that creep in? :)

classes/Search.php
```diff
@@ -220,5 +225,13 @@ private function set_page_title($data) {
         $DATA->set_page_metadata($this_page, 'title', $pagetitle);
     }
 
+    private function prettifySearchString($string) {
+        if ( preg_match('/speaker:(\d+)/', $string, $m) ) {
+            $MEMBER = new MEMBER(array('person_id'=>$m[1]));
+            $name = 'speaker:' . $MEMBER->full_name();
+            $string= str_replace("speaker:$m[1]", $name, $string);
```
dracos: I thought elsewhere we had done a loop for if there were multiple speaker entries in the search string, but can't find it now, probably imagining it. Anyway, it'd be good if this replaced all speaker: references, not just one of them :)

dracos: Ooh, or I see the sidebar and the title do it using query_description_long (via variables email_desc and searchdescription) – perhaps an opportunity to factor this so it's only in one place?

dracos: (It was `criteria_pretty` in alert.php that I was thinking of in first comment, looks like.)

dracos:
Looks fine, though I think we already have 'pretty' search string functions it would be good to use/factor together first.

dracos:
There's the old `criteria_pretty` that could be changed to match `prettifyCriteria`, I guess, but that's it :)

PR #973, opened 2015-09-07
remove unused pages

### LINKED ISSUE #937: Remove unused pages from repo
- [x] get involved - now on ms.org
- [x] add glossary entry - no longer used
- [x] jargon page - no longer used
- [x] news page - possibly port to blog?
- [x] parsing directory - unlinked and not used.

struan:
Fixes #937


dracos:
Looks good; there are some addterm related things to remove too (e.g. glossary_search_form, user prompt bit when type==2, few other places), plus there's `addlink` (basically same as addterm).

www/includes/easyparliament/page.php
```diff
@@ -1733,60 +1733,6 @@ public function glossary_search_form($args) {
...
-
-        $add_link = $URL->generate('url');
-        ?>
-        <form action="<?php echo $add_link; ?>" method="get">
-        <?php echo $type; ?>
-        <p>Help make TheyWorkForYou.com better by adding a definition:<br>
-        <label for="g"><input type="text" name="g" value="<?php echo $formcontent; ?>" size="45">
-        <input type="submit" value="Search" class="submit"></label>
-        </p>
-        </form>
```
dracos: Looks like the whole `glossary_search_form` function can be removed, it's now not doing anything but setting some variables that aren't used.

dracos:
The bit in `user/prompt/index.php` is still there, though it's only used by glossary_search_form which isn't used anywhere any more. Lastly, the entries for the two pages in metadata.php can be removed, and the two references to addterm in glossary.php and user.php (for whether a user is able to add terms). Think that's it then :)

PR #974, opened 2015-09-08
add content disposition to set filename for person CSV

### LINKED ISSUE #654: CSV downloads should suggest a .csv extension via Content-Disposition

struan:
Means that when you click on the Download list as CSV link on the people
pages then it downloads a file rather than just spewing out the CSV into
the browser.

Fixes #654

### DIFF
--- a/classes/People.php
+++ b/classes/People.php
@@ -113,6 +113,7 @@ class People extends \PEOPLE {
     }
 
     public function sendAsCSV($data) {
+        header('Content-Disposition: attachment; filename=' . $this->type . '.csv');
         header('Content-Type: text/csv');
         $out = fopen('php://output', 'w');
 


classes/People.php
```diff
@@ -113,6 +113,7 @@ protected function getCSVRow($details) {
     }
 
     public function sendAsCSV($data) {
+        header('Content-Disposition: attachment; filename=' . $this->rep_plural . '.csv');
```
dracos: The rep_plural for lords is "Members of the House of Lords" which won't work without quotes, and might not be sensible to use anyway just in case. `$this->type` probably okay?

PR #979, opened 2015-09-15
Refactoring 1: Autoloading (Utilities) (Take 2)

jacksonj04:
This PR includes changes which move various 'utility' functions from scripts which are `include`d to using autoloaded classes and static methods. A further PR will move existing classes to autoloadable namespaces and remove additional inclusions.

* [Diff from previous branch](https://github.com/mysociety/theyworkforyou/compare/52f830e8047865fc13d55f44277f7499fc96ff0f...refactoring-autoloading)


dracos:
`git diff f0edf54..490b18c` to see the changes since first review. Looking good, couple of things:
* 490b18c should presumably be merged into 1b274bf.
* The warning (and emails) `uasort() expects parameter 2 to be a valid callback, function 'sort_by_count' not found or invalid function name` does not appear to be fixed by the new changes. I get this warning at the command line running the test suite, but I guess Travis doesn't have warnings switched on.

jacksonj04:
* `uasort()` now uses an anonymous function instead of a named callback, since `sort_by_count()` wasn't used anywhere else.
* A strange conditional function declaration in `Wikipedia` has been refactored. `format_to_output()` doesn't *look* to be declared anywhere, but Strange Things™ might be afoot and have it declared somewhere in the nest of dependencies loaded at that point, so the underlying logic is still there.

dracos:
Thanks; looks ready to merge :) (with those fixes squashed appropriately).

PR #985, opened 2015-09-17
Update privacy policy with an improved Google Analytics section

jacksonj04:
Doesn't change any of the intent of the policy (we still collect aggregate data to improve the site, nothing else), but it does allow for us to enable the demographic features in Google Analytics.

It also brings us up to the current guidelines from Google on what we should include in the privacy policy (explicit list of features used, statement from Google, and opt-out information).

Using text from http://www.mysociety.co.uk/about/privacy/.

### DIFF
--- a/www/includes/easyparliament/templates/html/static/privacy.php
+++ b/www/includes/easyparliament/templates/html/static/privacy.php
@@ -3,32 +3,105 @@
         <div class="panel">
             <h1>Privacy Policy</h1>
 
-            <p>Our privacy policy is very simple:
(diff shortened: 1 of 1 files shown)


abimysoc:
All the privacy stuff looks good to me. I'd like it if "How to unsubscribe to TheyWorkForYou email alerts" turned into "How to unsubscribe _from_ TheyWorkForYou email alerts", because grammars (yes, I know that wording is currently on the site and wasn't part of the proposed changes...).

PR #988, opened 2015-09-26
WIP: Remove composer.phar

jacksonj04:
`composer.phar` is a large binary blob, which is occasionally updated and the new version committed. This isn't what git is good at, so it would be preferable to remove it from the repo and use versions which are installed instead.

* [x] Update Travis configuration to use the installed version, not one from the repository.
* [ ] Make sure Composer is installed on servers (in the same vein as bundle/pip).
* [ ] Update deployment script to use the installed version, not one from the repository.


dracos:
I don't really care about composer being in the TheyWorkForYou repo; I don't think it's worth the hassle of setting something up for a server. FixMyStreet bundles cpanm and carton in a similar vein, specifically to reduce installation issues in that area.

PR #989, opened 2015-09-26
Automatically generate code documentation via Travis

jacksonj04:
Make Travis also generate code documentation and push it to [GitHub Pages](http://mysociety.github.io/theyworkforyou/) whenever `master` is updated.

**Heads up!** Before this gets merged:

* [x] Make sure the branch is actually set to `master`, rather than `travis-auto-docs`, or we'll never generate any useful documentation and things will be sad.
* [x] Swap out the access token, so it won't all break in future when my account no longer has push permissions to this repo.

### DIFF
--- a/build-docs.sh
+++ b/build-docs.sh
@@ -1,10 +1,6 @@
 #!/bin/bash
 set -e # exit with nonzero exit code if anything fails
 
-# install graphviz to generate diagrams
-sudo apt-get update -qq
-sudo apt-get install -qq graphviz
-
 # clear and re-create the out directory
 rm -rf www/docs/docs || exit 0;
 mkdir www/docs/docs;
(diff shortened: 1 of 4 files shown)


build-docs.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e # exit with nonzero exit code if anything fails
+
+# install graphviz to generate diagrams
+sudo apt-get update -qq
+sudo apt-get install -qq graphviz
```
dracos: This isn't necessary; see http://docs.travis-ci.com/user/apt/ for how to add packages without having to reenable `sudo`.

jacksonj04:
`sudo` now no longer required, builds are set to `master` branch, and the access token now belongs to the pusher account.

If someone (@dracos, @struan) could give this a thumbs up (assuming no problems) I'll squash it and merge it!

dracos:
Merged as 5267f8a3165ef81edd6a2b142fe9ad6f8e51e905

PR #992, opened 2015-09-29
Always process checks for strong votes in motions

### LINKED ISSUE #961: Gay marriage policy always listed as not enough information
For all MPs we always say we don't have enough information to decide their position on gay marriage - policy 6686, e.g. http://www.theyworkforyou.com/mp/10780/gregory_campbell/east_londonderry/divisions?policy=6686

struan:
If we skip the whole of the vote processing of a motion then it's
possible that a policy that only contains motions that are in other
policies never has the has_strong_vote mp info set and so always
displays as not having enough information to decide a position.

This moves the check into the motions loop and only skips the setting of
the vote information in the persondivisionvotes table. We also move
setting the motion_seen code to after the vote processing to avoid
always skipping updates.

Fixes #961

### DIFF
--- a/scripts/json2db.pl
+++ b/scripts/json2db.pl
@@ -130,19 +130,9 @@ foreach my $dreamid ( @policyids ) {
             }
         }
 
-        # some divisions are in more than one policy but the votes don't change so
-        # just skip them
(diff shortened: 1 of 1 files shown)


dracos:
:+1:

Issue #983, opened 2015-09-16
Put back search help


struan:
In the redesign the quick help for the search was removed and we should add it back to prevent people emailing support.

zarino:
There was talk of adding these search tips back into the sidebar, but I’m worried people won’t see them when they need them. Especially since the tips are quite verbose, so will be very easy to ignore as a big block of uninteresting text.

![screen shot 2015-09-30 at 16 10 06](https://cloud.githubusercontent.com/assets/739624/10197055/c158efa6-678d-11e5-8f0c-47f0e2a2b3b7.png)

From the sound of it, when users have contacted support about this, it's been because they want to find a specific phrase. I wonder whether a link, named exactly like that, near to the search input would be easier to spot?

![screen shot 2015-09-30 at 16 05 44](https://cloud.githubusercontent.com/assets/739624/10196958/5e7e216c-678d-11e5-9abd-e7241d0f8590.png)

Clicking it could unfold a little section explaining the search stemming / quoting tips, just like clicking "Advanced search" unfolds a section with extra inputs.

Thoughts, @struan @dracos?

abimysoc:
Can we explain in pictures? http://www.makeuseof.com/tag/become-better-at-finding-stuff-with-search-engines-boolean-search-logic-explained/

dracos:
The old text wasn't that, it was as follows:
<hr>
* TheyWorkForYou search is case-insensitive, and tries to match all the search terms within a document.
* To search for a person, just enter their name &ndash; we will search our database of people, as well as searching speeches made by that person, or that mention that person.
* To search for an exact phrase, use quotes (""). For example to find only documents contain the exact phrase "Hutton Report":<br>
`"hutton report"`<br>
Put a word in quotes if you don't want to perform stemming (where e.g. searching for `horse` will also return results with `horses`).
* To exclude a word from your search, put a minus ("-") sign in front,
for example to find documents containing the word "representation" but not the word "taxation":<br>
`representation -taxation`<br>
* If you want to search for words only when they're used near each other in the text, use "NEAR". For example, to find documents containing the word "elephant" near the word "room":<br>
`elephant NEAR room`<br>
* <strong>Advanced Users:</strong> You can perform boolean searches, with brackets, AND, OR, and XOR. The filters can be entered directly, as well as from the advanced search form, here's a selection:
  * column:123
  * party:Lab
  * department:Defence
  * section:uk section:wms
  * date:20080716
  * 20080101..20080131

[...]

PR #995, opened 2015-10-01
"Exact word" and "Excluding" inputs in advanced search form

zarino:
Part of #983.

screen shot 2015-10-01 at 15 57 47

The inputs use a set of URL parameters that were removed from the UI years ago (f96eaff), but still seem to work.

The "boolean operators" link currently goes to a non-existent FAQ entry. So that needs writing before this can be deployed.

Styling for `<code>` elements has been tweaked, to stop them looking so much like links.

### DIFF
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -139,7 +139,7 @@ class Search {
         $data['search_keyword'] = $this->searchkeyword;
 
         $is_adv = false;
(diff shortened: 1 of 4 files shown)


dracos:
Merged as c3aca6a76f4f4dac9bb3a3a6f830025b3092a09b

PR #996, opened 2015-10-02
amends h1 line-height

droom:
The line-height for H1's feel a bit too loose when wrapped.
This sets a 1.15em rule in app.scss

twfy-h1-leading


zarino:
Gets a :+1: from me. Thanks @droom.

dracos:
Thanks :)

droom:
Cheers @zarino, @dracos

PR #1003, opened 2015-10-20
Limit line lengths on static pages

### LINKED ISSUE #921: Review line lengths, especially on static pages
On pages where we have a single column of text, the lines can get uncomfortably long. We should review this.

screen shot 2015-09-28 at 15 41 40

zarino:
Fixes #921.

A really minimal fix that sets a hard limit for the length of lines on `.static-page` pages.

screen shot 2015-10-20 at 17 00 20

It's still a *bit* wide, but reducing it further would mean doing extra work to stop the headings wrapping etc. This is a good compromise, and much easier to read than the full width lines that are there now.

### DIFF
--- a/www/docs/style/sass/pages/_static.scss
+++ b/www/docs/style/sass/pages/_static.scss
@@ -4,4 +4,12 @@
     .panel > h1:first-child {
         margin-top: 0;
     }
+
+    .panel > * {
+        max-width: 700px; // avoids unmanageable line lengths
+    }
+
+    ol {
+        margin-left: 2em; // restore left padding on ordered list items
+    }
 }


droom:
I noticed there was a page that constrained div.main to 66% for viewports above 1040px. Would it make sense to standardise? It does have a .legacy-page class in there though.

(NB: The super loose leading in the h1 is down to it being two h1 blocks, not a line break.)

measure-page


Live page: http://www.theyworkforyou.com/wales/

zarino:
@droom – Good spot! I'll take a look to see where else `.legacy-page` is used (if anywhere) and replace it with `.static-page`.

It seems a shame that `/wales` has such a bare bones page, but looking at the analytics, it doesn't get much traffic at all, so probably not worth the time to make it look more substantial.

zarino:
@droom Turns out it was actually no more hassle to give the page a more substantial layout, like the other assembly sections. So I've opened a new PR for that: https://github.com/mysociety/theyworkforyou/pull/1005

There are a few other *really legacy* pages in [/www/docs](https://github.com/mysociety/theyworkforyou/tree/66e81417461a777001223c92173284076075873e/www/docs) but not convinced it's even worth touching them.

PR #1005, opened 2015-10-21
Make Wales page look less of an afterthought

zarino:
As spotted by @droom in #1003 – the "Wales" page was pretty poor, considering its position in the country switcher effectively makes it a top level page.

So this gives it a layout in keeping with the other national assembly sections.

screen shot 2015-10-21 at 16 57 26


droom:
Looks great :)

PR #1022, opened 2015-12-04
Stop imposing max-width to limit lines on legacy pages

### LINKED ISSUE #1009: Upcoming page looks broken after .legacy-page margin changes
#1003 limited line lengths on static pages, including `.legacy-page` pages.

Unknown to me, this also affected the "Upcoming" page, which isn't ideal:

screen shot 2015-10-28 at 13 25 28

Really this page should use the same sections-and-sidebars styling as the other business pages.

zarino:
Fixes #1009.

Not pretty, but it's better than before.

screen shot 2015-12-04 at 15 22 12

### DIFF
--- a/www/docs/style/sass/pages/_static.scss
+++ b/www/docs/style/sass/pages/_static.scss
@@ -13,3 +13,10 @@
         margin-left: 2em; // restore left padding on ordered list items
     }
 }
+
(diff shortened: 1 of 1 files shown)


struan:
Looks good

PR #1034, opened 2016-01-04
New GET variable to show/hide advanced search form

### LINKED ISSUE #1001: Search results advanced form flash of content
It's appearing before disappearing, would be good if it didn't.

zarino:
Fixes #1001.

The advanced search form is hidden by default, and shown if `?show_advanced_options` is something truthy. Since it's hidden by the server, rather than client-side, there is no longer a FOUC. There's then a javascript listener as before, that shows the form when the "Advanced search" link is clicked. If Javascript doesn't work, then the "Advanced search" link reloads the page with the `show_advanced_options` query parameter set to `true`.

### DIFF
--- a/www/includes/easyparliament/templates/html/search/form.php
+++ b/www/includes/easyparliament/templates/html/search/form.php
@@ -12,7 +12,7 @@
                     <?php } ?>
                     <p>
                         <ul class="search-result-display-options">
-                        <li><a href="#options" class="search-options-toggle js-toggle-search-options">Advanced search</a></li>
+                        <li><a href="?show_advanced_options=true" class="search-options-toggle js-toggle-search-options">Advanced search</a></li>
                         <?php if ( $is_adv ) { ?>
                             <?= $search_phrase ? '<li>Exactly: ' . _htmlentities($search_phrase) . '</li>' : '' ?>
                             <?= $search_exclude ? '<li>Excluding: ' . _htmlentities($search_exclude) . '</li>' : '' ?>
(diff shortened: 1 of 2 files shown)


www/includes/easyparliament/templates/html/search/form.php
```diff
@@ -12,7 +12,7 @@
                     <?php } ?>
                     <p>
                         <ul class="search-result-display-options">
-                        <li><a href="#options" class="search-options-toggle js-toggle-search-options">Advanced search</a></li>
+                        <li><a href="?show_advanced_options=true" class="search-options-toggle js-toggle-search-options">Advanced search</a></li>
```
zarino: I am a bit worried that this will completely erase any advanced options the user has set elsewhere on the page.

(eg: if they perform an advanced search, with Javascript disabled, and then click the "Advanced search" link to show the form again and edit their parameters.) 

@dracos: Should we somehow concatenate `show_advanced_options=true` onto the end of whatever query parameters are already set?

dracos: Yes, this won't really work, I'm afraid. What I was thinking was that nothing had to change server side, but that the form was hidden in CSS rather than JS using a `.js` selector (this might need to be set earlier in the HTML but that'd be good to do anyway).

For bonus points, the advanced search form could be in the HTML placed underneath the search results, and then moved with JavaScript to underneath the main form. That way if you have search results, you see them immediately, rather than the form you've just filled in, and the internal #options advanced search still works nicely.

zarino:
Closed in favour of https://github.com/mysociety/theyworkforyou/pull/1037.

PR #1037, opened 2016-01-05
Avoid advanced search options flash on page load

### LINKED ISSUE #1001: Search results advanced form flash of content

zarino:
Fixes #1001.

The FOUC is basically fixed by just the addition of [these 4 lines of CSS](https://github.com/mysociety/theyworkforyou/blob/7f6f1ae0e5de03ab4c2e51c01c2690e01dcf6932/www/docs/style/sass/pages/_search.scss#L64-L66) and the removal of [this line of JavaScript](https://github.com/mysociety/theyworkforyou/commit/898e27481ef40db92be5efbd9cd754a409821002#diff-96906dac833d9062f945ce65fb284a39L157).

But we also took the opportunity to improve the UX for users without JavaScript enabled (moving the advanced search inputs to the end of the page by default), and to tidy up the indentation in the search templates.

Unlike https://github.com/mysociety/theyworkforyou/issues/1034, there is no way for non-javascript users to show and hide the advanced options – they are just *always* visible at the end of the page.

### DIFF
--- a/www/includes/easyparliament/templates/html/search/by-person.php
+++ b/www/includes/easyparliament/templates/html/search/by-person.php
@@ -1,7 +1,9 @@
 <div class="full-page">
     <div class="full-page__row search-page">
 
-        <?php include 'form.php'; ?>
+        <form class="js-search-form-without-options">
+            <?php include 'form_main.php'; ?>
+        </form>
 
         <div class="search-page__section search-page__section--results">
             <div class="search-page__section__primary">
@@ -93,6 +95,11 @@
 
             <?php include 'sidebar.php' ?>
         </div>
+
+        <form class="js-search-form-with-options">
+            <?php include 'form_main.php'; ?>
+            <?php include 'form_options.php'; ?>
+        </form>
 
     </div>
 </div>
--- a/www/includes/easyparliament/templates/html/search/by-person.php
+++ b/www/includes/easyparliament/templates/html/search/by-person.php
@@ -7,45 +7,79 @@
 
         <div class="search-page__section search-page__section--results">
             <div class="search-page__section__primary">
-            <h2>Who says <em class="current-search-term"><?= _htmlentities($searchstring) ?></em> the most?</h2>
+                <h2>Who says <em class="current-search-term"><?= _htmlentities($searchstring) ?></em> the most?</h2>
 
-            <?php if ( isset($error) ) {
-                if ( $error == 'No results' && isset( $house ) && $house != 0 ) { ?>
-                <ul class="search-result-display-options">
-                    <li>
-                        <?php if ( $house == 1 ) { ?>No results for MPs only<?php }
-                        else if ( $house == 2 ) { ?>No results for Peers only<?php }
-                        else if ( $house == 4 ) { ?>No results for MSPs only<?php }
-                        else if ( $house == 3 ) { ?>No results for MLAs only<?php } ?> |
-                        <a href="<?= $this_url->generate('html') ?>">Show results for all speakers</a></li>
-                </ul>
+              <?php if ( isset($error) ) { ?>
+                <?php if ( $error == 'No results' && isset( $house ) && $house != 0 ) { ?>
+                  <ul class="search-result-display-options">
+                      <li>
+                        <?php if ( $house == 1 ) { ?>
+                          No results for MPs only
+                        <?php } else if ( $house == 2 ) { ?>
+                          No results for Peers only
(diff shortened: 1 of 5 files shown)


www/includes/easyparliament/templates/html/search/results.php
```diff
@@ -17,104 +18,104 @@
...
-                    </div>
-                    <?php } ?>
-                <?php } ?>
+                      <?php if ( $pagination_links ) { ?>
+                        <div class="search-result-pagination">
+                          <?php if ( isset($pagination_links['prev']) ) { ?>
+                            <a href="<?= $pagination_links['firstpage']['url'] ?>" title="First page">&lt;&lt;</a>
+                            <a href="<?= $pagination_links['prev']['url'] ?>" title="Previous page">&lt;</a>
+                          <?php } ?>
+                          <?php foreach ( $pagination_links['nums'] as $link ) { ?>
```
dracos: Here you've added `?>[newline][spaces]<?php` but above at https://github.com/mysociety/theyworkforyou/pull/1037/files#diff-35055ed30e73355e6ad23b7b10c537f5R29 you removed one - which should we have? :)

www/includes/easyparliament/templates/html/search/by-person.php
```diff
@@ -1,98 +1,134 @@
...
-                <?php } ?>
+                      <?php if ( !$wtt ) { ?>
+                        <ul class="search-result-display-options">
+                            <li>Results grouped by person</li>
+                            <li>
+                              <?php if ( $house == 0 ) { ?>
+                                Show All
+                              <?php } else { ?>
+                                <a href="<?= $this_url->generate('html') ?>">Show All</a>
+                              <?php } ?> |
```
dracos: Is it worth putting the `|` on a new line so it's not missed at the ends here?

www/includes/easyparliament/templates/html/search/results.php
```diff
@@ -17,104 +18,104 @@
...
+                </div>
+
+                <?php include 'sidebar.php' ?>
             </div>
+          <?php } ?>
 
-            <?php include 'sidebar.php' ?>
-        </div>
-        <?php } ?>
+            <?php include 'form_options.php'; ?>
```
dracos: The sidebar can currently include a form (a "mini survey", looks like); if you ended the "if results" statement after the end of the primary div, included form_options there, then add another "if results" statement for sidebar include and closing div, would that be okay?

www/includes/easyparliament/templates/html/search/results.php
zarino: I had a quick look at http://mysociety.github.io/coding-standards.html and couldn't spot any definitive guidance either way. Do you have a preference? Mine would be for single-line `<?php … ?>` statements where possible. It makes it easier to follow the nesting, since you only have to look for `<` characters, rather than also considering the indentation and trailing `{` brackets inside the PHP code block. It's a bit more Jekyll-like, and it also subconsciously discourages big chunks of view logic inside the templates.

I'll push a change to get https://github.com/mysociety/theyworkforyou/pull/1037/files#diff-35055ed30e73355e6ad23b7b10c537f5R29 into the same style too, if you're happy with that?

dracos: I don't mind at all, I just wanted it to be consistent :-) Single line is fine.

www/includes/easyparliament/templates/html/search/results.php
zarino: As discussed with @dracos in IRC, this actually raises a bigger issue – if the sidebar can itself contain a form (the "Did you find what you were looking for?" mini survey) then we can't wrap the entire page in a form as this PR suggests.

The alternatives, therefore, are:

* Move the advanced search options div back up to the top of the page, before the search results, and end the form immediately after them. This means the sidebar will no longer be inside the main form, so the mini survey will work. But it forces non-javascript users to scroll or tab through all the advanced options before they get to the search results. (TBF, this is how it currently works!)
* Use two separate forms – one for the primary search input before the results section, and another for the advanced search options after the results. We'd need to duplicate the primary `#q` search input inside the advanced search form too. Then, when the javascript moves the advanced options div up above the results list, it would need to remove the extra input.

zarino:
@dracos: I've fixed the nested form issue we spotted last week, so this should now be good to go. Let me know if you're happy, and I'll squash and merge into master.

PR #1043, opened 2016-01-20
Update policies and add a new environmental issues policy segment

struan:
This groups some existing policies into a new environment segment and also adds recently created policies.


classes/Policies.php
```diff
@@ -98,6 +98,12 @@ class Policies {
         6719 => 'limits on success <b>fees</b> paid to lawyers in <b>no-win no fee cases</b>',
         6720 => 'a statutory <b>register of lobbyists</b>',
         6721 => 'requiring the mass <b>retention of information about communications</b>',
+        6731 => 'more restrictive <b>regulation of trade union activity</b>',
+        6732 => 'allowing <b>terminally ill people</b> to be given <b>assistance to end their life</b>',
+        6733 => 'higher <b>taxes on banks</b>',
+        6734 => 'stronger <b>enforcement of immigration rules</b>',
+        6736 => '<b>a veto for MPs</b> from England, Wales and Northern Ireland <b>over laws specifically impacting their part of the UK</b>',
+        6741 => 'greater regulation of <b>hydraulic fracturing (fracking)</b> to extract shale gas'
```
dracos: Do you want to put end of line commas in to reduce the next change diff? Here, and in a few of the added numbers later on, and in policyids.pl.

dracos:
Looks like there's a test for total number of policies that needs updating :)

Issue #1055, opened 2016-02-11
TWFY doesn't know about MPs role in government


wrightmartin:
I can't find someone based on their job title, EG. Secretary of State for Health it doesn't return Jeremy Hunt.

http://www.theyworkforyou.com/search/?q=Secretary+of+State+for+Health

I'm aware TWFY is about **parliament**, and not **government** bit I think this is a use-case that stands

dracos:
Looks like this works somewhat now.

PR #1058, opened 2016-02-11
Better position in header; searching of offices

### LINKED ISSUE #1056: Currently held position should be more prominent
It's hidden away in the centre of the 'profile' section. 

screen shot 2016-02-11 at 15 10 57

http://www.theyworkforyou.com/mp/11859/jeremy_hunt/south_west_surrey

Happy for 'previously held positions' to stay there though

dracos:
This fixes #1056, and starts on #1055 (it does a search of any office positions ordered by from date descending, so you now get too much and it doesn't push exact matches to the top, so JH is at least there, but fifth).

### DIFF
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -250,12 +250,6 @@ $former_offices = $MEMBER->offices('previous', TRUE);
 // If this person has named non-committee offices, they override the default
 if (count($current_offices) > 0) {
     $position = $current_offices[0];
-} elseif (count($former_offices) > 0) {
-    # only use the position if it's current, i.e.
-    # it does not have an end date
-    if (!$former_offices[0]->to_date) {
-        $position = $former_offices[0];
-    }
 }
 
 // Finally, if this is a Votes page, replace the page description with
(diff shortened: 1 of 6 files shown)


www/docs/mp/index.php
```diff
@@ -245,7 +245,11 @@
 if (count($current_offices) > 0) {
     $position = $current_offices[0];
 } elseif (count($former_offices) > 0) {
-    $position = $former_offices[0];
+    # only use the position if it's current, i.e.
+    # it does not have an end date
+    if (!$former_offices[0]->to_date) {
```
dracos: I don't think this line will ever match, as all the former_offices have end dates from `if ($include_only == 'previous' AND $to_date == '9999-12-31')` line (in fact, even current offices will have `9999-12-31`, looks like). Think we either don't show a former position at all, or stick "Former" in front of it?

struan: Yes, you're right. No idea what I was thinking there :/

I think probably not at all is my preference. Mostly as for e.g Jack Straw this would list him as "Former Chair, House of Commons Governance Committee" which is probably the least relevant position we could list :)

PR #1066, opened 2016-02-19
Fix party stance text for single member parties

### LINKED ISSUE #1019: Wording when a party only has a single MP
As noted via a tweet:

> Douglas Carswell consistently votes the same way as "Other UKIP MPs" 

As he's the only one, we should probably drop this line if possible.

struan:
Add a getCurrentMemberCount method to party class and use that to check
if a party has only one member, and if so do not print a differs
from/votes the same as bit.

Fixes #1019

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/profile.php
+++ b/www/includes/easyparliament/templates/html/mp/profile.php
@@ -80,7 +80,7 @@ include_once INCLUDESPATH . "easyparliament/templates/html/mp/header.php";
                     <h2 data-magellan-destination="votes"><?= $full_name ?>&rsquo;s voting in Parliament</h2>
 
 
-                    <?php if (count($sorted_diffs) > 0): ?>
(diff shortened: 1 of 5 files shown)


www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -74,13 +74,13 @@
 
                     </div>
                     <?php } ?>
-                <?php } else if (count($policyPositions->positions) > 0 || count($sorted_diffs) > 0): ?>
+                <?php } else if ($party_member_count > 0 && (count($policyPositions->positions) > 0 || count($sorted_diffs) > 0) ): ?>
```
dracos: Is it possible for this to be 0 if it was the member's own party that was used for the count?

dracos:
:+1: with query about whether a new check is needed.

www/includes/easyparliament/templates/html/mp/profile.php
struan: Good point :)

PR #1071, opened 2016-02-23
Addition of FAQ about wikipedia

MyfanwyNixon:
For what it's worth (unsure it will ever be read by our users!). Please feel free to check I haven't mucked up the formatting.

### DIFF
--- a/www/includes/easyparliament/templates/html/static/help.php
+++ b/www/includes/easyparliament/templates/html/static/help.php
@@ -356,7 +356,7 @@
                 <dd>
                     <p>TheyWorkForYou automatically scans the text of debates to find phrases of more than two consecutive words which are also the title of a Wikipedia page, and then links them to those pages. Where the topic could refer to one of multiple Wikipedia entries (e.g. 'Michael Jackson' could refer to the pop star or to the once U.S. Deputy Secretary of Homeland Security), we link to the top page in Wikipedia's disambiguation list.</p>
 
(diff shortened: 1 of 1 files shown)


www/includes/easyparliament/templates/html/static/help.php
```diff
@@ -349,6 +349,15 @@
...
                        these to display their latest activity on their own websites.</p>
                 </dd>
+                <dt id="wikipedia">
+                    How do you decide which Wikipedia pages to link to?
+                </dt>
+
+                <dd>
+                    <p>TheyWorkForYou automatically scans the text of debates to find phrases of more than two consecutive words which are also the title of a Wikipedia page, and then links them to those pages. Where the topic could refer to one of multiple Wikipedia entries (e.g. 'Michael Jackson' could refer to the pop star or to the once U.S. Deputy Secretary of Homeland Security), we link to the top page in Wikipedia's disambiguation list.</p>
+
+                    <p>This generally provides a useful service whereby you can look up any unknown name or phrase easily. However, the automation of this process does sometimes mean that we link to an inappropriate Wikipedia page. If you find an example of this, please do <a href="mailto:support@theyworkforyou.com">let us know</a> and we will correct the link manually.</p>
```
tmtmtmtm: Perhaps better to say we'll _remove_ the link, rather than correct it, as otherwise people will expect us to be able to make the link go to a different page instead.

struan:
Merged in 5fb09c2e5be077858ff319c636307c3747e2bdb8 after a small rebase :)

PR #1072, opened 2016-02-23
Add AdWords conversion tracking to alert signup confirmation

### LINKED ISSUE #1077: Add Adwords conversion tracking code
See https://github.com/mysociety/Non-Site-Specific/issues/687

As a condition of the higher level Google Adwords grant, we need to be tracking at least one goal. The sooner we can get this in place, the sooner I can reapply and the sooner we'll get more users back!

@jacksonj04 says:

> https://support.google.com/adwords/answer/6095821 is the guide on doing it.

> Basically, on each site, on each page which counts as a 'conversion' (eg the WTT success page, the WDTK success page, the TWFY postcode search result page, the FMS success page) we add a bit of code which tells AdWords "whatever you did, worked". When AdWords spots an incoming "success" message, it records that against the relevant ad (assuming the user arrived on the site via an ad) as a conversion.

For TWFY I suggest the goal should be signing up for alerts or clicking on the 'contact my MP' button. Although obviously, feel free to add more if you think it's worthwhile.

jacksonj04:
Closes #1077.

### DIFF
--- a/www/includes/easyparliament/templates/html/alert/index.php
+++ b/www/includes/easyparliament/templates/html/alert/index.php
@@ -26,6 +26,25 @@
                     be later due to issues at our or parliament.uk&rsquo;s end.
                 </p>
 
+                <!-- Google Code for TWFY Alert Signup Conversion Page -->
+                <script type="text/javascript">
+                /* <![CDATA[ */
+                var google_conversion_id = 1067468161;
+                var google_conversion_language = "en";
+                var google_conversion_format = "3";
+                var google_conversion_color = "ffffff";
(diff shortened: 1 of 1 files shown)


struan:
This only works if they sign up by confirming from a link in an email and not if they sign up while already logged in. Assuming that's ok this looks good to me.

MyfanwyNixon:
Can I bump this please? Be great to get this sorted so I can reapply for the higher tier Google Grant.

tmtmtmtm:
I assume we need a response to @struan's comment/question first…

MyfanwyNixon:
As this is, right now, a tickbox exercise then that's fine, thanks. The data may prove to be useful in which case we can always revisit.

PR #1088, opened 2016-04-04
Add Contact link to main nav

### LINKED ISSUE #1053: Make contact page easier to find
e.g. by adding a link to the header, instead of hiding it in the footer.

TWFY support mail frequently arrives via the contact form on the mySociety website. Users must be looking for a contact link on TWFY, not finding it, clicking the mySociety logo top right, and coming through to the mySociety website - which _does_ have a prominent 'contact' link in the header, so they write there.

zarino:
Fixes #1053.

Nav text is now 16px rather than 18px, to make room for the extra menu item.

screen shot 2016-04-04 at 10 54 29

### DIFF
--- a/classes/Renderer/Header.php
+++ b/classes/Renderer/Header.php
@@ -240,10 +240,10 @@ class Header
         // we're within that section.
         $nav_items = array (
             array('home'),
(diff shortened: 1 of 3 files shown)


struan:
Looks good

PR #1090, opened 2016-04-05
Add always_return parameter to getMSP/getMLA.

dracos:
Factor previous special MP behaviour into getPerson function.


struan:
This looks good to me.

PR #1099, opened 2016-04-29
Implement standard mySociety footer on theyworkforyou.com

### LINKED ISSUE #1095: Implement standard mySociety footer on theyworkforyou.com
Moving this from mysociety/sysadmin#676, where there is lots of background and discussion.

TWFY does have the mySociety logo top right and a fairly discreet "This website is run by mySociety," bottom left; however the introduction of a new footer would make the affiliation more prominent.

zarino:
Todo:

- [x] Implement the footer.
- [x] Remove the old footer styles.
- [x] Work out what to do with the newsletter signup form.
- [x] Check whether we need to implement a replacement for the Facebook/Twitter event tracking.
- [x] Get site description and social links from @MyfanwyNixon.

---

Fixes #1095.

I've done a very simple switch to the new layout and styling. It's lost a few features in the migration – mainly the links to regional assemblies, international PMOs, and the twitter/facebook widgets.

screen shot 2016-04-29 at 15 16 13


struan:
This interacts badly with the sidemenu on the MP votes page:

screen shot 2016-05-23 at 15 03 50

Admittedly this is also the case on the current site but it's not quite as bad:

screen shot 2016-05-23 at 15 04 52

www/includes/easyparliament/templates/html/footer.php
```diff
@@ -1,150 +1,83 @@
...
+                </ul>
+            </nav>
+        </div>
+
+        <div class="medium-3 columns">
+            <div class="mysoc-footer__donate">
+                <p>Your donations keep this site and others like it running</p>
+                <a href="https://www.mysociety.org/donate" class="mysoc-footer__donate__button">Donate now</a>
+            </div>
+        </div>
```
struan: Should we add click tracking to the donate button?

www/includes/easyparliament/templates/html/footer.php
```diff
@@ -1,150 +1,83 @@
...
-                  </div>
+<div id="footer" class="mysoc-footer" role="contentinfo">
+    <div class="row">
+
+        <div class="medium-5 columns">
+            <h2 class="mysoc-footer__site-name">TheyWorkForYou</h2>
+            <div class="mysoc-footer__site-description">
+                <p>Making it easy to keep an eye on the UK&rsquo;s parliaments. Discover who represents you, how they&rsquo;ve voted and what they&rsquo;ve said in debates &ndash; simply and clearly.</p>
+            </div>
+            <form method="post" class="footer__newsletter-form" action="//mysociety.us9.list-manage.com/subscribe/post?u=53d0d2026dea615ed488a8834&id=287dc28511">
```
struan: We've lost the tracking on this form being used.

struan:
Not sure if we want to put the tracking things in - I imagine we probably want to leave the email signup one there just to see if the new footer changes response rates there.

And I suspect we should probably pull the side bar issues on the votes page into a new ticket rather than resolve that here.

Also, probably worth squashing the legal text changes in to the main commit.

Otherwise this looks good.

PR #1100, opened 2016-05-05
Google Analytics event tracking for footer mailing list signup form

zarino:
Data on how many people use the footer signup form would help us work out whether the signup form should be included in the new standard footer (https://github.com/mysociety/theyworkforyou/issues/1095).


struan:
Looks fine to me

PR #1110, opened 2016-06-27
add back in search all speeches/$person's speeches radio buttons

### LINKED ISSUE #1041: Search from "More recent appearances" no longer has person restriction
e.g. http://www.theyworkforyou.com/search/?pid=25269&pop=1 used to have (you can still see it in page.php search_form) a "[x] Search only Mhairi Black [ ] Search all speeches" radio buttons. As it is, if you're on that page and search, I assume it will stay within the speaker's speeches, but doesn't. I think these went missing at the time of the new design.

struan:
Fixes #1041

### DIFF
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -46,6 +46,12 @@ class Search {
             }
         }
 
+        $data['person_id'] = $argparser->person_id;
+        if ($argparser->person_id) {
+            $MEMBER = new \MEMBER(array('person_id'=>$argparser->person_id));
+            $data['person_name'] = $MEMBER->full_name();
+        }
+
         if ( isset($data['info']['spelling_correction']) ) {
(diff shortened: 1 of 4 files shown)


classes/Search.php
```diff
@@ -15,9 +16,9 @@ public function __construct() {
 
     public function display() {
         $data = array();
-        $argparser = new Search\ParseArgs();
-        $this->searchstring = $argparser->construct_search_string();
-        $this->searchkeyword = $argparser->searchkeyword;
+        $this->argparser = new Search\ParseArgs();
+        $this->searchstring = $this->argparser->construct_search_string();
+        $this->searchkeyword = $this->argparser->searchkeyword;
```
dracos: Does `$argparser` need changing to a `$this->argparser`? It doesn't appear to be used anywhere but this function.

dracos:
Yay!

PR #1111, opened 2016-06-27
stop duplication of speaker: part of search strings

### LINKED ISSUE #1105: Search duplicates PID twice
On a page such as http://www.theyworkforyou.com/search/?pid=25269&pop=1 the e.g. 'relevance' link has the PID in both `pid` and `q`. This means the query string can get longer and longer as you follow these links.

Fixing #1041 would presumably help/fix this too.

Partly caused #1104 as the Scrapy bot got longer and longer search strings.

struan:
The URL creation already pulls in the search string so we don't need to
re-add it, especially as doing so means on the next page the pid is
added again.

Fixes #1105

### DIFF
--- a/classes/Search.php
+++ b/classes/Search.php
@@ -122,7 +122,6 @@ class Search {
         $urls = array();
 
         $url = new \URL($this_page);
-        $url->insert(array('q' => $this->searchstring));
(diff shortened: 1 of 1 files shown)


dracos:
:+1:

Issue #1118, opened 2016-06-30
Revisit what share of voting pages look like on Facebook and Twitter


MyfanwyNixon:
@JenMysoc - sorry! This is another action that would really help if we can turn it round in the next week or so, too. But I think it's a quick one.

Over the next few days and weeks, people will be looking at our voting records for the leadership candidates. Currently when you share these on Facebook they look like this:

![fireshot screen capture 716 - joseph nixon - www_facebook_com_joseph_nixon_37_fref ts](https://cloud.githubusercontent.com/assets/944656/16479404/fa54dd42-3e97-11e6-9622-d90f66c8e538.jpg)

Yes that is me conversing with my husband via the medium of Facebook :(

What people really want to share is something like this:

![fireshot screen capture 717 - joseph nixon - www_facebook_com_photo_php_fbid 10153455468126735 set p_10153455468126735 type 3 theater](https://cloud.githubusercontent.com/assets/944656/16479439/21f16622-3e98-11e6-8455-37fc9bacdbd7.jpg)

...where you can see, clearly and obviously, what the voting lines are that accompany the Facebook user's comment. Even better if it somehow included a 'click through to see more voting information on [MP's name]'.

Even if it's difficult to include a specific section of the votes, there must be ways of making it more obvious that these pages lay them out simply and clearly.

MyfanwyNixon:
@zarino @wrightmartin cc-ing for thoughts, thanks!

zarino:
Facebook is pulling the image, title, and description out of the page's `<head>`. 

**Easy quick win:**

We could fairly easily change the title to be something like "Angela Eagle’s voting record – TheyWorForkYou" or whatever, and similarly, the description could be something generic like "See how Angela Eagle voted on topics like Employment, Social Issues, and Foreign Policy."

**Bit more involved:**

We could attempt to summarise, in human language, the MP’s voting record. [waves hands] But I'm a bit worried we'll never be able to construct interesting, coherent descriptions for an MP’s entire voting history.

We should also improve the image that gets shared. (See https://github.com/mysociety/orgsites/issues/873#issuecomment-229098508). I don't think we'll be able to use the MP’s photo, as it's too low resolution. But it's maybe worth investigating.

**Sharing specific bits of the page:**

[...]

PR #1121, opened 2016-06-30
Replicate page title and description in OpenGraph tags

zarino:
This way, users sharing inner pages of TWFY (eg: MP profiles) on
Facebook will benefit from the customised text that’s currently
hidden away in the `<meta type="description">` element.

Part of #1118.


zarino:
This assumes `<meta name="description">` is populated for all the pages on the site. As it happens, it’s empty on the TWFY homepage!!

Maybe there should be a default fallback for `$meta_description` on line 90 of https://github.com/mysociety/theyworkforyou/blob/623ded18aedafac29150f1202d40df85045e0296/classes/Renderer/Header.php#L90 ?

www/includes/easyparliament/templates/html/header.php
```diff
@@ -18,11 +18,12 @@
...
     <link rel="home" title="Home" href="http://<?= DOMAIN ?>/">
 
-    <meta property="og:title" content="TheyWorkForYou">
+    <meta property="og:site_name" content="TheyWorkForYou">
+    <meta property="og:title" content="<?= preg_replace('#<[^>]*>#', '', $page_title) ?>">
     <meta property="og:type" content="website">
     <meta property="og:url" content="http://<?= DOMAIN ?>">
     <meta property="og:image" content="http://<?= DOMAIN ?>/images/favicon-256.png">
-    <meta property="og:description" content="TheyWorkForYou is a website which makes it easy to keep track of your local MP's activities.">
+    <meta property="og:description" content="<?= _htmlentities($meta_description) ?>">
```
struan: Not all pages have a metadescription so we should fallback to the original if it's missing.

zarino: I have just pushed a commit that should fix that, by adding a non-empty fallback in the Header.php class. https://github.com/mysociety/theyworkforyou/pull/1121/commits/013cee1b815b5d1381c61a273ad4728080816638

struan:
Looks good to me.

zarino:
From IRC earlier:

> @zarino: I'm thinking the description could be much better on an MP's /votes page – mentioning some of the vote categories for example, and not bothering with the contributions/speeches/questions stuff…
> @zarino: …but I can't work out where to add it in the PHP controller for the /votes page
> @struan: I'll have a look tomorrow

After some more digging I think maybe it's best as an `if($pagetype == 'votes')` after the MP `$desc` is created from the MP’s hansard data in [www/docs/mp/index.php](https://github.com/mysociety/theyworkforyou/blob/master/www/docs/mp/index.php#L176-L251). Do you think that's sensible, @struan?

struan:
@zarino: yup, that looks as good a place as any.

PR #1126, opened 2016-07-05
Show vote data under human-readable voting summaries

### LINKED ISSUE #1125: Show me some real voting data on MP /votes page, not just a weasel-word summary
Spun off from #1118, we've noticed a few users mistrust the human readable voting summaries (eg: "MP X consistently voted against Topic Y"), and don’t necessarily click on the "Show votes" button to see the votes in detail.

This is particularly problematic when people take screenshots of the page, and share them on social media – making it hard for second-hand viewers to verify the claims in the screenshot.

Including some data below each human-readable statement would ensure that people can verify the claims – and if the data is kept unintrusive, it shouldn't harm the legibility of the page as a whole.

Super quick mockup:

screen shot 2016-07-04 at 13 38 56

zarino:
**Work in progress**

Fixes #1125.

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -69,6 +69,76 @@ class Divisions {
         return $this->divisionsByPolicy($q);
     }
 
+    public function getMemberDivsionSummaryForPolicy($policyID) {
+        $args = array(':person_id' => $this->member->person_id);
+        $args[':policy_id'] = $policyID;
+
+        $q = $this->db->query(
+            "SELECT count(division_id) as total, max(year(division_date)) as latest, min(year(division_date)) as earliest
+            FROM policydivisions JOIN persondivisionvotes USING(division_id)
+            WHERE person_id = :person_id AND direction <> 'abstention' AND policy_id = :policy_id",
+            $args
+        );
+
+        $max = $q->field(0, 'latest');
+        $min = $q->field(0, 'earliest');
+        $total = $q->field(0, 'total');
+
+        $q = $this->db->query(
+            "SELECT policy_vote, vote, count(division_id) as total
+            FROM policydivisions JOIN persondivisionvotes USING(division_id)
+            WHERE person_id = :person_id AND direction <> 'abstention' AND policy_id = :policy_id
+            GROUP BY policy_vote, vote",
+            $args
+        );
+
+        $votes = array('for' => 0, 'against' => 0, 'absent' => 0, 'both' => 0, 'tell' => 0);
+
+        $row_count = $q->rows();
+        for ($n = 0; $n < $row_count; $n++) {
+          $vote = $q->field($n, 'vote');
+          $policy_vote = str_replace('3', '', $q->field($n, 'policy_vote'));
+          if ( $vote == 'absent' ) {
+              $votes['absent'] += $q->field($n, 'total');
+          } else if ( $vote == 'both' ) {
+              $votes['both'] += $q->field($n, 'total');
+          } else if ( strpos($vote, 'tell') !== FALSE ) {
+              $votes['tell'] += $q->field($n, 'total');
+          } else if ( $policy_vote == $vote ) {
+              $votes['for'] += $q->field($n, 'total');
+          } else if ( $policy_vote != $vote ) {
+              $votes['against'] += $q->field($n, 'total');
+          }
+        }
+
+        $vote_summary = $votes['for'] . " for, " . $votes['against'] . " against";
+        if ( $votes['both'] ) {
+          $vote_summary .= ', ' . $votes['both'] . ' abstention';
+        }
+        $extras = array();
+        if ( $votes['absent'] ) {
+          $extras[] = "was absent for " . $votes['absent'] . ' ' . $this->votePluralise($votes['absent']);
+          $total -= $votes['absent'];
+        }
+        if ( $votes['tell'] ) {
+          $extras[] = "was a teller for " . $votes['tell'] . ' ' . $this->votePluralise($votes['tell']);
+          $total -= $votes['tell'];
+        }
+        if ($extras) {
+          $vote_summary .= ( count($extras) == 2 ? ', ' : ' and ' ) . implode(' and ', $extras );
+        }
+        $vote_str = $this->votePluralise($total);
+        if ($max == $min) {
+          return "$total $vote_str in $min - " . $vote_summary;
+        } else {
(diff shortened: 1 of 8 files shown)


struan:
I think this is now ready for some design @zarino.

screen shot 2017-01-24 at 11 07 36

zarino:
Oh, interesting – my instinct was that "X votes" at the start should be the total number of votes held on that topic, and then the "X for, X against, X absences" should add up to that total. But it looks like you’ve set the initial number to be the total number of votes *that person made* on that topic, which means, when we show "X for, X against, X absences", it adds up to a bigger number than the total.

Hmmm, not sure which is the least surprising option here – they both have their drawbacks.

@MyfanwyNixon can you think of a clear way of wording this?

struan:
I'm not hugely wedded to the wording I used. I think I just liked the idea of people able to easily see of the number of votes where they turned up was outnumbered by the ones where they did not as a proxy for how interested they were.

Also note that we need to include votes where they were a teller and votes where they abstained.

struan:
I realise it's a crude measure, and there are many reasons why MPs do not attend votes, but if you care about something and it turns our that your MP is rarely in the room then it's something you are likely to want to ask them about.

abimysoc:
I too was confused by the wording as it stands. I think showing the total number of votes on the topic, then breaking down into for/against/absent/etc would be a good thing.

MyfanwyNixon:
As it stands, I guess you could clarify by prefacing with 

>**This MP's votes** [date period]:

I would be inclined to drop the total, as Zarino suggests.
So for example:

> **This MP's votes** [2013 -2014] For: 4 Against: 0 Absent: 0

struan:
Updated text:

screen shot 2017-01-25 at 14 11 45

struan:
@zarino Not sure if you want to tweak the styling on the text here. I assume from your original mockup that the idea was that this was a link to the detail page (i.e the same place as the votes button)

screen shot 2017-01-26 at 16 42 36

zarino:
@struan: Yep, can I take a look tomorrow morning?

struan:
@zarino totally, I was just flagging it up so it got on your radar :)

zarino:
@MyfanwyNixon @struan: Would you be happy with this slightly simplified formatting?

screen shot 2017-01-27 at 12 17 51

Reasoning:

* The bold tags are distracting.
* "This MP’s votes" should be implied, just like "Generally voted for".
* Square brackets around dates are messy.

MyfanwyNixon:
Let's give it a go and see if we get any confused user support letters.

classes/Divisions.php
```diff
@@ -69,6 +69,139 @@ public function getMemberDivisionsForPolicy($policyID = null) {
         return $this->divisionsByPolicy($q);
     }
 
+    public function getMemberDivsionDetails() {
```
dracos: Division misspelt here.

classes/Divisions.php
```diff
@@ -69,6 +69,139 @@ public function getMemberDivisionsForPolicy($policyID = null) {
...
+              $summary['against'] += $q->field($n, 'total');
+          }
+
+          $policy_divisions[$q->field($n, 'policy_id')] = $summary;
+        }
+
+        return $policy_divisions;
+    }
+
+    public function getMemberDivsionSummaryForPolicy($policyID) {
```
dracos: This function isn't called anywhere, as it got removed (And also has Division misspelled)

classes/PolicyPositions.php
```diff
@@ -115,12 +117,17 @@ private function getMemberPolicyPositions ($limit = NULL) {
...
+                      $summary = $this->divisions->generateSummary($this->summaries[$policy[0]]);
+                    }
                     $this->positions[] = array(
                         'policy_id' => $policy[0],
                         'policy' => $policy[1],
                         'desc' => $dream_info['full_sentence'],
                         'has_strong' => $dream_info['has_strong'],
-                        'position' => $dream_info['position']
+                        'position' => $dream_info['position'],
+                        'summary' => $summary
```
dracos: Just as an aside, should this construct the array and then save it in both positions[] and positionsById (why does only the latter have score?). Or could we drop positions, looks like it's only used by topics and that could use positionsById easily enough (unless it's maintaining order in some way?).

classes/PolicyPositions.php
```diff
@@ -56,11 +56,12 @@ class PolicyPositions {
      * @param int      $limit    The number of policies to limit the list to.
      */
 
-    public function __construct(Policies $policies, Member $member, $limit = NULL)
+    public function __construct(Policies $policies, Member $member, $limit = NULL, $summaries = array())
```
dracos: Rather than having to pass NULL in to all the ones without a limit, perhaps the constructor could be (Policies, Member, Options) with Options being a hash that can supply limit or summaries?

classes/Divisions.php
```diff
@@ -69,6 +69,63 @@ public function getMemberDivisionsForPolicy($policyID = null) {
         return $this->divisionsByPolicy($q);
     }
 
+    public function getMemberDivsionDetails() {
+        $args = array(':person_id' => $this->member->person_id);
+
+        $policy_divisions = array();
+
+        $q = $this->db->query(
+            "SELECT policy_id, count(division_id) as total, max(year(division_date)) as latest, min(year(division_date)) as earliest
```
dracos: Could this query not be done in code using the results of the second query (if you added max/min year)? The second query fetches all the same data just grouped into more categories, so summing up total/min/max in that loop would also work. I don't know if this query is slow enough that that would matter, but I imagine code might be faster anyway?

www/docs/mp/index.php
```diff
@@ -368,77 +368,79 @@
...
-                    $policiesList->limitToSet('home'), $MEMBER
+                    $policiesList->limitToSet('home'), $MEMBER, NULL, $policySummaries
                 )
             ),
             array(
                 'key'   => 'environment',
                 'title' => 'Environmental Issues',
                 'votes' => new MySociety\TheyWorkForYou\PolicyPositions(
-                    $policiesList->limitToSet('environment'), $MEMBER
+                    $policiesList->limitToSet('environment'), $MEMBER, NULL, $policySummaries
```
dracos: Any reason the two under here don't have this added?

struan:
There's a ticket elsewhere to redo the policy pages so I'm going to leave updating those for whenever that happens I think.

PR #1175, opened 2016-09-14
remove expenses explantion and just show links

### LINKED ISSUE #313: Design of MP's expenses

struan:
Instead of explaining why we don't have expenses for an MP just link to
the information. Also fix it so we only link to the relevant
information.

Fixes #313

### DIFF
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -316,6 +316,7 @@ $data['image'] = $MEMBER->image();
 $data['member_summary'] = person_summary_description($MEMBER);
 $data['enter_leave'] = $MEMBER->getEnterLeaveStrings();
 $data['entry_date'] = $MEMBER->getEntryDate();
+$data['leave_date'] = $MEMBER->getLeftDate();
 $data['is_new_mp'] = $MEMBER->isNew();
 $data['other_parties'] = $MEMBER->getOtherPartiesString();
 $data['other_constituencies'] = $MEMBER->getOtherConstituenciesString();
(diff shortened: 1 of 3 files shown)


www/docs/mp/index.php
```diff
@@ -335,11 +336,24 @@
 # XXX This is current behaviour, but should probably now just be any recent MP
 $data['has_expenses'] = isset($MEMBER->extra_info['expenses2004_col1']) || isset($MEMBER->extra_info['expenses2006_col1']) || isset($MEMBER->extra_info['expenses2007_col1']) || isset($MEMBER->extra_info['expenses2008_col1']);
 
-// Set the expenses URL if we know it
-if (isset($MEMBER->extra_info['expenses_url'])) {
-    $data['expenses_url_2004'] = $MEMBER->extra_info['expenses_url'];
+if (strtotime($data['entry_date']) > strtotime('2010-05-05')) {
```
dracos: You don't need the `strtotime`s here, string comparison will be fine.

www/docs/mp/index.php
```diff
@@ -335,11 +336,24 @@
 # XXX This is current behaviour, but should probably now just be any recent MP
 $data['has_expenses'] = isset($MEMBER->extra_info['expenses2004_col1']) || isset($MEMBER->extra_info['expenses2006_col1']) || isset($MEMBER->extra_info['expenses2007_col1']) || isset($MEMBER->extra_info['expenses2008_col1']);
```
dracos: `has_expenses` is presumably out of date in that all new MPs it'll be false? As per comment, should probably be any MP with a leave date since 2004 sometime. That'll then simplify the new code as well I think.

www/docs/mp/index.php
```diff
@@ -335,11 +336,24 @@
...
+    // Set the expenses URL if we know it
+    $data['post_2010_expenses'] = False;
+    if (isset($MEMBER->extra_info['expenses_url'])) {
+        $data['pre_2010_expenses'] = True;
+        $data['expenses_url_2004'] = $MEMBER->extra_info['expenses_url'];
+    } else {
+        $data['pre_2010_expenses'] = True;
+        $data['expenses_url_2004'] = 'http://mpsallowances.parliament.uk/mpslordsandoffices/hocallowances/allowances%2Dby%2Dmp/';
+    }
+    if (strtotime($data['leave_date']) > strtotime('2010-05-05')) {
```
dracos: Ditto strtotime.

www/docs/mp/index.php
```diff
@@ -335,11 +336,24 @@
...
+    $data['post_2010_expenses'] = True;
 } else {
-    $data['expenses_url_2004'] = 'http://mpsallowances.parliament.uk/mpslordsandoffices/hocallowances/allowances%2Dby%2Dmp/';
+    // Set the expenses URL if we know it
+    $data['post_2010_expenses'] = False;
+    if (isset($MEMBER->extra_info['expenses_url'])) {
+        $data['pre_2010_expenses'] = True;
+        $data['expenses_url_2004'] = $MEMBER->extra_info['expenses_url'];
+    } else {
+        $data['pre_2010_expenses'] = True;
```
dracos: So pre is always set to True :)

www/docs/mp/index.php
```diff
@@ -333,13 +334,19 @@
 # Everyone who is currently somewhere has email alert signup, apart from current Sinn Fein MPs who are not MLAs
 $data['has_email_alerts'] = ($MEMBER->current_member_anywhere() && !($MEMBER->current_member(HOUSE_TYPE_COMMONS) && $MEMBER->party() == 'SF' && !$MEMBER->current_member(HOUSE_TYPE_NI)));
 # XXX This is current behaviour, but should probably now just be any recent MP
```
dracos: Can remove the XXX comment now :)

PR #1197, opened 2017-01-10
Update TWFY to use UTF-8

### LINKED ISSUE #815: Use UTF-8 charset in new HTML pages
This would probably incur significant developer time, but it feels like 2015 is the year TheyWorkForYou pages should get served with a UTF-8 content-type. (They're currently served as `iso-8859-1`.)

I mention it because earlier today I typed a curly quote into a TWFY template, and I got gobbledegook out of the browser, and it felt like I was back in 2005. Replacing every unicode character with HTML entities is all well and good, but I can't recall the last time I had to do it in any other mySociety product.

Feel free to `wontfix` if this isn't a priority, but @jacksonj04 said I should vent somewhere ;-)

struan:
This updates the site to output and store UTF-8 everywhere rather than ISO-8859-1.

Along with this branch the following changes will need to be made to the database:

```sql
ALTER database twfy default character set utf8mb4;
ALTER database twfy default collate utf8mb4_general_ci;
```

and then something like this for all the tables:

```sql
ALTER table member convert to character set utf8mb4;
```

This latter command also updates the collation on columns for those where we have overridden it.

Generate this for all table with:

```bash
#!/bin/bash

QUERY="select table_name from information_schema.TABLES T, information_schema.         COLLATION_CHARACTER_SET_APPLICABILITY CCSA WHERE character_set_name =         'latin1' AND CCSA.collation_name = T.table_collation AND T.table_schema =       'twfy';"

TABLES=$(mysql twfy_struan --skip-column-names -e "$QUERY")

for t in $TABLES; do
    echo "alter table $t convert to charset utf8mb4;"
done;
```

The one exception to this is the `mentions` table which we need to convert to innoDB and only change to `utf8` before this in order to get round limitations with mysql index sizes. Or just leave as it is as I don't think it's used any more and it doesn't contain anything that should be matter.

NB: we use `utf8mb4` as `utf8` only supports up to 3 bytes.

Fixes #815

### DIFF
--- a/www/docs/api/api_functions.php
+++ b/www/docs/api/api_functions.php
@@ -210,6 +210,8 @@ function api_output($arr, $last_mod=null) {
         $out = api_output_php($arr);
     } elseif ($output == 'rabx') {
         $out = api_output_rabx($arr);
+    } elseif ($output == 'json') {
+        $out = json_encode($arr, JSON_PRETTY_PRINT);
     } else { # JS
         header('Access-Control-Allow-Origin: *');
         $out = api_output_js($arr);
@@ -237,6 +239,9 @@ function api_header($o, $last_mod=null) {
         $type = 'text/php';
     } elseif ($o == 'rabx') {
         $type = 'application/octet-stream';
+    } elseif ($o == 'json') {
+        header('Access-Control-Allow-Origin: *');
+        $type = 'application/json';
     } else {
         header('Access-Control-Allow-Origin: *');
         $charset = 'iso-8859-1';
--- a/www/docs/api/api_functions.php
+++ b/www/docs/api/api_functions.php
@@ -204,7 +204,7 @@ function api_output($arr, $last_mod=null) {
         if ($cond) return;
     }
     if ($output == 'xml') {
-        $out = '<?xml version="1.0" encoding="iso-8859-1"?>'."\n";
+        $out = '<?xml version="1.0" encoding="utf-8"?>'."\n";
         $out .= '<twfy>' . api_output_xml($arr) . '</twfy>';
     } elseif ($output == 'php') {
         $out = api_output_php($arr);
--- a/www/docs/api/api_functions.php
+++ b/www/docs/api/api_functions.php
@@ -230,6 +230,7 @@ function api_header($o, $last_mod=null) {
             return true;
         }
     }
+    $charset = 'utf-8';
     if ($o == 'xml') {
         $type = 'text/xml';
     } elseif ($o == 'php') {
@@ -238,10 +239,11 @@ function api_header($o, $last_mod=null) {
         $type = 'application/octet-stream';
     } else {
         header('Access-Control-Allow-Origin: *');
+        $charset = 'iso-8859-1';
         $type = 'text/javascript';
     }
     #$type = 'text/plain';
-    header("Content-Type: $type; charset=iso-8859-1");
+    header("Content-Type: $type; charset=$charset");
     if ($last_mod>0)
         header('Last-Modified: ' . date('r', $last_mod));
     return false;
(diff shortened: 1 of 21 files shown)


www/includes/utility.php
```diff
@@ -596,11 +586,8 @@ function htmlentities_notags($text) {
     $text = strtr($text, $tbl);
 
     // This turns any out of bounds characters like fancy quotes
-    // into the ISO-8859-1 equivalent if possible.
-    $text = iconv('Windows-1252', 'ISO-8859-1//TRANSLIT', $text);
-
-    # Remove all illegal HTML characters (damn you, Windows-1252)
-    $text = preg_replace('/[\x80-\x9f]/', '', $text);
+    // into the UTF-8 equivalent if possible.
+    $text = iconv('Windows-1252', 'UTF-8', $text);
```
dracos: This will break UTF-8 text, because UTF-8 uses bytes from the Windows-1252 128-159 range. e.g. It would turn é (c3a9) into Ã© (c383c2a9). I don't think you need any iconv here at all.

www/includes/utility.php
```diff
@@ -893,6 +879,8 @@ function entities_to_numbers($string) {
 
 function make_member_url($name, $const = '', $house = HOUSE_TYPE_COMMONS, $pid = NULL) {
 
+    # convert name to ISO so we have sensible URLs
+    $name = iconv('UTF-8', 'ISO-8859-1//TRANSLIT', $name);
```
dracos: I don't understand what this is for, we want UTF-8 in URLs too :)

www/docs/api/api_functions.php
```diff
@@ -230,18 +232,23 @@ function api_header($o, $last_mod=null) {
...
+    $charset = 'utf-8';
     if ($o == 'xml') {
         $type = 'text/xml';
     } elseif ($o == 'php') {
         $type = 'text/php';
     } elseif ($o == 'rabx') {
         $type = 'application/octet-stream';
+    } elseif ($o == 'json') {
+        header('Access-Control-Allow-Origin: *');
+        $type = 'application/javascript';
```
dracos: This should be application/json.

www/docs/api/api_functions.php
```diff
@@ -204,12 +204,14 @@ function api_output($arr, $last_mod=null) {
...
     if ($output == 'xml') {
-        $out = '<?xml version="1.0" encoding="iso-8859-1"?>'."\n";
+        $out = '<?xml version="1.0" encoding="utf-8"?>'."\n";
         $out .= '<twfy>' . api_output_xml($arr) . '</twfy>';
     } elseif ($output == 'php') {
         $out = api_output_php($arr);
     } elseif ($output == 'rabx') {
         $out = api_output_rabx($arr);
+    } elseif ($output == 'json') {
+        $out = json_encode($arr, JSON_PRETTY_PRINT);
```
dracos: This presumably needs an Access-Control-Allow-Origin header too.

www/includes/utility.php
struan: We do, but I can't get the redirects to work with UTF-8 at the moment.

struan:
I think this is ready for another check over.

www/includes/utility.php
dracos: This still needs sorting – is the issue that `$s2` contains ISO-8859-1 characters and needs updating for UTF-8? Really, we should be able to use `iconv("utf-8", "ascii//TRANSLIT", $name)` but it looks like iconv versions might cause issues there, so probably easier to stick with the search and replace if that covers everything.

dracos: (Or alternatively, we could leave accents alone in the URLs now, we do on other sites. Few enough here, though.)

struan: The issue is that in Apache the redirects break when there are accented characters because it seems to cut the URL off at the accented character and you end up with a 404. I'm failed to get this to work as there didn't seem to be a way to get the regular expressions to handle the accented characters [:alpha:] for example didn't seem to do the right thing.

dracos: You will presumably need to `urlencode` the name if you're keeping the accents in the URL? You could replace all the hex in `s2` with the corresponding UTF-8 hex and then it would still ASCIIfy them all? Or do you mean you did that and that was what didn't work.

struan: ah right, I see what you mean now.

struan:
Following conversation with Matthew and some testing I don't think we need to make any changes to the database for this to work. It seems that marking the DB connection as being utf-8 is sufficient for everything to work as expected.

struan:
@dracos am I good to squash this down and bit and merge it?

dracos:
Yep, I approved it above :)

PR #1209, opened 2017-02-01
Use translation table rather than chr().

### LINKED ISSUE #1206: Including a search term on a debate page breaks encoding
e.g. https://www.theyworkforyou.com/wrans/?id=2015-01-12.220482.h is fine but https://www.theyworkforyou.com/wrans/?id=2015-01-12.220482.h&s=test is not.

### LINKED ISSUE #1208: UTF-8 rendering issue when highlighting is active
Example: [with highlighting](https://www.theyworkforyou.com/debates/?id=1996-01-16a.628.0&s=billion), [without highlighting](https://www.theyworkforyou.com/debates/?id=1996-01-16a.628.0)

An older debate contains the pound sterling symbol. The page is specified as UTF-8 and the sterling symbol is specified as `&#163;` which renders as expected. When phrase-highlighting is active the sterling symbol is delivered as "A3" (code 163) which is correct for ISO-8859-1 encoding, but probably displays an unknown character in a UTF-8 render.

dracos:
This appears to fix #1206 and fix #1208 – but I do not understand why I switched away from the translation table to the current behaviour in 2f9d9f1e35b8483aa467b53b85b17b3a6afb95d7 which is somewhat worrying… (search results do *appear* to still be working with this change, so that commit message is not helpful).

### DIFF
--- a/www/includes/easyparliament/searchengine.php
+++ b/www/includes/easyparliament/searchengine.php
@@ -83,8 +83,8 @@ class SEARCHENGINE {
         // Any characters other than this are treated as, basically, white space
         // (apart from quotes and minuses, special case below)
         // The colon is in here for prefixes speaker:10043 and so on.
-        $this->wordchars = "A-Za-z0-9,.'&:_\xc0-\xff";
-        $this->wordcharsnodigit = "A-Za-z0-9'&_\xc0-\xff";
+        $this->wordchars = "A-Za-z0-9,.'&:_\x80-\xbf\xc2-\xf4";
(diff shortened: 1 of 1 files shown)


www/includes/easyparliament/searchengine.php
```diff
@@ -482,7 +488,7 @@ public function highlight_internal($body, $stemmed_words) {
             return $body;
 
         # Does html_entity_decode without the htmlspecialchars
-        $body = preg_replace('/&#(\d\d\d);/e', 'chr($1)', $body);
+        $body = strtr($body, $this->translation_table);
```
dracos: This will only translate words, translation table doesn't contain numbers.

PR #1214, opened 2017-02-03
Show recent votes by an MP

struan:
As part of making it a bit easier to see recent things that have happened this adds a page to MP's profiles which show the most recent votes they have, or have not, participated in.

As ever it's only the ones that are associated with a policy we are interested in from PW so there will need to be some words about that, or another solution.

screen shot 2017-04-06 at 10 13 44

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -179,6 +179,32 @@ class Divisions {
     }
 
 
+    /**
+     * Get the last n votes for a member
+     *
+     * Returns an array of divisions
+     */
+    public function getRecentMemberDivisions($number = 20) {
+        $args = array(':person_id' => $this->member->person_id, ':number' => $number);
+        $q = $this->db->query(
+            "SELECT division_id, division_title, yes_text, no_text, division_date, vote, gid, direction
+            FROM policydivisions JOIN persondivisionvotes USING(division_id)
+            WHERE person_id = :person_id
+            GROUP BY division_id
+            ORDER by division_date DESC LIMIT :number",
+            $args
+        );
+
+        $divisions = array();
+        $row_count = $q->rows();
+        for ($n = 0; $n < $row_count; $n++) {
+          $divisions[] = $this->getDivisionDetails($q->row($n));
+        }
+
+        return $divisions;
+    }
+
+
     private function constructYesNoVoteDescription($direction, $title, $short_text) {
         $text = ' voted ';
         if ( $short_text ) {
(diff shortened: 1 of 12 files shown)


zarino:
@struan I’ve tidied up the styling on this a bit, and added a magellan nav to the sidebar.

classes/Divisions.php
```diff
@@ -179,6 +179,32 @@ public function getAllMemberDivisionsByPolicy() {
...
 
+    /**
+     * Get the last n votes for a member
+     *
+     * Returns an array of divisions
+     */
+    public function getRecentMemberDivisions($number = 20) {
+        $args = array(':person_id' => $this->member->person_id, ':number' => $number);
+        $q = $this->db->query(
+            "SELECT division_id, division_title, yes_text, no_text, division_date, division_number, vote, gid, direction
```
mhl: Not that it really matters, but it looks like you don't need the `division_number` column here.

www/docs/mp/index.php
```diff
@@ -494,6 +494,17 @@
 
         break;
 
+    case 'recent':
+        $policiesList = new MySociety\TheyWorkForYou\Policies;
+        $positions = new MySociety\TheyWorkForYou\PolicyPositions( $policiesList, $MEMBER );
+        $divisions = new MySociety\TheyWorkForYou\Divisions($MEMBER, $positions, $policiesList);
+
+        $data['divisions'] = $divisions->getRecentMemberDivisions();
+        //print_r($data['divisions']);
```
mhl: I guess this was left in when debugging, and could be removed...

www/docs/mp/index.php
```diff
@@ -494,6 +494,17 @@
...
+        //exit();
+
+        MySociety\TheyWorkForYou\Renderer::output('mp/recent', $data);
+      break;
```
mhl: The indentation here isn't consistent with the other case statements here (although I don't know what standard we follow for PHP - there's none mentioned here: http://mysociety.github.io/coding-standards.html )

www/includes/easyparliament/templates/html/mp/recent.php
```diff
@@ -0,0 +1,81 @@
...
+
+                    <div class="panel">
+                        <p>This person has not voted recently.</p>
+                    </div>
+
+                <?php } ?>
+
+
+
+                <div class="panel">
```
mhl: Since this same wording is also used on `divisions.php` and `votes.php`, perhaps in an earlier commit it could have been extracted to a partial?

PR #1217, opened 2017-02-10
WIP display last updated dates for policy sets

struan:
Should hopefully alleviate issues where we've not updated the
information since an important vote and hence the voting record
doesn't take it into account

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -37,6 +37,21 @@ class Divisions {
         $this->db = new \ParlDB;
     }
 
+    public static function getMostRecentDivisionDate() {
+      $db = new \ParlDB;
+      $q = $db->query(
+        "SELECT policy_id, max(division_date) as recent from policydivisions GROUP BY policy_id"
+      );
+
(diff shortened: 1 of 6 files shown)


classes/Divisions.php
```diff
@@ -37,6 +37,24 @@ public function __construct(Member $member, PolicyPositions $positions = NULL, P
...
+
+      $policy_maxes = array();
+      $policy_maxes['latest'] = '';
+      $row_count = $q->rows();
+      for ($n = 0; $n < $row_count; $n++) {
+        $policy_maxes[$q->field($n, 'policy_id')] = $q->field( $n, 'recent' );
+        if ($q->field($n, 'recent') > $policy_maxes['latest']) {
+          $policy_maxes['latest'] = $q->field($n, 'recent');
+        }
+      }
```
dracos: You could not have the `latest` bit inside and the loop and have `$policy_maxes['latest'] = max(array_values($policy_maxes));` after 🤷‍♂️

PR #1224, opened 2017-02-20
move MP EU referendum campaigning statement to profile

### LINKED ISSUE #1122: Add details on EU referendum stance to Profile section of MP page
At some point we will remove the current banner with the MP's stance on the EU referendum. It would be good to include this in the profile section if people want to refer back to it later.

struan:
remove the banner and move the statement down into the profile section
of the page.

Now looks like: 

screen shot 2017-02-20 at 16 26 38

Fixes #1122

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/profile.php
+++ b/www/includes/easyparliament/templates/html/mp/profile.php
@@ -41,34 +41,6 @@ $display_wtt_stats_banner = '2015';
             </div>
             <div class="primary-content__unit">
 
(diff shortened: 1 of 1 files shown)


struan:
@myf, @JenMysoc  - does this seem like a sensible thing to do now?

JenMysoc:
Sounds good to me :) I like the banner idea for notifying people of interesting votes so if we do decide on a way to signal votes of interest then we could maybe look at reinstating the banner for that!

PR #1227, opened 2017-02-27
new Topics pages

### LINKED ISSUE #1212: Redevelop topics pages
Topics pages should show
- Recent debates/divisions related to the topic

And redesign them generally to make them look a bit nicer/slicker

Part of #1210

### LINKED ISSUE #1213: Add topical topics to homepage
Add 1-5 links to topics as part of the 'in the news' section to link directly to relevant topics pages

Add admin mechanism to make these easy to change as often as daily

Part of #1210

struan:
This revamps the existing topics pages to make them both more functional and also by moving them into the database easier to edit and update. It also adds the ability to feature topics on the front page to make it easier to find them.

A topic consists of a title, a description, a banner image and then it can have:
* a suggested search term for setting up alerts
* associated debates in the Commons, Lords or Westminster Hall
* associated policy sets for which the votes will be shown along with how either the Commons as a whole voted or, if set, the user's MP voted.
* associated policies in order to construct, or augment, policy sets for a topic.

screen shot 2017-04-12 at 15 06 25
screen shot 2017-04-12 at 15 11 20
screen shot 2017-04-12 at 15 11 52

Fixes #1212 
Fixes #1213


conf/general-example
```diff
@@ -87,6 +87,9 @@ define ("INCLUDESPATH", BASEDIR . "/../includes/");
 // Web path to the directory where the site's images are kept.
 define ("IMAGEPATH", WEBPATH . "images/");
 
+// location for topic image uploads. ideally outside webtree
+define ("TOPICIMAGEPATH", "BASEDIR . "/../../topic_images");
```
zarino: Don’t think you want a `"` before `BASEDIR` here.

struan: doh, no I don't.

www/docs/topic/topic.php
```diff
@@ -17,146 +17,50 @@
...
 // Make sure the requested topic actually exists, otherwise throw a 404.
-if (isset ($topics[$topicname]))
+if ($topic = $topics->getTopic($topicname))
 {
 
-    // Set the actual topic data.
-    $data = $topics[$topicname];
-
+    $data = $topic->data();
+    $topic_policies = $topic->getAllPolicies();
```
dracos: Why isn't this included in a call to `data()`?
(`data` seems a bit loose, I guess is my underlying point - it appears to be acting as a sort of `extract` for the topic, but I wonder if it would be clearer if the template had a `topic` object it could use instead, shrug)

www/docs/topic/topic.php
```diff
@@ -17,146 +17,50 @@
...
     // Assume, unless we hear otherwise, that we don't want the postcode form displayed.
     $data['display_postcode_form'] = false;
+    $data['image'] = $topic->image_url();
```
dracos: This is confusing to me, as topic has an `image` function too (which is its filename, I think?). The admin template calls image_url if image is set, I think calling $topic->image_url in the template would be clearer to someone reading the template alone.

www/docs/topic/image.php
```diff
@@ -0,0 +1,29 @@
...
+}
+
+if (!$slug || !isset($topic) || $topic === NULL) {
+    header('HTTP/1.0 404 Not Found');
+    exit();
+}
+
+$path = $topic->image_path();
+
+if (!$path || !file_exists($path) || !is_file($path) || !is_readable($path)) {
```
dracos: Does is_readable not do the file_exists check and the string being blank for you?

www/docs/admin/topics.php
```diff
@@ -0,0 +1,89 @@
...
+            <thead>
+              <th>Name</th>
+              <th>Description</th>
+              <th>Front page</th>
+              <th></th>
+            </thead>
+            <?php foreach ($all_topics as $name => $topic) { ?>
+            <tr>
+              <td><a href="<?= $topic->url() ?>"><?= $topic->title() ?></a></td>
+              <td><?= $topic->description() ?></td>
```
dracos: Do these need escaping?

www/docs/admin/edittopic.php
```diff
@@ -0,0 +1,244 @@
...
+          <p>
+          <input type="submit" value="Update">
+          </p>
+        </form>
+
+
+        <h3>Related Content</h3>
+        <ul>
+          <?php foreach ($topic->getContent() as $content) { ?>
+          <li><a href="<?= $content['href'] ?>"><?= $content['title'] ?></a> <a class="delete" title="Remove" href="edittopic.php?action=deletecontent&id=<?= $topic->slug() ?>&content=<?= $content['id'] ?>">X</a></li>
```
dracos: A delete shouldn't be a GET request.

www/docs/admin/edittopic.php
```diff
@@ -0,0 +1,244 @@
...
+          <label for="description">Description</label>
+          <textarea id="description" rows="5" name="description"><?= $topic->description() ?></textarea>
+
+          <p>
+          <input type="submit" value="Save">
+          </p>
+        </form>
+
+        <h3>Set Image</h3>
+        <form enctype="multipart/form-data" action="edittopic.php" method="post">
```
dracos: Is there a reason you have to set the image separately from the other fields? Guess it's rare so doesn't really matter, just seemed a bit odd.

db/0011-add-topics-tables.sql
```diff
@@ -0,0 +1,29 @@
...
+CREATE TABLE `topic_policies` (
+  `topic_key` int(11) NOT NULL,
+  `policy_id` int(11) NOT NULL,
+  UNIQUE KEY `topic_policy` (`topic_key`, `policy_id`)
+);
+
+CREATE TABLE `topic_epobjects` (
+  `topic_key` int(11) NOT NULL,
+  `epobject_id` int(11) NOT NULL,
+  UNIQUE KEY `topic_object` (`topic_key`, `epobject_id`)
```
dracos: If you're going to link to epobject (rather than on gid, say), then you should have a foreign key reference to make it worth something, I guess!

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+
+    function slug() {
+        return $this->slug;
+    }
+
+    function set_slug($slug) {
+        $this->slug = $slug;
+    }
+
+    // FIXME
```
dracos: Not sure what the fixme is for :)

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+
+    function set_title($title) {
+        $this->title = $title;
+    }
+
+    function slug() {
+        return $this->slug;
+    }
+
+    function set_slug($slug) {
```
dracos: Probably not worth it here, and I haven't seen what other classes do, but do/could we consider having `slug` be a getter and a setter?

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+        }
+        if (isset($data['front_page'])) {
+            $this->front_page = $data['front_page'];
+        }
+
+        $this->raw['actions'] = array();
+    }
+
+
+    function data() {
```
dracos: I guess my other concern with `data` is that it is not updated when any `set` functions are called. It looks like the code *either* changes things or displays and uses `data`, so it's okay, but you could see in future something perhaps setting and fetching data assuming it would be up to date.

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+
+        $this->raw = $data;
+        if (isset($data['id'])) {
+            $this->id = $data['id'];
+        }
+        $this->title = $data['title'];
+        $this->slug = $data['slug'];
+        $this->description = $data['description'];
+        $this->image = $data['image'];
+        if (isset($data['search_string'])) {
```
dracos: As far as I can see, this constructor is always called with this and `front_page` being present, if you've got here.
And ID too in fact, it's never called with any data unless it's all of it.

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+
+    private $raw;
+    private $id;
+    private $title;
+    private $slug;
+    private $description;
+    private $image;
+    private $search_string;
+    private $front_page;
+    private $actions;
```
dracos: I don't think `$actions` is ever set?

classes/Topic.php
```diff
@@ -0,0 +1,339 @@
...
+        $this->description = $data['description'];
+        $this->image = $data['image'];
+        if (isset($data['search_string'])) {
+            $this->search_string = $data['search_string'];
+        }
+        if (isset($data['front_page'])) {
+            $this->front_page = $data['front_page'];
+        }
+
+        $this->raw['actions'] = array();
```
dracos: I don't think `actions` is ever used?

www/docs/admin/edittopic.php
struan: It's rare and it logically just seemed like a different operation to me.

bin/install-php5-xapian.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
```
struan: Partly I just wanted to make this as simple as possible and `adding site-specific-install.sh` wasn't helping with that as it only installs under nginx. So yes, it probably could but it was going to be a lot more work than I really wanted to put in to this at the time.

bin/install-as-user
```diff
@@ -49,7 +50,7 @@ cd "$REPOSITORY"
 echo -n "Creating conf/general configuration... "
 # Write sensible values into the config file:
 sed -r \
-    -e 's!^(.*"OPTION_TWFY_DB_USER", *)"[^"]*"!'"\\1'$UNIX_USER'!" \
+    -e 's!^(.*"OPTION_TWFY_DB_USER", *)"[^"]*"!'"\\1'twfy'!" \
```
struan: Er, no, not sure what that was about in retrospect.

db/0011-add-topics-tables.sql
struan: That would make it the first one in the database. Is there a reason there haven't been any till now?

struan:
This is ready for a second look now.

db/0011-add-topics-tables.sql
dracos: Gulp, dunno. Perhaps MySQL doesn't have them ;)

PR #1231, opened 2017-03-02
Future fetch utf8 tweaks

struan:
Handle odd UTF characters that upset the database.


scripts/future-fetch.py
```diff
@@ -45,6 +45,9 @@ class Entry(object):
     witnesses_str = ''
     location = ''
 
+    # strip 4 byte unicode chars plus zero width space
+    re_pattern = re.compile(u'(?:[^\u0000-\uD7FF\uE000-\uFFFF]|[\u200B])', re.UNICODE)
```
dracos: I think any character that won't transcode to CP1252 will cause an issue; it looks like we either need to put back the ISO-8859-1-and-charrefs encoding removed in https://github.com/mysociety/theyworkforyou/commit/87768f573b3351ccc77bbe3d16e5351db5b12fb2 or actually switch the DB column character sets as originally suggested in #1197.

struan: I think that converting the table to UTF-8 is going to be the easiest solution here as the character conversion option generates further problems when the conversion for the DB connection happens.

db/schema.sql
```diff
@@ -265,6 +265,7 @@ CREATE TABLE `future` (
   KEY `event_date` (`event_date`),
   KEY `modified` (`modified`)
 );
+ALTER TABLE `future` CONVERT TO CHARACTER SET 'utf8';
```
dracos: I think you can do this in the create table statement with `) CHARACTER SET 'utf8'`?

PR #1233, opened 2017-03-03
Fix A-Z links on MP list page for non last name sorting

### LINKED ISSUE #1060: A-Z links have surprising behaviour when list is sorted by first name
http://www.theyworkforyou.com/mps/?o=f#H takes you to the first person whose _last name_ starts with 'H', rather than those _first name_ does:

screen shot 2016-02-15 at 08 50 26

struan:
This fixes the A-Z links so they refer to first names when sorting by first name.
It hides the A-Z links when sorting by party as they aren't actually useful in that case.

Fixes #1060

### DIFF
--- a/www/includes/easyparliament/templates/html/people/index.php
+++ b/www/includes/easyparliament/templates/html/people/index.php
@@ -165,9 +165,14 @@ if (!count($data)) {
                 <?php } ?>
 
                 <div class="people-list">
-                <?php $current_initial = ''; ?>
-                <?php foreach ( $data as $person ) {
-                    $initial = substr( strtoupper($person['family_name']), 0, 1);
+                <?php
+                $current_initial = '';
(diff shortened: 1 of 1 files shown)


www/includes/easyparliament/templates/html/people/index.php
```diff
@@ -159,11 +160,17 @@
                     <a href="#Y">Y</a>
                     <a href="#Z">Z</a>
                 </div>
+                <?php } ?>
 
                 <div class="people-list">
-                <?php $current_initial = ''; ?>
-                <?php foreach ( $data as $person ) {
-                    $initial = substr( strtoupper($person['family_name']), 0, 1);
+                <?php
```
dracos: Could this not be output if order is party too? Given it's not used (and is outputting multiple times)

PR #1237, opened 2017-03-16
do not merge paragraphs after a quote from official record

### LINKED ISSUE #1195: Parlparse missing the end of quotes
e.g https://goo.gl/qSMuqb vs https://www.theyworkforyou.com/lords/?id=2017-01-09a.1754.3

struan:
Improve the merging of run-on paragraphs to detect when something looks
like the end of a quote from the official record by checking for a
paragraph ending with ]. This avoids formatting a follow on sentence
into the indent for the quote.

Fixes #1195

### DIFF
--- a/www/includes/easyparliament/templates/html/section/_section_content.php
+++ b/www/includes/easyparliament/templates/html/section/_section_content.php
@@ -51,7 +51,7 @@
         $body = preg_replace('#<p>\s*</p>#', '', $body);
 
         # Assume a paragraph starting with a lowercase character should be run on
-        $body = preg_replace('#(?<!:)</p>\s*<p[^>]*>(?=[a-z])(?![ivx]+\.)#', ' ', $body);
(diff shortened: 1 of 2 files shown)


www/includes/easyparliament/templates/html/section/_section_content.php
```diff
@@ -51,7 +51,7 @@
         $body = preg_replace('#<p>\s*</p>#', '', $body);
 
         # Assume a paragraph starting with a lowercase character should be run on
-        $body = preg_replace('#(?<!:)</p>\s*<p[^>]*>(?=[a-z])(?![ivx]+\.)#', ' ', $body);
+        $body = preg_replace('#(?<!:|\])</p>\s*<p[^>]*>(?=[a-z])(?![ivx]+\.)#', ' ', $body);
```
mhl: *looks up negative lookbehind docs*

Might it be easier to read with a character class `[:\]]` rather than alternation `:|\]`? (I think that's equivalent in this case?)

struan: Yup, however I missed a third case which is in the new commit so we do need an alternation now :|

PR #1239, opened 2017-03-16
Disable adding annotations

### LINKED ISSUE #1220: Freeze annotations
At the Leeds team meeting, we discussed removing the ability to add annotations, but still keeping older ones in place.

This would also negate the need for #692 and #771.

struan:
Removed code for adding annotations and also references to this in the user sign-up process.

Fixes #1220


abimysoc:
The House Rules themselves need updating/removing, no?

struan:
I am leaving them in there for a bit as they get referred to in the emails about reported annotations. The should probably go at some point in the future though.

mhl:
I think there might be some other places where annotation-related material could be removed, but ICBW:

* https://github.com/mysociety/theyworkforyou/blob/e399bb2e92ff9db46309ec27c4e0cea7f0c1bb18/classes/SectionView/SectionView.php#L383-L390

The reference to "leave an annotation" here:

* www/includes/easyparliament/templates/html/static/getinvolved.php

mhl:
I don't have much confidence in my testing, but other than the comment above this looks good to me. (I've tried on my local installation to check that existing annotations are still shown, and they are, and I can't see a way of leaving further annotations.)

struan:
I think the getinvolved page isn't used anymore. I've created #1245 to confirm that and remove them.

PR #1240, opened 2017-03-16
minimal sanity checking on interesting links before display

### LINKED ISSUE #1236: Link to Scottish Parliament broken
https://www.theyworkforyou.com/msp/14105/nicola_sturgeon

The link marked "Page on the Scottish Parliament website" just links back to the same TWFY page.

struan:
Make sure that interesting links on the profile page contain http
at the start before displaying them.

Fixes #1236

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/profile.php
+++ b/www/includes/easyparliament/templates/html/mp/profile.php
@@ -222,9 +222,12 @@ $display_wtt_stats_banner = '2015';
 
                     <ul class="comma-list">
 
-                        <?php foreach ($useful_links as $link): ?>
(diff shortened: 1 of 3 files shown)


mhl:
This change looks good, but out of interest, where does this `sp_url` URL come from? (I see that it's from `extra_info()` for a member, and that looks like it's loaded with `scripts/mpinfoin.pl` from parlparse's `members/websites-sp.xml`, but there's no entry for Nicola Sturgeon (`uk.org.publicwhip/person/14105`) in there?

struan:
I think that is another hand created file that looks like it's not been updated in 8 years :(

PR #1270, opened 2017-06-02
Change banner to link to GE2017 candidates.

annapowellsmith:
During the general election campaign, it would be awesome (and useful) if the banner could link to [Who Can I Vote For](https://whocanivotefor.co.uk), which shows people their candidates in the general election. This site is currently being served in the Facebook newsfeed 😄 

This PR pre-fills the user postcode if available (hopefully, I have not actually run the PHP locally), or if not just falls back to a generic message and link.

Here's how it looks...

screen shot 2017-06-02 at 20 16 12

### DIFF
--- a/www/includes/easyparliament/templates/html/header.php
+++ b/www/includes/easyparliament/templates/html/header.php
@@ -186,10 +186,10 @@
     </div>
   <?php } ?>
 
-    <div class="banner" id="ms_srv_wrapper" style="display:none">
+    <div class="banner">
         <div class="full-page__row">
             <div class="banner__content">
-                <a href="https://www.surveygizmo.com/s3/3539768/TheyWorkForYou-Survey" id="ms_srv_link" data-transaction="report" target="_blank">We're running a survey to help make this site better; to take it click here.</a>
+                <a href="https://whocanivotefor.co.uk/<?php if ($user_postcode) { echo "elections/" . _htmlentities(urlencode($user_postcode)); } ?>" target="_blank">New! See your 2017 General Election candidates <?php if ($user_postcode) { echo " in " . $user_postcode; } ?></a>
             </div>
         </div>
     </div>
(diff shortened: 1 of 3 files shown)


www/includes/easyparliament/templates/html/header.php
```diff
@@ -186,10 +186,10 @@ function trackLinkClick(link, category, name, value) {
     </div>
   <?php } ?>
 
-    <div class="banner" id="ms_srv_wrapper" style="display:none">
+    <div class="banner">
         <div class="full-page__row">
             <div class="banner__content">
-                <a href="https://www.surveygizmo.com/s3/3539768/TheyWorkForYou-Survey" id="ms_srv_link" data-transaction="report" target="_blank">We're running a survey to help make this site better; to take it click here.</a>
+                <a href="https://whocanivotefor.co.uk/<?php if ($user_postcode) { echo "elections/" . _htmlentities(urlencode($user_postcode)); } ?>" target="_blank">See your 2017 General Election candidates <?php if ($user_postcode) { echo " in " . $user_postcode; } else { echo " here"; } ?></a>
```
dracos: $user_postcode only exists on an MP page, I'm afraid – I think you'll need to set a key (I guess `user_postcode` makes sense!) containing the postcode on `$this->data` in `classes/Renderer/User.php` and then it'll be available in that template.

annapowellsmith:
Thanks for the review. I've made the suggested fix though again, can't run the PHP locally so let me know if any issues!

dracos:
Looks good :) The single test failure is from tests/PageTest.php testing that a banner isn't normally shown unless banner text is set – I guess while this banner is in operation you could just short-circuit and return from that test? Think that's all that's needed :)

annapowellsmith:
Ah sorry, I only read your comment after deleting the tests! Is it better to re-add them and return, as you suggest?

dracos:
No worries, same outcome either way!

dracos:
Removed the last paragraph and word-wrapped commit message. Thank you :)

dracos:
Tweaked to use canonicalisePostcode.

annapowellsmith:
Thank you!

PR #1272, opened 2017-06-09
Remove mySociety header

### LINKED ISSUE #1246: Contact page still too hard to find
The problem described in #1053 still happens, because not everyone has ginormous wide screens, which means:

screenshot from 2017-03-29 09-18-42

they still can't find the contact link that we added to the header; the mySociety logo top right is still the most obvious click target for anyone looking for how to get in touch with this site. Perhaps it should link to the contact page (on which we could add "TheyWorkForYou is run by <a href="https://www.mysociety.org/">mySociety</a>, a UK charity." above the `h1`).

zarino:
We think people are clicking the mySociety logo in the header in an attempt to contact their MP (and thereby causing lots of erroneous emails to user support via the mysociety.org contact form). Removing the logo from the header should make the *real* path for contacting your MP more obvious.

Fixes #1246.

Here’s how it looks on large screens, with and without the temporary survey banner:

screen shot 2017-06-09 at 10 35 20

screen shot 2017-06-09 at 10 35 42

And here’s how it looks on medium screens:

screen shot 2017-06-09 at 10 36 05

screen shot 2017-06-09 at 10 35 58


abimysoc:
> in an attempt to contact their MP

For the record, no; usually (judging by the content of the emails) in an attempt to contact this website.

PR #1296, opened 2017-07-06
Facebook login

struan:
This allows a user to login using facebook rather than with a username/password.

It stores the user's Facebook ID and an access token in the database and then uses the access token in the database to check the login cookie.

Still to do:

- [x] update the user details page
- [x] we might want to ask for a postcode after login
- [x] Make it look nice (@zarino)
- [x] better error handling
- [x] tidy up the code a bit


www/includes/easyparliament/user.php
```diff
@@ -1190,6 +1321,63 @@ public function confirm($token) {
...
+            if ($q->field(0, 'postcode')) {
+                try {
+                    $MEMBER = new MEMBER(array('postcode'=>$q->field(0, 'postcode'), 'house'=>1));
+                    $pid = $MEMBER->person_id();
+                    # This should probably be in the ALERT class
+                    $this->db->query('update alerts set confirmed=1 where email = :email and criteria = :criteria', array(
+                            ':email' => $this->email,
+                            ':criteria' => 'speaker:' . $pid
+                        ));
+                } catch (MySociety\TheyWorkForYou\MemberException $e) {
```
jacksonj04: Empty `catch(){}`, is this right?

www/includes/easyparliament/user.php
```diff
@@ -70,6 +70,7 @@ class USER {
     public $optin = "";            // boolean - Do they want emails from us?
     public $deleted = "";          // User can't log in or have their info displayed.
     public $confirmed = '';        // boolean - Has the user confirmed via email?
+    public $facebook_id = '';      // Facebook ID for users who login with FB
```
jacksonj04: We should declare `$facebook_token` here as well.

jacksonj04:
\o/ for test passes!

I'm happy the code does what it says it should, and that some of the less-than-ideal fragments will eventually be cleared up in a refactoring.

:shipit:

PR #1318, opened 2017-10-11
Video on homepage

### LINKED ISSUE #1315: Add video - on homepage?
Is there an opportunity to add our video to TWFY? https://www.youtube.com/watch?v=L_19GumEQCM&t=1s

(Obviously, people already on TWFY don't need to know about TWFY, but it does give them the opportunity to share the site with other friends).

Where would be the best place for this to fit in without spoiling the design?

zarino:
Does what it says on the tin. Fixes #1315.

screen shot 2017-10-11 at 12 20 53

@wrightmartin @MyfanwyNixon do you think the video should be bigger?


MyfanwyNixon:
I think the size looks ok as far as I can judge it!

PR #1321, opened 2017-10-31
WIP: Get tests passing on PHP7

jacksonj04:
Makes the necessary changes to get tests passing on PHP 7 on Travis, more specifically:

* Do not attempt to overload `random_bytes()` if the function already exists.
* Conditionally use `memcached.so` instead of `memcache.so` on PHP 7 in Travis configuration.
* The `TWFY\Memcache` class now detects the presence of `\Memcached` and uses that (with appropriate syntax changes) where necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ php:
   - "5.5"
   - "5.6"
   - "7.0"
+  - "7.1"
 
 matrix:
   allow_failures:
-    - php: "7.0"
+    - php: "7.1"
 
 services:
   - memcached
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_script:
  - mysql -e "create database IF NOT EXISTS twfy_test;" -uroot
  - mysql -u root twfy_test < db/schema.sql
  - php composer.phar install --no-interaction --prefer-source
- - bash -c 'if [ "$TRAVIS_PHP_VERSION" == "7.0" ]; then echo "extension = memcached.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; else echo "extension = memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;'
+ - bash -c 'if [ "$TRAVIS_PHP_VERSION" == "7.0" ] || [ "$TRAVIS_PHP_VERSION" == "7.1" ]; then echo "extension = memcached.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; else echo "extension = memcache.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;'
```
struan: I wonder if this could be a >= but bash so not sure.

dracos: I think it would be: `if [ "$TRAVIS_PHP_VERSION" -ge "7.0" ]`

jacksonj04: Turns out bash can't do floating point comparison, but we can push it through `bc` and achieve the same result.

PR #1323, opened 2017-10-31
Move Facebook SDK to use Packagist/Composer

jacksonj04:
Swaps out loading the Facebook SDK via git submodule to using Composer and Packagist instead.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "foundation"]
 	path = foundation
 	url = https://github.com/zurb/foundation.git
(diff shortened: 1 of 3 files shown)


jacksonj04:
Not sure if deploying will automatically tidy up submodules. @sagepe, any thoughts on the matter?

sagepe:
Short answer, no. The deploy includes `git submodule update --init` but doesn't do anything particularly clever to clean up old submodules.

jacksonj04:
When I deploy it (not at 7:30pm on a Friday) I’ll spot-clean the old submodule.

PR #1328, opened 2017-11-02
Autoload Refactoring 7: editqueue.php

jacksonj04:
Part of [Refactoring Part 2](https://github.com/mysociety/theyworkforyou/projects/1)

Prerequisite: #1327


struan:
I think this, and the preceding ones look fine.

PR #1330, opened 2017-11-17
add missing pg value to user edit page

struan:
Also add some basic tests for the /user pages

### DIFF
--- a/classes/User.php
+++ b/classes/User.php
@@ -55,10 +55,10 @@ class User {
         return $data;
     }
 
-    public function getUpdateDetails($this_page, $is_facebook_user = False) {
+    public function getUpdateDetails($this_page, $user) {
         $details = array();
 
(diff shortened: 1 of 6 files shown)


classes/User.php
```diff
@@ -56,6 +56,7 @@ public function getUserDetails($user_id = False) {
     }
 
     public function getUpdateDetails($this_page, $is_facebook_user = False) {
+        global $THEUSER;
```
jacksonj04: Can we avoid a `global` here and pass `$THEUSER` in from wherever the function is called? It looks like it's only used in `www/docs/user/index.php`, where `$THEUSER` is in scope already so no need for massive refactoring to inject it all the way down.

dracos:
Could we expand the commit message? It says it's adding a pg line but then there are other changes to User.php that aren't explained.

PR #1340, opened 2017-12-06
New header/nav layout

### LINKED ISSUE #1275: Improve the nav at medium screen widths
See #1053, #1246, #1272 for previous discussions: people contact mySociety instead of TheyWorkForYou because for some reason that path is easier than the right one, and perhaps this is the reason.

Full nav:

wide

But many people don't have a screen that wide:

medium

An always-visible 'Contact' or 'Help' link might be a good thing to try next. WriteToThem and FixMyStreet both have that.

zarino:
Fixes #1275.

# Narrow (closed)

narrow-closed

# Narrow (open)

narrow-open

# 768px – eg portrait iPad

768px

# 1024px – eg landscape iPad

1024px

# Wide

wide


abimysoc:
Seeing this again with fresh eyes, I do wonder if the right place for "Contact" (maybe even "Help/Contact"?) in the menu is with "Sign in" and "Join" - ways that the user can interact with the website, rather than the list of the website's content.

zarino:
> I do wonder if the right place for "Contact" (maybe even "Help/Contact"?) in the menu is with "Sign in" and "Join"

@abimysoc – Good idea! I’ve taken a look at how to move the link, and I can’t wrap my head around how to move it. @jacksonj04 – perhaps the two of us could take a look together?

zarino:
Ok, the "Contact" link has now been moved up, next to "Sign in / Join".

struan:
This doesn't handle being signed in and having a regional assembly selected very well:

screen shot 2017-12-12 at 09 55 54

Also, in that case the Contact and Your MP links read a bit too much like 'Contact Your MP' which I can see causing us some issues :(.

zarino:
@struan thanks for spotting that layout issue. I’ve added a commit that aims to prevent the search box and user links colliding on regional assembly pages, and also reorganised those user links so that "Contact" and "Your MP" are no longer next to each other.

zarino:
@struan Yeah, keeping the nav on a single line in that 768px–1024px range will be ticky without more special-case code. I think we deploy what we’ve got here, and revisit later.

PR #1359, opened 2018-05-21
Remove alerts from database when manually deleted.

dracos:


tests/AlertsTest.php
```diff
@@ -58,48 +58,14 @@ public function testFetchConfirmedUndeleted()
...
-
-        // Make sure the response has the correct attributes
-        $this->assertEquals(0, $response['data'][0]['confirmed']);
-        $this->assertEquals(1, $response['data'][0]['deleted']);
-    }
-
-    /**
-     * Test that confirmed, deleted tests are correctly retrieved
-     */
-    public function testFetchConfirmedDeleted()
```
crowbot: If bounces are still going to result in alerts that are flagged as deleted, do we want to keep this test?

www/includes/easyparliament/alert.php
```diff
@@ -413,7 +413,7 @@ public function confirm($token) {
     // If all goes well the alert will be flagged as deleted.
```
crowbot: Comment should probably reflect that the alert is now actually deleted rather than just flagged.

tests/AlertsTest.php
dracos: No code ever calls `fetch` like that, so it seemed a test wagging the code.

www/includes/easyparliament/alert.php
dracos: Oh yes, dur.

PR #1362, opened 2018-05-23
Restart Apache when booting Vagrant VM

zarino:
No idea why this is needed, but without it, whenever I start the VM and SSH in, Apache isn’t running, so I have to run it manually.

I’ve had this fix hanging around, unstaged, in my git directory for months. Figured I should finally see whether we want to merge it into `master` 😉


zarino:
@dracos – ah! If that’s the reason it’s happening, I’ll update the comment to make that clearer, then merge it in.

PR #1369, opened 2018-06-05
Fixes #665 Convert remainder of house-based queries to use house constants

### LINKED ISSUE #665: Convert remainder of house-based queries to use house constants.
Following from #637, a lot of the queries are still using integers for determining the house in question, rather than the sitewide constants.

antonionoca:
Replaced magic numbers with constants.
Updated SQL queries.
Verified that tests are passing.

### DIFF
--- a/www/includes/dbtypes.php
+++ b/www/includes/dbtypes.php
@@ -3,11 +3,8 @@
 // $Id: dbtypes.php,v 1.8 2009-04-28 13:13:04 matthew Exp $
 
 // The type field in the epobject database table
-/*
-$eptype = array(
-    1 => "hansard",
-    2 => "glossary" );
-*/
+define ('EPTYPE_HANSARD', 1);
+define ('EPTYPE_GLOSSARY', 2);
 
 // The major field in the hansard database table
 $hansardmajors = array(
--- a/classes/EditQueue.php
+++ b/classes/EditQueue.php
@@ -153,20 +153,22 @@ class EditQueue {
             // entries in different tables.
             switch ($data['epobject_type']) {
 
-                // glossary item
-                case 2:
+                case EPTYPE_GLOSSARY:
                     $previous_insert_id = $q->insert_id();
-                    $q = $this->db->query("INSERT INTO glossary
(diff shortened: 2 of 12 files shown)


www/includes/dbtypes.php
```diff
@@ -164,6 +164,9 @@
 );
 */
 
+// epobject types
+define ('GLOSSARY_ITEM', 2);
```
dracos: This could go right at the top of the file, where there is a comment about `eptype` that this could replace.

www/includes/easyparliament/member.php
```diff
@@ -21,7 +21,7 @@ class MEMBER {
     public $extra_info = array();
     // Is this MP THEUSERS's MP?
     public $the_users_mp = false;
-    public $house_disp = 0; # Which house we should display this person in
+    public $house_disp = HOUSE_TYPE_ROYAL; # Which house we should display this person in
```
dracos: Whilst I see the reason for this, this is actually an undefined value, rather than the royal house constant. It should probably not have a default here.

www/includes/easyparliament/member.php
```diff
@@ -502,7 +502,7 @@ public function party_text($party = null) {
         }
     }
 
-    public function entered_house($house = 0) {
+    public function entered_house($house = HOUSE_TYPE_ROYAL) {
```
dracos: Similarly, here (and `current_member` below) should really have `$house = null` or similar, royal should not be the default, this is a legacy issue.

antonionoca:
Thanks for the feedback. These are good suggestions. I'll apply them.

www/includes/easyparliament/member.php
antonionoca: I'll revert this change back to setting house_disp to zero. I'm not confident to set it to undefined until I add unit tests, I'd better not introduce null pointer exceptions :)

www/includes/easyparliament/member.php
antonionoca: same here, I'll revert this change

dracos:
Great, thanks very much! I've now merged this.

PR #1373, opened 2018-06-27
New layout for person page header, including social links

### LINKED ISSUE #1133: Do we still include twitter links?
Have Twitter feeds fallen off MPs' profiles? #97 shows that they were once included, but a few obvious contenders I've checked do not have the links in their profiles. (Anne Snelgrove, a la #101 still has).

zarino:
Fixes #1133.

To do:

- [x] Decide whether to deploy with small images, or wait until large images have been reimported to TWFY.

# Mobile

screenshot_2018-06-27 victoria atkins mp louth and horncastle - theyworkforyou

# Desktop, _with_ the larger photos commit a58100c

screenshot_2018-06-27 diane abbott mp hackney north and stoke newington - theyworkforyou

# Here’s how it’d look for an MP with only a tiny photo
screenshot_2018-06-27 mark field mp cities of london and westminster - theyworkforyou


www/docs/mp/index.php
```diff
@@ -219,37 +219,28 @@
...
-
-$current_offices = $MEMBER->offices('current', TRUE);
-$former_offices = $MEMBER->offices('previous', TRUE);
-
-// If this person has named non-committee offices, they override the default
-if (count($current_offices) > 0) {
-    $position = $current_offices[0];
+$role = '';
+foreach( array(
+    'Peer' => HOUSE_TYPE_LORDS,
```
dracos: This will produce output such as "Former Bishop Peer" which makes little sense.

www/docs/mp/index.php
```diff
@@ -219,37 +219,28 @@
...
+    'MSP' => HOUSE_TYPE_SCOTLAND,
+) as $noun => $house) {
+    if ($MEMBER->house($house)) {
+        if (!$MEMBER->current_member($house)) {
+            $role .= 'Former ';
+        }
+        if ($MEMBER->party_text()) {
+            $role .= $MEMBER->party_text() . ' ';
+        }
+        $role .= $noun;
```
dracos: This produces text such as "Former DUP MP DUP MLA" or "Crossbench Peer Former Labour MP" or "Conservative MP Conservative MSP"; think it needs some punctuation as the removed code did. Also on the output, it'll stick the current constituency on the end, which might not match the last element in this list (nothing says it's putting former ones first, I mean).

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -416,6 +416,30 @@
                             <p>This data was produced by TheyWorkForYou from a variety
                                 of sources. Voting information from
                                 <a href="http://www.publicwhip.org.uk/mp.php?id=uk.org.publicwhip/member/<?= $member_id ?>&amp;showall=yes">Public Whip</a>.</p>
+
+                          <?php if($image && $image['exists']) {
+                              if(isset($data['photo_attribution_text'])) { ?>
```
dracos: Does this mean there are pages (e.g. voting record) where a photo is displayed without attribution? Could this be bottom of all pages?

www/docs/mp/index.php
zarino: Aha, I see elsewhere in this file, we specially handle Bishops so they don’t get treated as a party – https://github.com/mysociety/theyworkforyou/blob/98134134cb0730e66aea4b06c9a6654572b7bbf3/www/docs/mp/index.php#L814 – I’ll try that here too.

www/includes/easyparliament/templates/html/mp/profile.php
zarino: I’ve now added it to `_vote_footer.php` as well, so it makes its way onto the other `/mp` subpages. 👍

zarino:
@dracos Thanks for the feedback. 

* I’ve replicated the profile photo attribution at the bottom of all the MP subpages now
* I’ve limited the header to _only_ show either the _current_ membership, or, if nothing is current, the _most recently ended_ membership. Obviously the person’s full membership history is available further down the profile page.
* Oh, and I’ve moved all that logic into the Member class, rather than the `mp/index.php` controller, so that we can potentially reuse it elsehwere on the site, like on the MPs/Lords list pages, or on the search results page.

So, please do have a poke around, and let me know if those changes aren‘t suitable.

Do you think you’ll have time to update TWFY’s image cache to use _larger_ versions of the "new" Parliament photos, before we deploy this? Or should we go ahead and deploy _without_ commit 17073ef538793cbbbd4e3c53c49b51be7ce4069c?

PR #1377, opened 2018-07-11
Update API to handle payment via Stripe.

dracos:
There is also a little bit of code in puppet to add TWFY as a varnish-api-key handling thing, as with MapIt.

### DIFF
--- a/www/includes/easyparliament/templates/emails/api_cancelled.txt
+++ b/www/includes/easyparliament/templates/emails/api_cancelled.txt
@@ -1,0 +1,12 @@
+Subject: Your subscription to the TheyWorkForYou API has been cancelled
+Hi,
+
+I'm afraid that when we tried to take payment for your subscription to the
+TheyWorkForYou API, it failed again. We've cancelled your subscription, so
+you are now back on the free tier. If you wish to resubscribe, please just
+log in to TheyWorkForYou and pick a new plan:
+    https://www.theyworkforyou.com/api/key
+
+Yours,
+mySociety
+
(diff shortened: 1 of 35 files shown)


www/includes/easyparliament/templates/html/api/update.php
```diff
@@ -0,0 +1,122 @@
...
+If you update your plan below, it will be reactivated.
+</p>
+<?php } ?>
+
+<form id="signup_form" method="post" action="/api/update-plan" autocapitalize="off">
+
+
+<div class="row">
+    <label for="id_plan_0">Please choose a plan:</label>
+    <ul id="id_plan">
```
struan: We should say this is ex VAT as I was confused that it said 20 here and then 24 later on in the process.

dracos: Ah, that's my mistake setting it up. It will be inc VAT and should be the same in both places, but I mismatched the pricing in test Stripe.

dracos:
"Also, if I create a £100 plan, then cancel it and then create a £20 plan I move from having 100k credits to 10k credits which seems confusing." - hopefully an unlikely occurrence, we can always refund/deal with manually if someone ever does that. But it looks like it prorated correctly given all that, and so you will only be paying for that 10k, if you see what I mean? Or have I misunderstood :)

www/includes/easyparliament/templates/emails/api_cancelled.txt
```diff
@@ -0,0 +1,11 @@
+Subject: Your subscription to the TheyWorkForYou API has been cancelled
+Hi,
+
+I'm afraid that when we tried to take payment for your subscription to the
+TheyWorkForYou API, it failed again. We've cancelled your subscription, so
+you are now back on the free tier. If you wish to resubscribe, please just
+log in to TheyWorkForYou and pick a new plan.
```
struan: I wonder if including a link to the api key page would be useful here?

PR #1389, opened 2018-11-28
WIP better division results display

### LINKED ISSUE #746: Division results design should be looked at
I did a bit of work on the list of Yeas and Nays in division results but a quick glance over from a designer to check/advice would be good.

Example at http://www.theyworkforyou.com/debates/?id=2015-02-03e.160.0#g187.0

struan:
Fixes #746
Connects to https://github.com/mysociety/community/issues/41


www/includes/easyparliament/templates/html/divisions/vote.php
```diff
@@ -11,146 +10,101 @@
...
+                          <h3>How your MP voted</h3>
+                          <p>
+                              Based on postcode <strong><?= $data['mp_data']['postcode'] ?></strong>
+                              <a href="<?= $data['mp_data']['change_url'] ?>">(Change postcode)</a>
+                          </p>
+                        </div>
+                      <?php } ?>
+                        <a href="<?= $member->url() ?>" class="your-mp__content">
+                          <?php if ( isset($mp_vote) ) { ?>
+                            <span class="your-mp__vote your-mp__vote--<?= $mp_vote['vote'] ?>"><?php
```
dracos: This and the identical code in section_content should be factored into a shared template.

www/includes/easyparliament/templates/html/divisions/vote.php
```diff
@@ -11,146 +10,101 @@
...
+                                      break;
+                                  case 'absent':
+                                      echo 'Absent';
+                                      break;
+                                  case 'both':
+                                      echo 'Abstain';
+                                      break;
+                                  case 'tellaye':
+                                  case 'tellno':
+                                      echo 'Teller';
```
dracos: We seem to have lost the information as to whether they were teller for the ayes or the noes.

www/includes/easyparliament/templates/html/divisions/vote.php
```diff
@@ -11,146 +10,101 @@
...
+                            ?></span>
+                          <?php } else if ( isset($before_mp) || isset($after_mp) ) { ?>
+                            <span class="your-mp__vote">N/A</span>
+                          <?php } ?>
+                            <img class="your-mp__image" src="<?php $img = $member->image(); echo $img['url']; ?>">
+                            <div class="your-mp__person">
+                                <h2 class="people-list__person__name"><?= $member->full_name() ?></h2>
+                                <p class="people-list__person__memberships">
+                                    <span class="people-list__person__constituency"><?= $member->constituency() ?></span>
+                                    <span class="people-list__person__party <?= strtolower( $member->party() ) ?>"><?= $member->party() ?></span>
```
dracos: This should use slugify.

www/includes/easyparliament/templates/html/divisions/vote.php
dracos: Oh I see how the member data comes in differently, sigh; well could do the switch and then leave the multi-object member/user issue to later/never...

www/includes/easyparliament/hansardlist.php
```diff
@@ -2309,7 +2319,16 @@ public function _get_data_by_gid($args) {
 
         // We need to get the data for this gid.
         // Then depending on what htype it is, we get the data for other items too.
-        global $DATA, $this_page, $hansardmajors;
+        global $DATA, $this_page, $hansardmajors, $MEMBER, $THEUSER;
+
+        // use this for generating member vote data
+        if (isset($THEUSER) && $THEUSER->postcode_is_set()) {
```
dracos: I wonder if this should instead go in the www/docs/section.php file, so it's set up outside this class (which shouldn't rreally need to know about it here).

PR #1390, opened 2018-11-29
Factor out policydivisions table into two tables

dracos:
Consequential updates.

### DIFF
--- a/db/0017-split-up-policies-divisions.sql
+++ b/db/0017-split-up-policies-divisions.sql
@@ -1,0 +1,35 @@
+CREATE TABLE `divisions` (
+  `division_id` varchar(100) NOT NULL,
+  `house` varchar(100),
+  `gid` varchar(100) default '',
+  `division_title` text NOT NULL,
+  `yes_text` text,
+  `no_text` text,
+  `division_date` date NOT NULL default '1000-01-01',
+  `division_number` int(11),
+  `yes_total` int(3) default 0,
+  `no_total` int(3) default 0,
+  `absent_total` int(3) default 0,
+  `both_total` int(3) default 0,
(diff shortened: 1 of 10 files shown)


db/0017-split-up-policies-divisions.sql
```diff
@@ -0,0 +1,35 @@
...
+SELECT DISTINCT
+    division_id,
+    house, gid, division_title,
+    yes_text, no_text, division_date, division_number,
+    yes_total, no_total, absent_total, both_total, majority_vote
+    FROM policydivisions;
+
+ALTER TABLE policydivisions DROP
+    house, gid, division_title,
+    yes_text, no_text, division_date, division_number,
```
struan: It turns out mysql requires you to put `DROP` in front of every column name so this should be

```
ALTER TABLE policydivisions DROP
    house, DROP gid, DROP division_title...
```

scripts/json2db.pl
```diff
@@ -35,10 +35,13 @@
 my $policycheck = $dbh->prepare("SELECT policy_id from policies where policy_id = ?");
 my $policyadd = $dbh->prepare("INSERT INTO policies (policy_id, title, description) VALUES (?, ?, ?)");
 
-my $motioncheck = $dbh->prepare("SELECT division_title, gid, direction, policy_vote, yes_text, no_text, yes_total, no_total, absent_total, both_total, majority_vote FROM policydivisions WHERE division_id = ? AND policy_id = ?");
+my $divisioncheck = $dbh->prepare("SELECT division_title, gid, yes_text, no_text, yes_total, no_total, absent_total, both_total, majority_vote FROM policydivisions WHERE division_id = ?");
```
dracos: Wrong table name here, fixed.

PR #1391, opened 2018-11-29
Import divisions into database

dracos:


scripts/import_old_divisions.pl
```diff
@@ -0,0 +1,364 @@
...
+    exit;
+}
+
+if ($datefrom || $dateto) {
+    $datefrom = "1000-01-01" if !defined $datefrom;
+    $dateto = "9999-12-31" if !defined $dateto;
+} else {
+    $datefrom = "9999-12-31";
+    $dateto = "1000-01-01";
+}
```
struan: Are these not the wrong way round?

scripts/import_old_divisions.pl
```diff
@@ -0,0 +1,364 @@
...
+            die unless $vote_direction eq $side;
+            $vote_counts_by_pid{$person_id}++;
+        }
+    }
+    foreach (keys %vote_counts_by_pid) {
+        $totals->{both}++ if $vote_counts_by_pid{$_} > 1;
+    }
+
+    # Okay, now construct HTML and add to database
+    foreach my $list (@lists) {
```
struan: This looks like an out of date comment :)

scripts/import_old_divisions.pl
dracos: It's the same in xml2db.pl, which there is because you could be using `--recent`, and $use could be false and then you definitely *wouldn't* want anything to match. But here, no, doesn't make sense, does it!

PR #1393, opened 2018-11-30
Store debate heading with division.

dracos:


scripts/import_old_divisions.pl
```diff
@@ -265,17 +267,26 @@ sub load_debate_division {
     }
 
     my $majority_vote = $totals->{aye} > $totals->{no} ? 'aye': 'no';
-    $divisionupdate->execute($division_id, $house, "Division No. $divnumber", $divdate, $divnumber, $gid,
+    my $title = division_title($divnumber);
+    $divisionupdate->execute($division_id, $house, $title, $divdate, $divnumber, $gid,
```
struan: Is this not going to replace any division titles that have been pulled in from PW? Of course, it will then be reset again by the pw import script but we should possibly avoid doing that, somehow...

struan:
I also realised when checking this that `json2db.pl` has mismatched bind values on line 43:

`my $motionadd = $dbh->prepare("INSERT INTO policydivisions (division_id, policy_id, direction, policy_vote) VALUES (?, ?, ?, ?, ?)");`

4 values vs 5 ?

PR #1398, opened 2018-12-07
Add Delete All, and swap columns

### LINKED ISSUE #737: Improve UX of alert management
OK, effectively if you delete all your alerts, you have unsubscribed? But it may not feel like it. User comment:

"Because of a major change in job and interests, I need to unsubscribe. But there appears to be no way to do this (I clicked the link at the foot of the page but this takes you to a "manage subscriptions" page).

An unsubscribe button should do that, and quickly, or you are in danger of being seen as a spammer (not true, obviously)."

dracos:
This reimplements the old branch and fixes #737.

### DIFF
--- a/www/includes/easyparliament/templates/html/alert/index.php
+++ b/www/includes/easyparliament/templates/html/alert/index.php
@@ -257,34 +257,7 @@
               <?php if ($email_verified) { ?>
 
                   <?php if ( $alerts ) { ?>
-                    <h3>Your current email alerts</h3>
-
-                    <ul class="alerts-manage__list">
-                      <?php foreach ( $alerts as $alert ) { ?>
-                        <li>
-                            When <?= _htmlspecialchars($alert['criteria']) ?>.
-                            <form action="<?= $actionurl ?>" method="POST">
-                                <input type="hidden" name="t" value="<?= _htmlspecialchars($alert['token']) ?>">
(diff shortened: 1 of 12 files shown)


www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -374,6 +381,12 @@
                         </li>
                       <?php } ?>
                     </ul>
+
+                    <form action="<?= $actionurl ?>" method="POST" class="pull-right">
+                        <input type="hidden" name="t" value="<?= _htmlspecialchars($alert['token']) ?>">
+                        <input type="submit" class="button button--negative small" name="action" value="Delete All">
+                    </form>
```
struan: This doesn't interact well with the current mp section below:

screenshot 2018-12-07 at 16 24 31

dracos:
"I do wonder if we should remove the alerts stuff from /user and just put in a link to /alert." - forgot about that! They give different results too, on /user I get "spoken by Diane Abbott OR spoken by Gerry Adams" on /alert I get "When Diane Abbott or Gerry Adams mentions [OR]."...

www/includes/easyparliament/templates/html/alert/index.php
dracos: Whoops, should be okay now. Also is now identical on /user/ page.

PR #1408, opened 2018-12-14
Improve database querying code

dracos:
This makes the query object iterable, so that you can start to replace all the needless for loops with foreach loops.


www/includes/easyparliament/searchlog.php
```diff
@@ -166,7 +166,7 @@ public function admin_failed_searches() {
                 ORDER BY count_ips DESC, max_time DESC");
         $searches_array = array();
         for ($row=0; $row<$q->rows(); $row++) {
```
struan: Need to update this loop.

PR #1410, opened 2018-12-14
[WIP] Display recent votes on homepage

### LINKED ISSUE #1101: Highlight recent votes on the homepage
A regular query from users is where they can find a recent vote. Generally they've read about it in the press, possibly with different wording than is used in the debate title (or they haven't thought/been able to search).

We could pull out recent votes and display them on the homepage. This would help the people above and also draw casual browsers' attention to what has actually been decided upon in Parliament recently. Less significantly, it would also save time for me, since I generally search on whatever small information they have given me - a date or subject matter - and, unless there's a trick I don't know, it's by no means easy to locate the debates where votes happened.

zarino:
Fixes #1101.

Used to look like: 

image

### DIFF
--- a/www/docs/style/sass/pages/_home.scss
+++ b/www/docs/style/sass/pages/_home.scss
@@ -283,17 +283,17 @@ input.homepage-search__button {
         font-size: 1em;
         color: $colour_dark_grey;
         font-weight: normal;
+    }
 
-        // Extra space above when not first child
-        * + & {
-            margin-top: 2em;
-        }
+    // Extra space above when not first child
+    * + h2 {
+        margin-top: 2em;
     }
 }
 
 .homepage-recently {
     @include grid-column(12);
-    padding-top: 3em;
+    padding-top: 2em;
     padding-bottom: 2em;
     @media (min-width: $large-screen) {
         @include grid-column(8);
@@ -306,7 +306,7 @@ input.homepage-search__button {
     @media (min-width: $large-screen) {
         @include grid-column(4);
     }
-    padding-top: 3em;
+    padding-top: 2em;
     padding-bottom: 2em;
 }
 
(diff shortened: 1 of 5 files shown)


zarino:
@dracos I’m running out of design time on this one before Christmas, so pinging you in case you have time to finish off?

Things on my mind:

1. Should recent votes _replace_ the "Random debate" we get if no featured debate has been chosen? Or maybe we _always_ show _a few_ recent votes, but increase the number if there’s no featured debate. @MyfanwyNixon might have thoughts.
2. I’m a bit worried that the list will get clogged up with aparent "duplicate" divisions – multiple divisions from the same debate, with identical—or very similar—titles. Perhaps we group by debate (and therefore show the _debate_ title rather than the division title, and link to the _debate_ page rather than the division page), to avoid that?

MyfanwyNixon:
I don't feel that strongly, but I do quite like to random debate for its potential to pull people into content they might not otherwise have found. So I think I'd be in favour of keeping it and always having the votes separately.

www/docs/style/sass/pages/_home.scss
```diff
@@ -327,10 +327,10 @@ input.homepage-search__button {
 
 .excerpt__title {
     margin-bottom: 0.1em;
-    max-width: 21em;
     font-size: 1.125em;
     @media (min-width: $medium-screen) {
         font-size: 1em;
+        padding-right: 10em;
```
dracos: Not sure this was enough:
screen shot 2018-12-20 at 09 49 39

zarino: Ah poop, didn’t spot "Written ministerial statements"!

zarino:
@dracos I’ve pushed two commits that:

1. Move the recent votes down into the second homepage section, just before the "Recently in Parliament" heading.
2. Fix that overlapping `.excerpt__category` bug you spotted.

On point 1 – it was tempting put the list of votes _inside_ that "Recently in Parliament" list, since divisions _do_ happen in Parliament, but then you’d have an odd mix of votes and speeches under a single heading. Or you’d have a frankenheading, like the "Public bill committees" heading-thats-not-a-heading. Keeping them separate felt neater, but let me know if you disagree.

PR #1454, opened 2019-04-29
Remove "recent edits to Wikipedia from political IPs"

jacksonj04:
Removes the frontend and the cron script for the "recent edits to Wikipedia from political IPs" page.

No edits have been recorded from these IPs since 2014 at the latest, and for some of them in over a decade. Google also doesn't index the page, which suggests nobody knows about it.

## To Deploy

* [ ] Remove crontab entry from production


dracos:
Also needs to remove the public /wikipedia/ page then :)

jacksonj04:
Stray frontend page also now removed.

PR #1466, opened 2019-05-31
Link directly to screenshot of votes

stuartlangridge:
It is nice that easily-shareable screenshots of an MP's voting record are available, but at the moment they're hard to get at; they're set up to be automatically used when shared on social media (through the `og:image` meta element) but it may be useful to allow people to get at those images directly so they can copy them to share in other places (for example, pasting them into chat apps, or sharing on social media without having to use the provided "share" buttons). So, make the word "screenshot" a link to that screenshot explicitly so it can be got at.


dracos:
Thanks, should now be live!

PR #1471, opened 2019-06-20
Plain-English versions of left_reason don't include entry for new value "recall_petition"

handelaar:
Adding missing string for new ENUM value of "recall_petition"


handelaar:
(Rationale: https://www.theyworkforyou.com/mp/25638/fiona_onasanya/peterborough#profile )

dracos:
Ah, whoops, thank you!

handelaar:
(Found it by accident while idly considering a canonical list of all reasons politicians might leave office, at midnight. This was more polite than filing an issue ;) )

PR #1486, opened 2019-09-09
Added large image sizes from parliament API for profile page

ajparsons:
Adds double size images to take advantage of new profile sizes allowed by https://github.com/mysociety/theyworkforyou/pull/1373


ajparsons:
@zarino: looking at the failures for this, it looks I should actually adjust some of the tests for large images to take out the existing work-around. Does that seem right?

https://github.com/mysociety/theyworkforyou/blob/master/tests/MemberTest.php#L250

zarino:
@ajparsons Yes, it looks like that test will need to change.

Everything else _should_ just work, but best to test it out locally, to check the large images actually come though.

ajparsons:
Looks good locally: 
image

Others correctly default to smaller image

PR #1489, opened 2019-10-31
[WIP] display show votes even with no vote

### LINKED ISSUE #1480: Display "show votes" even if someone has never voted on a topic
The "show votes" button was removed in cases where someone has never voted on a topic in #934 on the grounds its presence could be confusing.  

While the button could be confusing in such circumstances not having the button means that users who want to drill down to discover which votes were missed, or see if "not voting" was due to recorded active abstentions (voting both) or absences isn't easily able to do so. 

Perhaps "details" could be used rather than "show votes" in cases where there are no votes. 

I think "show votes" would be fine too, as there still are votes to show, just not votes where the person in question voted one way or the other. 

This ticket has been created following correspondence from a user.

rbibby:
Closes #1480

`_vote_description.php` seems to be used in quite a few places so I've changed the link to be always shown and then use the previous condition to change the link text rather than whether the link is shown

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/_vote_description.php
+++ b/www/includes/easyparliament/templates/html/mp/_vote_description.php
@@ -1,7 +1,7 @@
 <li>
     <?= $description ?>
     <?php if ( $show_link ) { ?>
-        <a class="vote-description__source" href="<?= $link ?>"><?= $link_text ?></a>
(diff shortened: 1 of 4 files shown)


rbibby:
I've not managed to test this as much as I would want to yet so marked as WIP. I'll finish testing and unmark as WIP when I get chance

PR #1490, opened 2019-11-14
Vagrant adjustments

sagepe:
This updates the php7-xapian build script and uses a Vagrant box with Guest Additions installed rather than NFS.

Tested on a fresh clone (with submodules) - the box built successfully.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,11 +8,11 @@ Vagrant.configure(2) do |config|
 
   # Every Vagrant development environment requires a box. You can search for
   # boxes at https://atlas.hashicorp.com/search.
-  config.vm.box = "debian/stretch64"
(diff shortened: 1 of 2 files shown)


dracos:
Does this still need including?

sagepe:
> Does this still need including?

Seeing as we might be about to upgrade Xapian, probably worth waiting until that's done, they I'll review and either close or update and nudge you for a final review.

ajparsons:
Given we don't use vagrant anymore and we haven't merged this so far - could we close or do we need to merge at some point?

sagepe:
I'd say if we're not supporting Vagrant any more it can be closed, but I'd also suggest we remove the rest of the Vagrant stuff for cleanliness as well.

ajparsons:
Closing as Vagrant support removed

PR #1496, opened 2020-01-01
Correct broken policy link

### LINKED ISSUE #1488: One of the links to PublicWhip broken on pages on an MP's stance on a particular policy
For example at:

https://www.theyworkforyou.com/mp/10555/nicholas_soames/mid_sussex/divisions?policy=6693

In: "Vote information from [PublicWhip](http://www.publicwhip.org.uk/mp.php?mpid=42059&amp=6693)."

the link points to 
`http://www.publicwhip.org.uk/mp.php?mpid=42059&amp=6693`

The correct link, as used further up the page, contains `dmp` not `amp`, it should be:

`http://www.publicwhip.org.uk/mp.php?mpid=42059&dmp=6693`

chloecollins:
Resolves #1488 

Updates URL parameter from amp (which is not currently a valid parameter and so is disregarded) to dmp (which restricts results to a specific policy).

### DIFF
--- a/www/includes/easyparliament/templates/html/mp/divisions.php
+++ b/www/includes/easyparliament/templates/html/mp/divisions.php
@@ -102,7 +102,7 @@ include_once INCLUDESPATH . "easyparliament/templates/html/mp/header.php";
                                         <p><button class="button secondary-button small js-show-all-votes">Show all votes, including <?= $policy['weak_count'] ?> less important <?= $policy['weak_count'] == 1 ? 'vote' : 'votes' ?></button></p>
                                       <?php } ?>
                                         <p class="voting-information-provenance">
-                                            Vote information from <a href="http://www.publicwhip.org.uk/mp.php?mpid=<?= $member_id ?>&amp=<?= $policy['policy_id'] ?>">PublicWhip</a>.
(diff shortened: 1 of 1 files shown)


dracos:
Thank you! Interestingly (well, sort of), it should actually be `&amp;dmp=` as the ampersand needs escaping in HTML - which is presumably what led to the confusion in the previous commit :) If you're able to fixup that, I'll get it merged in. Thanks again for your contribution :)

chloecollins:
Thank you, that is interesting - I had noticed an &amp usage in line 69 but mistakenly disregarded it, thinking it was only needed in PHP interpolation. That's good to know; it's due to URL encoding instead :) Thanks for the correction - I'll update that!

dracos:
"That's good to know; it's due to URL encoding instead" - to be pedantic, due to HTML encoding (e.g. if the variable had been called `copy`, not `dmp` then `&copy=` in the URL could lead to the copyright symbol appearing in the link instead). URL encoding would be if you needed to pass the ampersand as data rather than a variable separator, in which case it would be `%26`. Hope that makes sense!

I've squashed your two commits together and merged them, that should now be live, thanks for your contribution :)

PR #1517, opened 2020-06-22
Covid19 voting banner

zarino:
New banner with Covid-19 wording from @MyfanwyNixon and @ajparsons.

I’m thinking we’ll squash the final three commits (the ones about the panel) into one before merge.

## Desktop, before and after opening the details

Screenshot_2020-07-06 Nigel Huddleston MP, Mid Worcestershire - TheyWorkForYou(3)

Screenshot_2020-07-06 Nigel Huddleston MP, Mid Worcestershire - TheyWorkForYou(4)

## Mobile, before and after

Since the `<details>` element isn’t supported in any version of IE, we include a quick polyfill for it at the end of `main.js`.

Also, while I was there, I committed some README changes I’ve had hanging around for ages, removed some old unused code, and fixed some incorrect wording in the alert emails.


MyfanwyNixon:
For wording, how about something like: 

> The COVID-19 pandemic has affected which MPs are able to vote, and how some votes are conducted. Please take this into consideration when assessing voting records, and especially where an MP was absent. [See more detail on votes during the COVID-19 period here](https://blogs.mysociety.org/mysociety/?p=47332&preview=true).

zarino:
Thanks @MyfanwyNixon. I’ve put that wording into place. All we’re waiting for now is the URL of the live blog post, and we can get this live.

www/docs/js/main.js
```diff
@@ -349,4 +349,23 @@ $(function(){
     }
   });
 
+  var supportsDetailsElement = function(){
+    var el = document.createElement('details');
+    if ( ! ('open' in el) ) {
+      return false;
+    }
+    $('body')[0].appendChild(el);
```
dracos, line 357: I mean, fine, but I think this was to work around a bug in Chrome 10? ;)

zarino: Hah, you’re right. Just the `'open' in el` check is enough for IE9–11, which is what we care about here. :+1:

zarino:
@ajparsons @MyfanwyNixon I’ve merged this change into `master`, but I’ve not deployed it yet, as the blog post we link to (https://www.mysociety.org/2020/07/06/parliamentary-votes-during-covid-19/) doesn’t exist yet. Once that’s published, let us know, and we’ll deploy.

PR #1520, opened 2020-07-08
Add page showing invoices for download.

dracos:


www/docs/api/invoices.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+ini_set('error_reporting', E_ALL);
+include_once '../../includes/easyparliament/init.php';
+include_once './api_functions.php';
+
+MySociety\TheyWorkForYou\Utility\Session::start();
+
+if (1 || $THEUSER->loggedin()) {
```
struan, line 9: I think you may have left in some test code :)

classes/Subscription.php
```diff
@@ -176,6 +176,15 @@ private function add_subscription($form_data) {
         ]);
     }
 
+    public function invoices() {
+        $invoices = $this->api->getInvoices([
+            'subscription' => $this->stripe->id,
+            'limit' => 24
```
struan, line 182: Given this do we want some text somewhere to tell people we only display the last 24 invoices?

PR #1531, opened 2021-01-05
Various Xapian indexer fixes

### LINKED ISSUE #1530: Indexing errors when a person has a redirect
Running the Xapian indexer produces an error when there's been a redirect added for a duplicate`person_id`. This is a problem when trying to rebuild the index from scratch.

For example, James Callaghan has two `person_id` values, `16877` and `13677`, with the former redirecting to the latter. When trying to reindex debate 1945-09-20, we see the error `No options for person id 16877 on 19450820, major 1 at ./index.pl line 372.`

We'll need to ensure that the redirects are mapped correctly in `index.pl`.

sagepe:
This includes commits to fix the following issues with the Xapian indexer:

* Ensure person redirects are honoured (fixes #1530)
* Update the script to index everything to include all dates from 1919 to present
* Ensure the date format for the NI Speaker exceptions matches

### DIFF
--- a/search/index.pl
+++ b/search/index.pl
@@ -15,6 +15,8 @@ use lib "$FindBin::Bin/../commonlib/perllib";
 use mySociety::Config;
 mySociety::Config::set_file('../conf/general');
 use DBI;
+use JSON;
+use File::Slurp;
 #use Data::Dumper;
 #DBI->trace(2);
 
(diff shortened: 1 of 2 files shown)


dracos (approved):
Looks good, thanks!

Issue #1421, opened 2019-02-06
Improve page speed


zarino:
A few ideas:

* Javascript is blocking – move it to the footer, or make it async?
   * If we move it, we’ll need to update/remove the inline `<script>` tags that expect `$` to be defined.
* Wrap `respond.js` in an IE6–8 conditional comment – or remove it entirely. (IE 6–8 accounts for less than 0.1% of visits.)
* Use the more modern `document.body.className…` JS snippet at the top of the body to add a JS class for CSS to use.
* We almost certainly don’t need `foundation.reveal.js` in the footer.
* If we stop hiding the vote summary "Details" button until hover (281fb424057adfcec83f1e33928fab6679f7147c) – a good idea anyway – then we can remove `custom.modernizr.js` entirely.
* Replace `foundation.magellan.js` (and `foundation.js` itself!) with our own sticky sidebar solution – using CSS `position: sticky` and `intersectionObserver`s for highlighting the current page section.

zarino:
I’ve started work on this in the [`2021-performance-improvements`](https://github.com/mysociety/theyworkforyou/compare/2021-performance-improvements) branch.

dracos:
All deployed. webpagetest shows gentle improvement:
before: https://webpagetest.org/video/compare.php?tests=210205_DiQV_b5693a68c81de06206e68d6434018ba5-r:2-c:0
after: https://webpagetest.org/video/compare.php?tests=210205_DiTJ_3d75bdfe4401ef922c605f0a12e8dd13-r:2-c:0
It looks like all of foundation-icons is included only for social icons on MP profile page, feels like we could manually include just those? I have done that and turned on display=swap for the Google font, which leads to:
https://webpagetest.org/video/compare.php?tests=210205_DiFQ_123c48ff0558e32a3445a07cae70bc43-r:3-c:0
Will that do?

PR #1540, opened 2021-01-29
2021 performance improvements

zarino:
Performance improvements as part of https://github.com/mysociety/theyworkforyou/issues/1421.

Most of this will probably be imperceptible, but it’s nice to tidy stuff up.

Ripped out a bunch of JavaScript plugins, moved all of the remaining JavaScript to the end of the body for a slight performance boost, added cachebusting query params where they were missing, and removed some IE8 workarounds.

I’d hoped we could drop either the 600 or 700 weight of Source Sans Pro, but they’re both fairly important for different things :-/

The only user-facing change is that scrolling down the MP pages will no longer highlight the "current" link in the sidebar. We could write something to acheive this with IntersectionObserver, but it feels like a lot of faff, and I’m not sure it’s worth it. Any thoughts @dracos ?


www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -66,8 +65,8 @@
                 <?php if ( !$current_member[HOUSE_TYPE_COMMONS] ) { ?>
                     <?php if (count($policyPositions->positions) > 0) { ?>
                     <div class="panel">
-                        <a name="votes"></a>
-                        <h2 data-magellan-destination="votes">A selection of <?= $full_name ?>'s votes</h2>
+                        <a name="votes" id="votes"></a>
```
dracos, line 68: Why have both name and ID on these?

www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -66,8 +65,8 @@
...
+                        <h2>A selection of <?= $full_name ?>'s votes</h2>
```
dracos, line 69: curly quote to match others?

dracos (changes requested):
Looks good! A few small changes is all.

bin/make-css
```diff
@@ -0,0 +1,16 @@
...
+
+for argument in "$@"
+do
+    [[ "$argument" == "--watch" ]] && MODE="watch"
+done
+
+export PATH="$DIR/vendor/bundle-bin:$PATH"
+
+cd "$DIR/www/docs/style"
+compass "$MODE"
```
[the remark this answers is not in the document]
zarino: Ah, I was just copying what `deploy.bash` did. Maybe they should both use `bundle exec compass`?

www/includes/easyparliament/templates/html/mp/profile.php
zarino: Oops, yes, not sure why I did this. Jumping to elements by `name` attribute works fine, without an `id`. Force of habit.

bin/make-css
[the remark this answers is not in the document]
zarino: I’ve gone ahead and added `bundle exec` to both anyway.

dracos (approved):
Can we get a webpagetest or two before to compare to after? :)

PR #1543, opened 2021-02-09
Add party voting line to MP votes page.

### LINKED ISSUE #1512: Display party comparison information
Discussed and in principle approved at February 2020 meeting. 

Problem: Complaints that screenshots of TheyWorkForYou on social media do not include the relationship with the party. TheyWorkForYou does calculate this, but only presents it on the MP profile page, while screenshots tend to be  taken of topic areas from the full voting record page (images below). Screenshots do not contain hyperlinks and so our links to clarifying FAQs can’t address the issue. 

image

image

Solution: Similar to how the Guardian [put the year of publication](https://www.theguardian.com/help/insideguardian/2019/apr/02/why-were-making-the-age-of-our-journalism-clearer) in the social image preview, we could put the party stance calculations against every item on the voting record so that this information is captured within the screenshot. This involves no new information, but presents our current calculations in a new place so it is clearer we have done the comparison work. 

This was considered the more difficult change discussed and need an estimate of how long it would take.

dracos:
Fixes #1512.
image
Includes the same information as is shown on the main MP page (where one score <0.4 and other is >0.6), but also includes differences where one score is between 0.4 and 0.6 and other score is outside that range (so one is 'mixed' and other is not mixed). Thoughts welcome.
Tried to tidy up a bit too, having the diffs include the information they need rather than setting the policies/party_positions globals.

### DIFF
--- a/classes/Member.php
+++ b/classes/Member.php
@@ -388,11 +388,18 @@ class Member extends \MEMBER {
                 if ( $only_diffs && $score_diff < 2 ) {
                     continue;
                 }
-                $policy_diffs[$policy_id] = $score_diff;
+                $policy_diffs[$policy_id] = [
+                    'policy_text' => $details['policy'],
+                    'score_difference' => $score_diff,
+                    'person_position' => $details['position'],
+                    'party_position' => $party_positions[$policy_id]['position'],
+                ];
             }
         }
 
-        arsort($policy_diffs);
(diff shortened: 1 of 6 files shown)


struan (approved):
👍

ajparsons (approved):
I see it in the code, but the screenshot at the top is slightly out of date (doesn't have the 'current' bit)

PR #1544, opened 2021-02-09
Include party vote date ranges in voting summary lines

### LINKED ISSUE #1311: Clarify statements comparing MPs' stances to those of their parties
On MP's pages TheyWorkForYou carries statements like:

>[MP NAME] consistently voted against [SUBJECT], while most [PARTY} MPs generally voted for. 

sometimes the comparison being made is between an MP who has only voted 1-2 times recently and a party's voting record over 15+ years; this can be misleading.  

Already under the statements on the voting record pages, there is something like, for example:

>"2 votes for, 2 votes against, 3 absences, between 2013–2014"

that approach could be used on the statements comparing MP's stances to their parties on MP pages with something like:

>[MP name], one vote in 2016; Conservative Party, 20 votes between 2003-2016

placed below the statements. 

The increase in clarity for those looking for the detail is balanced by an increase in clutter. 

This ticket has been prompted by comments from a couple of MPs.

dracos:
Adds the voting summary lines to the diff votes thing on main MP profile page (it seemed like this was a good thing to do, though doesn't have to, I guess? Does make it bigger), and then adds party date range to the voting summary lines, both there and on voting record page. Thoughts welcome. Fixes #1311
MP voting page example:
image
MP profile example:
image


struan (approved):
👍

ajparsons (approved):

Issue #1509, opened 2020-05-15
Future calendar no longer working


dracos:
Parliament have stopped updating https://services.parliament.uk/calendar/ even though they still link to its RSS feed on https://www.parliament.uk/site-information/rss-feeds/ - and that is what we were using to update our calendar. They now have https://calendar.parliament.uk/ which looks data-backed but doesn't appear to have any programmatic data access I can easily see. It makes POST requests for daily/weekly per-house data. I doubt we have time to work out how to fetch the equivalent data, we should probably for now just turn off the calendar, perhaps ask Parliament if the data is available somewhere (data.parliament.uk hasn't been updated in a long time), otherwise scraping future weeks per house, sigh.

MyfanwyNixon:
Just to note that we're doing work on an AQA module, teaching young people how they can engage with politics and parliament, where this would be useful! In the past we've also told people that subscribing to future business alerts is great especially for campaigners, as per this [blog post](https://www.mysociety.org/2015/10/12/how-to-contact-your-mp-at-just-the-right-time/). 

I understand the issues above, but am just updating to add whatever small piece of impetus to this ticket I can.

dracos:
Perhaps look at https://github.com/mysociety/theyworkforyou/issues/572

PR #1547, opened 2021-02-12
Fix future calendar

dracos:
Fixes https://github.com/mysociety/theyworkforyou/issues/1509

Screenshot_2021-02-15 Wednesday, 24 February 2021 Upcoming business


struan (approved):
📆

dracos:
@zarino Have pushed two commits that move stuff out of the template and then do that more and refactor the two calendar display pages together (and removes a third that wasn't used by anything!), take a look and let me know what you think, then this is done I think apart from if you wanted to "Now I just need to add a .business-section__secondary__item class to
the calendars and minisurveys in section/business page sidebars"

zarino:
@dracos I‘ve added a commit that makes that `.business-section__secondary__item` change I was talking about.

Issue #1507, opened 2020-04-08
Spacing issues in policy stance image generator


DanAtkinson:
Using my local MP David Davis' social issues stance as an example below, there appears to be a spacing issue between the emboldened policy stance ([Link](https://www.theyworkforyou.com/mp/10162/david_davis/haltemprice_and_howden/policy_set_png?policy_set=social)).

![https://www.theyworkforyou.com/mp/10162/david_davis/haltemprice_and_howden/policy_set_png?policy_set=social](https://user-images.githubusercontent.com/23440/78808850-8d024880-79b5-11ea-887f-700ddde7aa39.png)

**What I'm expecting to see:**
> Consistently voted against allowing **marriage** between two people of same sex 

**What I actually see:**
> Consistently voted against allowing **marriage**between two people of same sex

dracos:
The code for this output works by generating a SVG, then converting to a PNG. Weirdly, the SVG does not have the spacing issue:
![image](https://user-images.githubusercontent.com/154364/78811228-6cd88680-79c1-11ea-925c-0657d82f3878.png)
(and also displays more policies, I'm guessing transparency is the difference there). So it is the SVG to PNG conversion that somehow loses the spacing, I'm guessing some issue with the font.

PR #1548, opened 2021-02-16
Remove bold body text from PNG output, and move heading to top

dracos:
Fixes https://github.com/mysociety/theyworkforyou/issues/1507

image


zarino (approved):
👨‍🍳 💋 👌

PR #1550, opened 2021-02-16
Improve large image sources

ajparsons:
This PR updates the `download_parliament_portraits.py` to do the following:

* Pull down all official where `people.json` has the parliament ID in a slightly larger size (the default for the ThreeFour size 260 x 346) and put these in an `mpsOffical` folder.
* Using name information and wikidata, bridge between twfy_id and wikipedia to pull down images used for MPs on Wikipedia and put these in an `mpsWikidata` folder. 
* Using name information and wikidata, bridge between twfy_id and wikipedia to pull down images used for MPs on wikidata and put these in an `mpsWikidata` folder [where not already sourced from wikipedia]. 
* Resize these images to the `mps` and `mpsL` folder, passing with white where wikipedia images are off format. 

I additionally did a pass to remove wikipedia photos that were not of people (graves, buildings, caricatures), although these will mostly apply to historical people before we actually start transcripts. 

This should improve the size and quality of images for current and recent MPs (for instance, gives good images for a number of former prime ministers).

I think this is repeating an older exercise that sourced a number of the small images in the first place - so I'm storing both the script and the slightly larger version of the file in case we want to adjust our top size in future. 

Currently:

```
There are big images for 3503
There are small images for 4280
Small but not big: 777
```


dracos:
This looks good, though I am a bit wary of putting so much in the repo (but where else to put it). Could we perhaps have just the ones we need in this repo and put the larger ones in a different archive repo so as not to clutter this one up, maybe?
Or saying that, should we put all the images in a different repo, deploy that as people-images.theyworkforyou.com or whatever and then just refer to them there from here? Would that be a good idea?

ajparsons:
I'd agree it's not ideal the images are all in here, especially if we're storing larger and larger sizes. Splitting all the images off seems the correct solution, but I guess it depends how much work?

The lookup function for different sizes currently relies on being able to look at the files in question to see what is and isn't there - so that might need a new approach if storing elsewhere. 

Could simplify the problem on that end by making sure there a 'large' sized image available for every MP just by adding a frame to MPs there are only small images for. 

The advantages of that would be being able to group that repo with the update script and have it be a bit clearer how to update and maintain in the long run.

ajparsons:
Closing as resolving the problem in a different way.

PR #1556, opened 2021-03-10
Catch errors at Stripe customer creation.

dracos:
Same change as for https://github.com/mysociety/mapit.mysociety.org/pull/120


struan (approved):
👍

PR #1565, opened 2021-06-20
Docker

atiro:
I was interested in setting the application up, but don't use Vagrant, so ported the setup to docker. Had to alter a couple of scripts (not totally sure why, different version of perl and php?) and some paths, so not a PR to be merged in haste!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+FROM alpine:3.14
+
+WORKDIR /twfy
+
+RUN apk add wget php7-dev php7-cli php7-common memcached
+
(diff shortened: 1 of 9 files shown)


sagepe:
Hi! Thanks for this @atiro. We're currently looking at a Docker-based development environment, so seeing how you have approached the problem is certainly interesting.

I think we'd take a slightly different approach to maintain compatibility (as you imply, I suspect some of the changes you needed to make were due to varying language versions and platform-specific implementations). We'll certainly include the ability to configure Memcache to use an external service.

So we may incorporate some of this PR - I'll respond in a bit more detail in due course.

atiro:
Cheers @sagepe , no problem if you want to close, I thought I might as well send the PR as I had done it, but I was definitely hacking away at some points!

sagepe:
Thanks @atiro - I'm going to close this in favour of #1569 (I've marked you as a co-author on one of the commits in that PR), but thanks again for taking the time to send this over - we're always glad to receive contributions.

PR #1569, opened 2021-07-01
Add a Docker-based development environment

sagepe:
This adds a new development environment based on Docker and Docker Compose and marks the old Vagrant environment as deprecated.


gbp (commented):
FWIW I was able to get a working (as far as I can tell) docker install on my M1 mac using these changes

mySocietyClive (approved):
Worked for me also.

PR #1571, opened 2021-07-06
User sign up / sign in UX improvements

### LINKED ISSUE #1559: User sign up / sign in UX improvements
Broken out from #1532.

Improvements to the signing in / joining pages:

### https://www.theyworkforyou.com/user/?pg=join

* Remove website field
* Check whether we still need name fields – if not, remove
* Email field should be `type="email"`
* Position field labels above fields, rather than beside them
* Style up the form elements to match [Sign in page](https://www.theyworkforyou.com/user/login/) – looks like a `.form-control` class would do it?

### https://www.theyworkforyou.com/user/password
* Style up the form elements to match [Sign in page](https://www.theyworkforyou.com/user/login/)

### https://www.theyworkforyou.com/alert
* Could we combine the account creation logic from `/user/?pg=join` into the alert subscription form on `/alert`? I can imagine a flow for unauthenticated users something more like – 1) enter search term, 2) pick my chosen alert at the confirmation/disambiguation step, 3) choose to set up one-off alert by email, or create an account to manage this and future alerts more easily. Sort of like the "check out as guest" vs "create an account" flow on a shopping site.
   * If we can do this, perhaps "Join" in the nav bar would become "Manage alerts"?
   * If we can’t, can we at least make it so that the "join" link in the sidebar return you back to `/alert`, like the "sign in" link does?

mySocietyClive:
Fixes #1559

Cut from @sagepe issues/1568-update-development-environment branch which is not merged into master yet.

Tidy up form fields styles.


mySocietyClive:
@wrightmartin I've added screens hots now.

Cheers

wrightmartin:
So much better! 🚀

www/docs/style/sass/pages/_legacy.scss
```diff
@@ -32,33 +38,56 @@
...
+                    margin-bottom: .5em;
                 }
             }
         }
     }
+
+    .password-form {
+        label[for="em"] {
+            font-size: inherit;
+            margin-bottom: .5em;
```
wrightmartin, line 88: Feeling like we should keep the `0.` part of these just to keep to the house style

www/includes/easyparliament/templates/html/user/join.php
```diff
@@ -8,10 +8,10 @@
             <div class="main">
 				<h1>Join TheyWorkForYou</h1>
 
-        <p>Joining TheyWorkForYou makes it easier to manage your email alerts.</p>
+        <p>Joining TheyWorkForYou makes it easier to manage your email alerts...</p>
```
wrightmartin, line 11: Should this be an ellipsis? `&hellip;`

mySocietyClive:
Amended those bits.

wrightmartin (approved):

dracos:
This has already been merged

PR #1572, opened 2021-07-09
Issues/1558 promo banner

mySocietyClive:
Added banner as an include.


zarino (changes requested):
A few comments inline.

zarino (approved):
YASSS. Everything lines up. Very pleasing. 👍 

image

Issue #1567, opened 2021-06-30
Move CI to GitHub Actions


sagepe:
This project was still using travis.org until builds stopped there on 15th June. We should port over the tests to GitHub Actions.

PR #1586, opened 2021-12-13
Switch to GitHub Actions.

dracos:
Fixes https://github.com/mysociety/theyworkforyou/issues/1567


sagepe (approved):
Looks good, thanks. Might be worth mentioning in the commit message that this fixes #1567.

Once we've completed the upgrade we can remove the tests against the older versions and start testing against 8.x so we get ahead of the game for next time - I'll add a ticket for this so it doesn't get forgotten.

PR #1590, opened 2022-01-21
Adapt docker for codespaces

ajparsons:
This pull request adapts the existing docker settings to work on codespaces (but should work as expected locally as well). 

There are two approaches this is taking modifying the docker I'd like to check (but comments on anything else good as well):

## Changing working directory

Codespaces doesn't respect the `workspaceFolder` setting in `devcontainer.json`, so I can't direct it to the `\twfy` directory things were being set up in before. 

I've adjusted the references in the dockerfile and docker-compose to `\workspaces\theyworkforyou`, where codespaces expects it. Rather than adjust the entrypoint and other script, I added a symbolic link to `\twfy` in the docker so those work. 

Case for: don't like pushing codespaces specific config too far into the scripts (especially if they might support the setting at some point).
Case against : this is all docker-specific config anyway, might as well be consistent and use the same folder for everything. 

## Adding extra database

To add the test database for the phpunit tests, I just added a new mariadb to docker-compose. I think this is the more docker way of handling it, rather than adding a script to add another database to the first mariadb, but wanted to check the logic.


Dockerfile
```diff
@@ -34,8 +36,12 @@ RUN /usr/sbin/a2enmod expires rewrite && \
       ln -sfT /proc/self/fd/1 /var/log/apache2/access.log && \
       ln -sfT /proc/self/fd/1 /var/log/apache2/other_vhosts_access.log
 
-# Bind mount your working copy here
-WORKDIR /twfy
+# Codespaces needs the working directory to be /workspaces/theyworkforyou
+# Some scripts want it in twfy, minimise changes needed for codespaces
+WORKDIR /workspaces/theyworkforyou
+RUN ln -s /workspaces/theyworkforyou /twfy
```
sagepe, line 42: There's a part of me that doesn't like having environment-specific things in here, but as long as we want to support choice in development environments it's probably going to be inevitable, or lead to lots of convoluted workarounds.

I'd agree that this is better here than in the scripts, etc, themselves where it's more hidden - as you say, they might support the setting at some point, so just setting that here in a backwards-compatible way seems like a reasonable compromise.

docker-compose.yml
```diff
@@ -29,11 +39,24 @@ services:
...
+    command: '--sql_mode="ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION"'
+    environment:
+      MARIADB_RANDOM_ROOT_PASSWORD: '1'
+      MARIADB_USER: twfytest
+      MARIADB_PASSWORD: password
+      MARIADB_DATABASE: twfytest
+    volumes:
+      - ./db/schema.sql:/docker-entrypoint-initdb.d/schema.sql
+      - testdb:/var/lib/mysql
+
```
sagepe, line 53: I agree that a container-per-database is fine if that's what makes sense for the tests.

docker-compose.yml
```diff
@@ -10,12 +10,22 @@ services:
     depends_on:
       - mariadb
       - memcache
```
sagepe, line 12: I wondered whether we should add mariatestdb here, but I guess it's not strictly needed immediately, just when you want to run the unit tests, so it's probablt appropriate not to.

sagepe (approved):
This looks fine to me, and works locally here on my M1 Mac - as long as I add `platform: linux/amd64` to the `twfy` service - this is needed because at this time we're not supplying arm packages in our repo. I think it might be worth adding this as a separate commit, so I'll push a commit to this branch with that as well.

Also noting that this branch is just on GitHub, not on git.mysociety.org.

ajparsons:
FYI have fixed "this branch is just on GitHub, not on git.mysociety.org."

ajparsons:
@sagepe : I've readapted this after the big change. Actually needs slightly fewer codespaces specific modifications now. 

Changes I've now made to the docker-compose:

- mariatestdb depends on mariadb being set up first. (Something to do with them sharing a schema file?)
- I had to remove the version number to get it to work on codespaces. It looks like this doing this is actually [the most up to date version of the compose spec](https://github.com/compose-spec/compose-spec/blob/master/spec.md) confusingly.

docker-compose.yml
```diff
@@ -1,15 +1,16 @@
 ---
-version: '3.9'
```
sagepe, line 2: Yes I see what you mean, the spec lists `version` as deprecated, and it seems it's no longer needed with recent versions. Might Do we just want to take this out completely?

sagepe (approved):
Works for me locally - at least I get a basic environment up and running when I run `docker compose up`. One comment on the compose file `version` thing but not a blocked for approval I think.

PR #1591, opened 2022-01-21
Speeding up populate scripts

ajparsons:
Building on https://github.com/mysociety/theyworkforyou/pull/1590 (which you might also have opinions on), because codespaces starts from scratch (and was getting an idle timeout on a script that didn't report back for a while), I wanted to improve the speed of the initial import of data.

I've done this by wrapping some of the INSERT statements in various scripts looks in START TRANSACTION and COMMIT. This seems to work ok for my testing, but might have other implications. 

There is also a `scripts/quick-populate` that handles pulling in a small set of data and running the correct scripts to process it. I think all the steps currently there are required, but anything extra or missing?

### DIFF
--- a/classes/PartyCohort.php
+++ b/classes/PartyCohort.php
@@ -435,6 +435,7 @@ class PartyCohort
     public static function calculatePositions($quiet=true){
         // get current hashes available
         $cohorts = PartyCohort::getCohorts();
+        $db = new \ParlDB;
         $policies = new Policies;
         $n_cohorts = count($cohorts);
 
@@ -448,9 +449,12 @@ class PartyCohort
 
             $cohort_count++;
 
+            $db->query("START TRANSACTION");
             foreach ( $positions as $position ) {
                 $cohort->cache_position( $position );
             }
+            $db->query("COMMIT");
+
             if (!$quiet) {
                 print("$cohort_count/$n_cohorts\n");
             }
--- a/scripts/json2db.pl
+++ b/scripts/json2db.pl
@@ -13,6 +13,14 @@ mySociety::Config::set_file("$FindBin::Bin/../conf/general");
 
 my $parldata = mySociety::Config::get('RAWDATA');
 
+my $verbose = 0;
+for( @ARGV ){
+  if( $_ eq "--verbose" ){
+    $verbose = 1;
+    last;
+  }
+}
+
 use DBI;
 use File::Slurp::Unicode;
 use JSON::XS;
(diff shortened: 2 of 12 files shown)


scripts/mpinfoin.pl
```diff
@@ -158,7 +158,11 @@
 my $personinfoupdate = $dbh->prepare('update personinfo set data_value=? where person_id=? and data_key=?');
 my $consinfoadd      = $dbh->prepare("insert into consinfo (constituency, data_key, data_value) values (?, ?, ?) on duplicate key update data_value=?");
 
+my $start_transaction = $dbh->prepare("START TRANSACTION");
```
dracos, line 161: This presumably works, but an easier way would be to call `$dbh->{AutoCommit} = 0;`, then no need to start any transactions, and can call `$dbh->commit` to commit, then set it back to 1 afterwards.
Does putting these in a transaction speed them up? Bit surprising to me.

scripts/load-people
```diff
@@ -146,20 +150,25 @@ sub load_people {
...
     }
+    $start_transaction->execute();
     foreach (@{$j->{memberships}}) {
         if ($_->{redirect}) {
             $gradd->execute($_->{id}, $_->{redirect});
         } else {
             load_member($_);
         }
     }
+    $query_commit->execute();
```
dracos, line 161: Not sure I follow this logic - if we're putting this in a transaction, don't we want all of memberships and people to be committed at once? Otherwise you could have memberships in the DB that talk about a person that's not yet been inserted.

scripts/load-people
```diff
@@ -78,6 +78,10 @@ sub db_connect {
 
     $gradd = $dbh->prepare("replace into gidredirect (gid_from, gid_to) values (?,?)");
     $grdelete = $dbh->prepare("delete from gidredirect where gid_from = ?");
+
+    $start_transaction = $dbh->prepare("START TRANSACTION");
```
dracos, line 82: Same as other file - looks like you only want to turn off AutoCommit for the load_people function, if that's the behaviour you want.

classes/PartyCohort.php
```diff
@@ -448,9 +449,12 @@ public static function calculatePositions($quiet=true){
 
             $cohort_count++;
 
+            $db->query("START TRANSACTION");
```
dracos, line 452: And in PHP with PDO that we use, you call $db->conn->beginTransaction() and $db->conn->commit();

scripts/json2db.pl
```diff
@@ -67,12 +78,16 @@
...
+    if ($verbose){
+        print("processing motions for $dreamid\n");
+    }  
     process_motions($policy, $dreamid);
 }
 
 # And recently changed ones
 my $policy_file = $motionsdir . "recently-changed-divisions.json";
 if (-f $policy_file) {
+    print("processing recently changed divisions\n");
```
dracos, line 90: Needs a verbose?

scripts/quick-populate
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# For a fresh install get a basic set of data in
+
+echo 'Cloning parlparse'
+git clone https://github.com/mysociety/parlparse data/parlparse
```
dracos, line 5: depth 1 if you want to be quick?

dracos (changes requested):
Only looked at the TWFY code bits - a couple of suggestions for doing it more in the language keeping. Would be interested in knowing what difference the transactions make.

scripts/load-people
ajparsons: Re grouped it so it was doing both in the same commit.

ajparsons:
Closed in favour of [679](https://github.com/mysociety/theyworkforyou/pull/1679)

PR #1597, opened 2022-03-15
Add link to professional use survey to email alerts

ajparsons:
This adds a link to an alchemer survey to the text and html alerts template.

### DIFF
--- a/www/includes/easyparliament/templates/emails/alert_mailout.txt
+++ b/www/includes/easyparliament/templates/emails/alert_mailout.txt
@@ -1,4 +1,12 @@
 Subject: Your TheyWorkForYou email alert
+
+---
+Do you use TheyWorkForYou as part of your work?
+We're trying to understand more about how people use TheyWorkForYou to help make the case for funding. 
+Can you take a quick five question survey to help us?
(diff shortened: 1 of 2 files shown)


www/includes/easyparliament/templates/emails/alert_mailout.txt
```diff
@@ -1,4 +1,17 @@
...
+
+---
+
+Do you use TheyWorkForYou as part of your work?
+    
+We're trying to understand more about how people use TheyWorkForYou to help make the case for funding. 
+
+Can you take a quick five question survey to help us?
+
+https://mysociety.org/surveys/twfy-professional-use/
```
dracos, line 11: Could we have www here? Just saves a redirect.

dracos (commented):
How long is this running for? Only as we can run email alerts with a different template at the command line rather than having to edit the code itself if it were only once or twice. Otherwise, looks fine, just couple of probably pointless queries about the text part.

ajparsons:
A week or so would be fine I think, happy to defer if that means its best to do a different template or not. 

Have made those changes.

dracos (approved):

PR #1679, opened 2022-11-01
Auto populate script

ajparsons:
This PR adjusts data loading scripts to be quicker for loading initial sets of data, and creates a new `script/quick-populate` to make loading a minimal set of data consisutent.

This replaces https://github.com/mysociety/theyworkforyou/pull/1591 because that was merging into an outdated version of a branch that has now been merged into master.

I have adjusted the populate scripts based on the feedback in that PR request.

### DIFF
--- a/scripts/json2db.pl
+++ b/scripts/json2db.pl
@@ -53,9 +53,6 @@ my $strong_for_policy_check = $dbh->prepare("SELECT count(*) as strong_votes FRO
 my $strong_vote_add = $dbh->prepare("INSERT into personinfo ( data_key, data_value, person_id ) VALUES ( ?, ?, ? )");
 my $strong_vote_update = $dbh->prepare("UPDATE personinfo SET data_value = ? WHERE data_key = ? AND person_id = ?");
 
-my $start_transaction = $dbh->prepare("START TRANSACTION");
-my $query_commit = $dbh->prepare("COMMIT");
-
 my $motionsdir = $parldata . "scrapedjson/policy-motions/";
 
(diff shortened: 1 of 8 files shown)


scripts/json2db.pl
```diff
@@ -82,9 +99,13 @@
 
 sub process_motions {
     my ($policy, $dreamid) = @_;
+    $start_transaction->execute();
```
dracos, line 102: Bit confused why this file calls transaction SQL, whereas another one you've changed you've altered the AutoCommit value. Best to do both the same way, unless there's a reason?

scripts/json2db.pl
```diff
@@ -13,6 +13,14 @@
 
 my $parldata = mySociety::Config::get('RAWDATA');
 
+my $verbose = 0;
+for( @ARGV ){
+  if( $_ eq "--verbose" ){
```
dracos, line 18: Could spacing generally match the existing spacing in the same file, just for consistency :)

dracos (changes requested):
Only real query is about the two different methods of transactions, probably misunderstood something.

scripts/json2db.pl
ajparsons: I've switched to this to the AutoCommit approach used elsewhere.  Run the import and it seems fine.

scripts/json2db.pl
ajparsons: Amended this spacing.

dracos (approved):
I meant the spaces round the brackets as well, but is fine ;)

PR #1694, opened 2023-02-14
[TWFY] Added mySociety 20th anniversary badge

lucascumsille:
Added badge in header:

### Preview here:
Screenshot 2023-02-14 at 11 53 36
Screenshot 2023-02-14 at 12 00 47


zarino (approved):
Tricky one, this. Think you’ve come up with the best solution :+1:

PR #1698, opened 2023-04-04
Remove top margin on form controls

ajparsons:
There's a margin that offsets the text in text boxes, which does this to the search:

image

Removing the line fixes it:

image

Looks like it was added a few years ago to add spacing on user sign up page - but this seems a worse problem that that.

### DIFF
--- a/www/docs/style/sass/_twfy-mixins.scss
+++ b/www/docs/style/sass/_twfy-mixins.scss
@@ -307,6 +307,7 @@ input.form-control {
     font-size: inherit;
     line-height: inherit;
     padding: 0.35em 0.4em;
+    margin: 0 0 0 0;
     border: 1px solid $colour_mid_grey;
     @include border-radius(3px);
 }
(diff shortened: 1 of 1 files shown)


www/docs/style/sass/_twfy-mixins.scss
```diff
@@ -307,7 +307,6 @@ input.form-control {
     font-size: inherit;
     line-height: inherit;
     padding: 0.35em 0.4em;
-    margin: .2em 0 0 0;
```
dracos, line 310: Could we not remove it but go back to `margin: 0;` as it was before, which still fixes this but doesn't add extra margin bottom to places we might not expect (not that it really matters I suspect).

dracos (changes requested):

www/docs/style/sass/_twfy-mixins.scss
ajparsons: Can do!

dracos (approved):

PR #1702, opened 2023-04-18
More info on multi rep page

ajparsons:
In anticipation of the Welsh Parliament, and improving information on Scotland and Northern Ireland, this pulls in the text from WriteToThem to improve the postcode page.

image

### DIFF
--- a/www/docs/postcode/repexplain.php
+++ b/www/docs/postcode/repexplain.php
@@ -1,0 +1,19 @@
+<h2><?= gettext("More information") ?></h2>
+
+<?php if ($multi == "scotland") { ?>
+
+    <p><?= gettext("Your MSPs represents you in the Scottish Parliament. The Scottish Parliament is responsible for a wide range of devolved matters in which it sets policy independently of the London Parliament. Devolved matters include education, health, agriculture, justice and prisons. It also has some tax-raising powers.")?></p>
+
+<?php } elseif ($multi == "northern-ireland") { ?>
+
+    <p><?= gettext("Your MLAs represent you on the Northern Ireland Assembly. The Northern Ireland Assembly has full authority over \"transferred matters\", which include agriculture, education, employment, the environment and health.")?></p>
+
+<?php } elseif ($multi == "wales") { ?>
+
+    <p><?= gettext("Your MSs represents you in the Senedd. The Senedd has a wide range of powers over areas including economic development, transport, finance, local government, health, housing and the Welsh Language.")?></p>
+
+<?php } ?>
+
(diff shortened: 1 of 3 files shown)


www/docs/postcode/repexplain.php
```diff
@@ -0,0 +1,19 @@
+<h2><?= gettext("More information") ?></h2>
+
+<?php if ($multi == "scotland") { ?>
+
+    <p><?= gettext("Your MSPs represents you in the Scottish Parliament. The Scottish Parliament is responsible for a wide range of devolved matters in which it sets policy independently of the London Parliament. Devolved matters include education, health, agriculture, justice and prisons. It also has some tax-raising powers.")?></p>
+
+<?php } elseif ($multi == "northern-ireland") { ?>
+
+    <p><?= gettext("Your MLAs represent you on the Northern Ireland Assembly. The Northern Ireland Assembly has full authority over \"transferred matters\", which include agriculture, education, employment, the environment and health.")?></p>
```
dracos, line 9: Can we use “ and ” as a) nicer, b) doesn't need escaping :)

www/docs/postcode/repexplain.php
```diff
@@ -0,0 +1,19 @@
...
+
+<?php } elseif ($multi == "wales") { ?>
+
+    <p><?= gettext("Your MSs represents you in the Senedd. The Senedd has a wide range of powers over areas including economic development, transport, finance, local government, health, housing and the Welsh Language.")?></p>
+
+<?php } ?>
+
+<p><?= gettext("Your MP represents you in the House of Commons. The House of Commons is responsible for making laws in the UK and for overall scrutiny of all aspects of government.")?></p>
+
+<p><?= gettext("You can write to any representative, or your local councillors through <a href=\"http://www.writetothem.com/who?pc=" . $pc . "\">WriteToThem.com</a>")?>.</p>
```
dracos, line 19: https

dracos (changes requested):
Couple of tiny tweaks is all.
More broadly I wondered if the information would be better in the sidebar (replacing the All MPs/MSPs links)? Not much in it, I imagine.

ajparsons:
I think I prefer it the way it is, but here's what that looks like:

image

Given *generally* people want to click through to a specific person - so I think the double column is confusing as to what they should do next (especially as the info column is bigger). If more closely mirroring the WriteToThem design (pairing descriptions with groups of reps), this might work better.

dracos:
Yeah, agree, leave it where it is :+1:

dracos (approved):

PR #1703, opened 2023-04-18
Remove blanket covid text in favour of highlighting specific votes.

ajparsons:
- Covid messaging moved to bottom to be linked to rather than at the top of all the pages.
- Removed from profile page altogether
- Absences for votes in early June 2020 have extra text.
- Policies with a vote in that time have extra text (only 1 for now, one more in new policies)

image

image

### DIFF
--- a/www/docs/style/sass/pages/_mp.scss
+++ b/www/docs/style/sass/pages/_mp.scss
@@ -641,7 +641,6 @@ html.js {
 .vote-description__covid {
     display: block;
     color: #757470;
-    text-decoration: none !important; // override .person-panels .panel a
     font-size: 0.7em;
     line-height: 1.5em;
     margin-top: 0.2em;
--- a/www/docs/style/sass/pages/_mp.scss
+++ b/www/docs/style/sass/pages/_mp.scss
@@ -638,6 +638,16 @@ html.js {
   }
 }
 
+.vote-description__covid {
+    display: block;
+    color: #757470;
+    text-decoration: none !important; // override .person-panels .panel a
+    font-size: 0.7em;
+    line-height: 1.5em;
+    margin-top: 0.2em;
+
+  }
+
 .vote-description__evidence {
   display: block;
   color: #757470;
--- a/www/includes/easyparliament/templates/html/mp/_vote_description.php
+++ b/www/includes/easyparliament/templates/html/mp/_vote_description.php
@@ -8,7 +8,7 @@
         <?= isset($party_voting_line) ? $party_voting_line : '' ?>
     </a>
     <?php if (isset($covid_affected) && $covid_affected) { ?>
-    <span style="font-size:60%">Absences for this policy may be affected <a href="votes#covid-19">COVID-19 restrictions</a>.</span>
+    <span style="font-size:60%">Absences for this policy may be affected <a href="<?= $member_url ?>/votes#covid-19">COVID-19 restrictions</a>.</span>
     <?php } ?>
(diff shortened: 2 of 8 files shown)


www/docs/style/sass/pages/_mp.scss
```diff
@@ -638,6 +638,16 @@ html.js {
   }
 }
 
+.vote-description__covid {
+    display: block;
+    color: #757470;
+    text-decoration: none !important; // override .person-panels .panel a
```
dracos, line 644: I don't understand this line, because there's no text decoration applied to a `p`?

www/includes/easyparliament/templates/html/mp/_vote_description.php
```diff
@@ -7,5 +7,8 @@
         <?= isset($key_vote) ? "$key_vote[summary]." : '' ?>
         <?= isset($party_voting_line) ? $party_voting_line : '' ?>
     </a>
+    <?php if (isset($covid_affected) && $covid_affected) { ?>
```
dracos, line 10: This template is included on profile.php as well as votes.php - just checking you only wanted it to be displayed on the votes.php entries.

dracos (changes requested):
Couple of small queries is all

www/docs/style/sass/pages/_mp.scss
ajparsons: I'll take this out - it was because I was copying the style for the evidence section.

www/includes/easyparliament/templates/html/mp/_vote_description.php
ajparsons: Ah right, these might appear in the summary items. Will add those in.

www/includes/easyparliament/templates/html/mp/_vote_description.php
```diff
@@ -8,7 +8,7 @@
         <?= isset($party_voting_line) ? $party_voting_line : '' ?>
     </a>
     <?php if (isset($covid_affected) && $covid_affected) { ?>
-    <span style="font-size:60%">Absences for this policy may be affected <a href="#covid-19">COVID-19 restrictions</a>.</span>
+    <span style="font-size:60%">Absences for this policy may be affected <a href="votes#covid-19">COVID-19 restrictions</a>.</span>
```
dracos, line 11: Ooh, I was going to say I don't think this URL will work, as main profile is /mp/ID/name/const so it would go to /mp/ID/name/votes - but that does appear to redirect to the right place. But it'd be better if it were `<?= $member_url ?>/votes` I think, to match other uses.

dracos (changes requested):
One suggestion for the votes link. Seems a bit odd to have the policy IDs in 4 different places, but I realise is a special case and it doesn't look like it'd be easily placed in some single policy lookup place anyway?

ajparsons:
I've moved the list into the Policies class - it is possible in future more will be added (if we add another affected range), so is better to keep that simpler.

dracos (changes requested):
Sorry to be a pain - but no need to create a new identical object and call getCovidAffected() in each iteration, can do it once before each loop?
(Seems like Policy objects could be passed around rather than IDs which would presumably simplify this somewhat, but not for now!)

ajparsons:
That's reasonable - for the profile page I've pushed it right to the top given the same thing works for all references.

dracos (approved):
Thanks :)

Issue #1689, opened 2023-01-17
Modular space for announcements and content


ajparsons:
What I'd like is a more nuanced version of our banner announcements, where we can more easily roll out a rotation of things to promote through the site - and have these running all the time (the banner loses its impact if run all the time).

For instance, we might have multiple messages in rotation:

- Donate to support mySociety's work
- Read latest blog post [x]
- Take part in our user survey
- Cross-site promotion of CAPE/WDTK, etc

These messages would ideally be in the admin with a light-weight interface rather than deployed through the code. 

I'd like to integrate into:

- Homepage
- Email alerts
- MP summary pages

For a display pov, these different locations suggest slightly different designs. An extra complication is we might sometimes want some elements that have custom html or forms (e.g. subscribe to newsletter), we we might (see https://github.com/mysociety/supporters/issues/2) want to experiment with integrating donation forms directly into sites. This feels special-caseable once the basic setup is in place. 

In terms of the fields to store, an announcement has a:

- Title
- Short description
- URL
- Large image
- Thumbnail
- Format (for future special casing - e.g. 'link', 'donate', 'newsletter').
- Currently live (booelan)

The first steps is some light mock-ups of what this would look like, and then scope out with design what is possible with available time.

zarino:
Notes from my discussion today with @lucascumsille:

**Potential locations on homepage:**

- Banner below the navbar (like Scorecards methodology banner)
- Replacing the (not used much) "random debate" and "topics in the news" below the hero

**Potential locations on MP page:**

- Banner below the navbar
- Replacing the yellow COVID panel
- Below the in-page nav sidebar (making the sidebar wider too)

@lucascumsille is going to mock up how a blog post (as just one example of a type of content) might look in those five positions, so we can see whether that sparks any ideas.

Meanwhile, we noted that donations feel like a different *kind* of content to blog posts etc – they feel less timely. @lucascumsille wondered whether we should treat them differently, visually. Thoughts, @ajparsons ?

We also wondered whether we actually *need* an image? Especially on our blog posts, the generic unsplash images always feel like filler. What’s the image for? If it’s to attract attention, we wondered whether that could that be achieved through colour alone?

ajparsons:
I don't mind different approaches on the image - possibly a generic image for the blog series overall might be fine rather than for each post.

lucascumsille:
I have made some mockups for the TWFY homepage:

### Banner for navbar
**Pros**

[...]

PR #1705, opened 2023-04-19
1689 announcement

lucascumsille:
Fixes:  https://github.com/mysociety/theyworkforyou/issues/1689

### BANNER NAV
A couple of observations:
I moved it below the main navbar, because at the beginning I didn't notice where the banner was seating, it took me a bit to realise that it was on the top of the page, because it gets camouflaged with the browser bar(I use dark themes). I reduced the font size and increased the padding, so it looks like it has more breathing space and the text doesn't look too bold.

### Option 1
option 1

### Option 2
Option 2

### Option 3
For option 3 I decided to move it to the top because, when using a button it looks more prominent. 
Option 3

I can just drop commits so let me know which options you think would work better. Personally I prefer option 2, gives enough highlight to the banner content without disturbing  the main navbar content and hero.


ajparsons:
2 is good!

lucascumsille:
@ajparsons the latest commit is for the featured blog below the hero section. But I need to add the "topics that might interest you" and also I'll ask for your help with enabling/testing this commit, because it has been a mix of editing the live version and adding the changes into the templates and CSS, so I'm not 100% sure it will work.

Screenshot 2023-04-19 at 10 33 42

ajparsons:
To check we're on the same page:

The featured option that currently exists refers to a featured debate, rather than a arbitrary URL. I'll need to implement a new backend bit to talk to that front end (doesn't currently exist) - or at least a dummy one for the moment.

Going back to my original commit, I was talking about:

- Title
- Short description
- URL
- Large image
- Thumbnail
- Format (for future special casing - e.g. 'link', 'donate', 'newsletter').
 - Currently live (booelan)

Image wise, which ones do you think you need?

ajparsons:
I've added a temp json file based way of sourcing the information.

My path of least resistance approach in the long term would just be to store the json in the current banner storage in the database (will investigate that once the front end stuff is more finalised).

ajparsons:
I was wondering about that.

Not sure if it's best to have multiple 'announcements' and be able to target them to multiple places with an extra line - or have a third option for sidebars.

Will have a think and a play with this the morning. Thanks so much for this!

lucascumsille:
> So something's not quite working on the sidebar:
> 
> image
> 
> I think a few things that are already confusing:
> 
> * That little side menu is now too small to be noticed. I think that's _ok_ because that was already true, and I'd like to move the summary page links to the top bar anyway.
> * But if we're doing that, and putting our content just below someone's face - I think we need a little separator just to say what it is. Mocked up a rubbish version above ('from mysociety'). Do you agree or is this not a problem?

Actually this reminds of Zarino's discovering( I think it was on the the Local hub) that some users didn't realise the list on the side bar was to browse content. 
Personally I prefer it on the side, because if you have a big list putting it on the top is a bit tricky, but not sure if that would be  a possible scenario here.

I made a quick mock up. Our interactive colour in TWFY is green and hopefully the title will give users enough clues to about that section.
Screenshot 2023-04-24 at 07 04 05

Regarding the From mySociety I think it makes sense. Actually I would probably change it for mySociety logo. 

Let me know what you think and I can add some commits to this.

ajparsons:
Yes, that green bit is great. Also good point on the logo.

If it's easy enough to do - yes please, could you add a commit with those?

ajparsons:
This PR adds a more general announcement feature to TheyWorkForYou.

- This extends the current Banner approach to allow multiple banners to rotate at once.
- New 'announcements' can be made in three locations: the homepage, MP sidebar, and a donation box in the MP sidebar.

The three locations style content in slightly different ways:

- 'homepage' - uses all the content, but is quite wide rather than in a sidebar.
- 'sidebar' - uses image, title and button options, but does not display the additional 'content' option.
- 'donation' - uses title, content and button options (no image). 

The backend has been adapted to store JSON that stores details of multiple options. Format is shown here. Invalid Json throws an error (but is not checked against a specific schema):

- Announcements https://gist.github.com/ajparsons/ff8029afd333122775864b58fbe29433
- Banners https://gist.github.com/ajparsons/778e752ac110f595e1230ee908be9c5e

Dev helpers:

For consistency with the script-to-rule-them-all approach elsewhere the 'script' directory as opposed to 'script', stores several useful management commands for use in dev. Currently this contains `watch-css` and `create-superuser'.

Other styling changes:

- The announcement section on homepage has demoted the 'featured' debate to below the search bar.
- A new green 'Browse content' helps draw attention to the sidebar on MPs profiles.

image

image

image

PR #1711, opened 2023-04-24
New announcement system for TheyWorkForYou

ajparsons:
This PR adds a more general announcement feature to TheyWorkForYou.

- This extends the current Banner approach to allow multiple banners to rotate at once.
- New 'announcements' can be made in three locations: the homepage, MP sidebar, and a donation box in the MP sidebar.

The three locations style content in slightly different ways:

- 'homepage' - uses all the content, but is quite wide rather than in a sidebar.
- 'sidebar' - uses image, title and button options, but does not display the additional 'content' option.
- 'donation' - uses title, content and button options (no image). 

The backend has been adapted to store JSON that stores details of multiple options. Format is shown here. Invalid Json throws an error (but is not checked against a specific schema):

- Announcements https://gist.github.com/ajparsons/ff8029afd333122775864b58fbe29433
- Banners https://gist.github.com/ajparsons/778e752ac110f595e1230ee908be9c5e

Dev helpers:

For consistency with the script-to-rule-them-all approach elsewhere the 'script' directory as opposed to 'script', stores several useful management commands for use in dev. Currently this contains `watch-css` and `create-superuser'.

Other styling changes:

- The announcement section on homepage has demoted the 'featured' debate to below the search bar.
- A new green 'Browse content' helps draw attention to the sidebar on MPs profiles.

image

image

image


ajparsons:
Things not covered in this PR:

- The example JSON format has a 'lang' option, with the intention being after merging with the Welsh language support we could make banners language sensitive.
- This doesn't cover - but probably should - adding announcements to alerts. Will do this as a future PR because testing it is slightly different.

classes/Model/AnnoucementManagement.php
```diff
@@ -0,0 +1,168 @@
...
+                ':banner_text' => $text,
+                ':editoral_option' => $editoral_option
+            )
+        );
+
+        if ( $q->success() ) {
+            if ( trim($text) == '' ) {
+                $text = NULL;
+            }
+            $this->mem->set('banner', $text, 86400);
```
dracos: This should be $editorial_option

classes/Model/AnnoucementManagement.php
```diff
@@ -0,0 +1,168 @@
...
+                }
+                $this->mem->set($editoral_option, $text, 86400);
+            }
+        }
+
+        return $text;
+    }
+
+    public function set_text($text, $editoral_option) {
+        $q = $this->db->query("REPLACE INTO editorial set item = :editoral_option, value = :banner_text",
```
dracos: As not just banner now, want a more generic placeholder name?

classes/Model/AnnoucementManagement.php
```diff
@@ -0,0 +1,168 @@
...
+function select_based_on_weight($items)
+{
+    # banners have a weight attribute, which is the probability of being selected
+    # the higher the weight, the higher the probability
+    $total_weight = 0;
+    foreach ($items as $item) {
+        $total_weight += $item->weight;
+    }
+
+    $random_number = rand(0, $total_weight);
```
dracos, line 60: If I've understood this function correctly, the 0 should be 1? As otherwise the first item in the list will have an extra 0 slot to get picked (e.g. three items, each of weight 1, rand will be 0, 1, 2, or 3, and first item will be 50%, others 25%).

www/docs/admin/banner.php
```diff
@@ -2,19 +2,30 @@
 
 include_once '../../includes/easyparliament/init.php';
 
-$this_page = 'admin_banner';
 
 $db = new ParlDB;
-$banner = new MySociety\TheyWorkForYou\Model\Banner;
+$banner = new MySociety\TheyWorkForYou\Model\AnnoucementManagement;
```
dracos, line 7: Announcement is spelt wrong in the filename/package name (and some of the commit messages).

www/docs/admin/banner.php
```diff
@@ -2,19 +2,30 @@
...
 $PAGE->stripe_start();
 
+$editoral_option = 'banner';
+if (get_http_var('editoral_option')) {
+    $editoral_option = get_http_var('editoral_option');
+}
+
+if ($editoral_option == "banner"){
+    $this_page = 'admin_banner';
+} else if ($editoral_option == "annoucements") {
```
dracos, line 19: `elseif` in PHP (though don't think it makes a difference).

script/create-superuser
```diff
@@ -0,0 +1,113 @@
...
+        ":confirmed" => $details["confirmed"],
+    ]
+);
+
+$q = $db->query("select * from users");
+
+if ($q->success()) {
+    echo "\nUser created: Use the following details to log in:\n";
+    echo "Username: twfytestadmin@mysociety.org\n";
+    echo "Password: password\n";
```
dracos, line 107: Presumably it's not this email/password

script/create-superuser
```diff
@@ -0,0 +1,113 @@
...
+    "confirmed" => 1,
+];
+
+$registrationtime = gmdate("YmdHis");
+$passwordforDB = password_hash($details["password"], PASSWORD_BCRYPT);
+
+# get database connection
+$db = new ParlDB();
+
+$q = $db->query(
```
dracos, line 55: Rather than this whole query, could we use the user->add($details, $confirmation_required=false) function that exists already? Looks like you pass in a hash of various entries, but it does some for you (e.g. password hash).

www/docs/style/sass/app.scss
```diff
@@ -162,12 +162,16 @@ a {
...
 }
 
-a:hover,
-a:focus {
+a:hover {
     text-decoration: underline;
     color: darken($colour_primary, 10%);
 }
 
+a:focus {
```
dracos, line 170: The change to focus is nothing to do with this commit, should be its own/different commit.

www/includes/easyparliament/templates/html/header.php
```diff
@@ -185,6 +175,17 @@
         </div>
     </nav>
 
+    <?php if (isset($random_banner) && $random_banner ) { ?>
+      <div class="banner">
+          <div class="full-page__row">
+              <div class="banner__content">
+                  <?= $random_banner->content ?>
+                  <a class="button button--small content__button <?= $random_banner->button_class ?>" href="<?= $random_banner->button_link ?>"><?= $random_banner->button_text ?></a>
```
dracos, line 183: It would be better if this would work even if the JSON lacked button_link/text/class and only had content, by testing to see if it had button info.

www/includes/easyparliament/templates/html/homepage/featured-content.php
```diff
@@ -0,0 +1,22 @@
...
+    <a href="<?= $featured_item->url ?>"><h3 class="featured-content__title"><?= $featured_item->title ?></h3></a>
+    <p class="featured-content__description"><?= $featured_item->content ?></p>
+    <a class="button featured-content__button" href="<?= $featured_item->url ?>"><?= (isset($featured_item->button_text) ? $featured_item->button_text : "Read more") ?></a>
+</div>
+</div>
+
+<?php } else { ?>
+
+    <?php if ( count($featured) > 0 ) {
+                        include 'featured.php';
```
dracos, line 18: This means the featured.php will be shown twice on the page if there's no homepage announcement.

www/includes/easyparliament/templates/html/homepage/featured-content.php
```diff
@@ -0,0 +1,22 @@
+
+<?php $featured_item = $annoucement_manager->get_random_valid_item("homepage"); ?>
+
+<?php if ( $featured_item ) { ?>
+<h2>The latest</h2>
+<div class="featured-content__wrapper">
+<a href="<?= $featured_item->url ?>"><img class="featured-content__image" src="<?= $featured_item->thumbnail_image_url ?>" alt="<?= $featured_item->thumbnail_image_alt_text ?>"></a>
```
dracos, line 7: Again, can we have it check/work if there's no thumbnail etc

www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -15,14 +15,19 @@
...
-                <?php } ?>
-                <ul>
-                    <li><a href="<?= $member_url ?>/votes">Back to all topics</a></li>
-                </ul>
+                <div>
+                    <?php if ( isset($policydivisions) && $policydivisions && count($policydivisions) == 1 ) { ?>
+                        <p class="policy-votes-intro">
+                            How <?= $full_name ?> voted on <?= $policydivisions[array_keys($policydivisions)[0]]['desc'] ?>.
+                        </p>
+                        <?php } ?>
```
dracos, line 23: If you're going to change indents (not sure it matters much in this type of file), this one and below is wrong :)

www/includes/easyparliament/templates/html/mp/recent.php
```diff
@@ -61,15 +61,19 @@
...
+                    <ul>
+                        <?php foreach($sidebar_links as $date) { ?>
+                          <li>
+                              <a href="#<?= strftime('%Y-%m-%d', strtotime($date)) ?>">
+                                  <?= strftime('%e %b %Y', strtotime($date)) ?>
+                              </a>
+                          </li>
+                        <?php } ?>
+                      </ul>
+                      <?php include '_featured_content.php'; ?>
```
dracos, line 75: presumably donation not in this one on purpose?

classes/Model/AnnoucementManagement.php
```diff
@@ -7,6 +7,48 @@
...
+
+    // set default end_time (in the future)
+    if (isset($item->end_time)){
+        $end_time = $item->end_time;
+    } else {
+        $end_time = "2100-01-01T00:00:00";
+    }
+
+    return $published
+        && $language == 'en'
```
dracos: I realise you say will need updating when merged with language branch, but this line should be using $current_language already.

dracos (changes requested):
Nothing major, but a few bugs and things to fix

www/docs/admin/banner.php
ajparsons: In a lot of places actually.

script/create-superuser
ajparsons: This probably *can* be done, but I was having trouble getting useful errors from why it wasn't working for me, so I gave up and brought the relevant code here till it was working.

www/includes/easyparliament/templates/html/mp/recent.php
ajparsons: As I remember, the long side bar pushed it out of visibility anyway. so I didn't include it for neatness.

www/includes/easyparliament/templates/html/homepage/featured-content.php
ajparsons: Yes, I couldn't see a good way around this because some content we want in this box above the search (the alerts link) on the right all the time. My assumption (which hopefully won't be wrong) is we'll always have something there, and if not, it won't be immediately apparent to most people landing it's duplicated above and below the search.

www/includes/easyparliament/templates/html/header.php
ajparsons: Yep, made this conditional on button_link.

ajparsons:
Should have fixed all of the big things and it looks like it works okay. 

If seems alright, I'll fix up the history to something sensible.

www/includes/easyparliament/templates/html/mp/recent.php
dracos: Don't follow this, sorry, as /votes sidebar seems far longer than /recent?

classes/Model/AnnouncementManagement.php
```diff
@@ -18,7 +18,7 @@ function is_valid_item($item)
     if (isset($item->lang)) {
         $language = $item->lang;
     } else {
-        $language = "en";
+        $language = $current_language;
```
dracos, line 21: Think this is the wrong "en" to change, I'm afraid - defaulting to English if not present in the item is fine, but it's the check at line 46 that should check $language against $current_language, not "en".

www/includes/easyparliament/templates/html/homepage/featured-content.php
dracos: We could set a global flag so if it's being included the first time and the second time checks if the flag has been set and doesn't show it again, but as you say, no biggie.

dracos (changes requested):
Just the one bug I think, and one query, is all.

classes/Model/AnnouncementManagement.php
ajparsons: Doh, yes.

www/includes/easyparliament/templates/html/mp/recent.php
ajparsons: Ah, I had it out of the voting summaries on purpose, recent is an oversight.

ajparsons: Yeah, this is an oversight - I was leaving it out of a different page (have just put it everywhere for consistency now).

www/includes/easyparliament/templates/html/homepage/featured-content.php
ajparsons: I've done this just so it isn't a surprise later (or when we first turn it on). Looks like it works fine.

ajparsons:
OK! Should be almost there.

www/includes/easyparliament/templates/html/index.php
```diff
@@ -93,6 +94,7 @@
             </div>
         </div>
 
+        <?php if ( $featured_debate_shown == false ) { ?>
```
dracos, line 97: Should this if go round just the lines 101-105, otherwise the topics aren't shown if featured has been shown above?

dracos (approved):
There's one issue as to whether the new `if` is around a bit too much, but not very important :)

www/includes/easyparliament/templates/html/index.php
ajparsons: Yep, can do that.

ajparsons:
Okay! History tidied up (and spelling fixed). Will merge in this afternoon!

PR #1715, opened 2023-05-04
GA4 analytics

ajparsons:
Uses the cookieless GA4 approach described in https://docs.google.com/document/d/1ZXbkXoLE0eLnUWJaVv_8xpnVOlOHGCxPP_Uo7cfoFJo/edit

- Adds GA4 config
- monkey-patches `document.cookies` to prevent ga setting cookies.

This PR originally used the measurement protocol - but those commits will come out in a final squish.

### DIFF
--- a/www/docs/js/main.js
+++ b/www/docs/js/main.js
@@ -9,46 +9,15 @@ var trackEvent = function(eventName, params) {
       dfd.resolve();
   };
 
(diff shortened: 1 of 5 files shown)


struan (approved):
looks fine, doesn't seem to set a cookie. Seems to send things to Google :)

PR #1724, opened 2023-06-15
Devolved navigation

ajparsons:
# Adds a multi Parliament menu to home page.

London excluded for moment because it needs some sorting out.

image

# Link to recent debates in top bar of Senedd

image

# Contextual parliament selection for search and alerts

Because these are general pages, they don't have the devolved parliament in their parent tree.

This adds additional checks to see if the alert is linked to a specific person, or the search is limited to a particular parliament, and changes the top bar:

image

image


dracos (commented):
Only thing is I think it's bit unclear the heading of the name is a link (to the debates) given the list then underneath - could the heading not be a link and have debates as the first link underneath it?

ajparsons:
Yep! 
image

I've also because I've noticed it updated the description of the NI Assembly (was a few years out of date).

Issue #1727, opened 2023-06-27
Cym - Move language selector to Top Right corner


it3986:
Convention is that the Language Selector on multi-lingual websites is consistently in the Top Right corner.
On Desktop it is currently in the Top Right nav but is the left-most item and so not in this sweet-spot.
On Mobile it is currently buried in the hamburger nav so not instantly visible on a mobile device that the other language is available.

Same problem in both languages on any of the pages now available in both Welsh and English.

Guidance on this from the Office of the Welsh Language Comissioner is on page 57 of their guidance on Technology, Websites and Software:
https://www.welshlanguagecommissioner.wales/media/lxechbz0/technoleg-gwefannau-a-meddalwedd-technology-websites-and-software.pdf

Screenshot of anticipated lang switcher placement using https://cy.theyworkforyou.com/mss/?o=p as an example:
![image](https://github.com/mysociety/theyworkforyou/assets/4873884/665bb396-f768-4a5d-bed9-10e0ff5d36d4)

it3986:
Screenshot of Language switcher within hamburger menu on mobile:
![image](https://github.com/mysociety/theyworkforyou/assets/4873884/731f04e0-82d3-4e6d-a717-d9366c91a78e)


Screenshot of example of good practice (above hamburger menu) from https://naturalresources.wales/?lang=en
![image](https://github.com/mysociety/theyworkforyou/assets/4873884/6e58cd53-6137-4d8d-8eef-f533c9c9763b)

PR #1749, opened 2023-12-08
Donate and language switch in top bar nav

ajparsons:
This PR:

* Adds a link to the mySociety donate page in the top bar.
* Moved the language selector to the top right (https://github.com/mysociety/theyworkforyou/issues/1727)

Doesn't entirely solve #1727 because of it doesn't change the hamburger approach.

Donate approach is a bit hacky - didn't want to add it to the metadata file because it's not a page. Is weird to have an external link in the normal navigation - but it's how we've got things set up at the moment.

### DIFF
--- a/classes/Renderer/User.php
+++ b/classes/Renderer/User.php
@@ -27,6 +27,21 @@ class User
 
     private function setupNavLinks() {
         $this->data['user_nav_links'] = array();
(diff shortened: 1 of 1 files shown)


dracos (approved):
👍

Issue #1745, opened 2023-11-21
Show in ‘more votes screen’ when a vote for one policy is also associated with other policies


ajparsons:
We should make it clearer on the ‘more details’ page when a vote for a policy is also related to a vote for a different policy. 

E.g. we should end up with something like this:

<img width="238" alt="image" src="https://github.com/mysociety/theyworkforyou/assets/8157058/41c02a73-08e8-452a-81de-95ff69ce0eea">

Reasoning:

Our display suggests that MPs are voting on policies in isolation. In practice, for big votes on legislation, many different policies are related to the same vote - but they are not decided independently. MPs do not have the chance to vote for A, but not B - the final bill packages A and B together. This is in part because they are in a related area and get packaged in the same bill (e.g. a Justice bill), but doing it is also a deliberate tactic by the government to force the opposition to vote against popular things they might support independently. 

There are clear problems in accurately and clearly presenting information, the more 'one vote' behind the scenes, is being presented as 'many issues' in front of the house. In practice, omnibus bills are a real problem for our current model of voting, which we resolve by not duplicating where possible, but also by just choosing to be ok with a certain amount of duplication.

The more we express this overlap in the interface, the more we can live with this problem. This requires some light data work in TheyWorkForYou, but works based on existing data and I have a draft version of this working in a branch. 

A quick draft of this can be found in this branch: https://github.com/mysociety/theyworkforyou/compare/master...policy-page-show-intersections.

I think this approach isn’t the most optimal however, and is adding more queries than is necessary to load the page (but also, might just be fine).

PR #1751, opened 2023-12-14
Minor changes to division display

### LINKED ISSUE #1744: Change display of strong/weak votes
Following the public whip, we have the concept of strong/weak votes. Currently in the ‘per policy/per MP’ screen (‘show  votes’ button) - we show ‘strong’ votes by default, with ‘less important’ votes hidden behind a button - which then get displayed in a little indent. 

image

image

In general, we want to move towards language around ‘scoring’ and ‘informative’ votes - where weak votes are worth nothing to 
score, but can be included here. But before we do that change in language, we can do some more fiddly work with the display.

**What we want is to split into two lists** that are displayed on the page (rather than indented behind a button), where there can be explanatory text for each section describing what ‘strong’/’weak’ means in the current context. 

A quick draft of this can be found in this branch: https://github.com/mysociety/theyworkforyou/compare/master...policy-page-show-intersections

dracos:
Fixes #1744 and #1745. Differences from the draft branch:
* Changed Divisions class to always have access to full set of Policies (seemed needless to restrict this);
* Got rid of more knock-on unneeded code (anything to do with weak_count or the JS for that button);
* Changed related_policies to be a two-entry array of array rather than separate two-entry arrays;
* reduced number of close/open PHP blocks a bit.

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -621,7 +621,6 @@ class Divisions {
             if ( !array_key_exists($policy_id, $policies) ) {
                 $policies[$policy_id] = array(
                     'policy_id' => $policy_id,
-                    'weak_count' => 0,
                     'divisions' => array()
                 );
                 if ( $this->policies ) {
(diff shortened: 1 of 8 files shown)


ajparsons (approved):
Works great on a click around!

I've made some small cosmetic changes - if you're happy with those - it's all good.

dracos:
Changed since merge - only show related if there's a policy involved; minor styling; strip links from policy titles.

Issue #1747, opened 2023-11-21
Dynamic list of policies


ajparsons:
Currently the policies that TWFY draws on are hardcoded as a series of arrays (different arrays for each policy group, and one to assign labels). 

Given policy management otherwise happens in twfy-votes (including moving items between draft and active, and adding to groups) - it’d be nice to avoid having to change things in multiple places. 

Ideally on deploy theyworkforyou would source a json from twfy-votes, and this would set the current policies.

dracos:
We do also have a policies table in TWFY that is populated and has an admin interface, and has a "title" field, which I'm not sure how it differs from the hardcoded titles in Policies.php... I guess that's for display as well somewhere?

Happy to source a JSON for a current list with groups if that's the best place to do it.

ajparsons:
The `/policies/commons/active/all.json` endpoint will give you all currently active policies for the commons, which is what we need at the moment.

Relevant bits for each item in `policies` are:

* id
* name
* context_description (the bold bit that's hardcoded at the moment)
*  policy_description
* groups (a policy can be multiple groups)

Issue #1743, opened 2023-11-21
Switch JSON download from public whip to twfy-votes


ajparsons:
Public Whip creates popolo JSON feeds of all votes associated with a policy, which TheyWorkForYou imports to create an association between a division, and update any division descriptions based on the content in the Public Whip.

TheyWorkForYou Votes recreates these popolo feeds (although part of this ticket is verifying they work as intended). 

Currently, the process is  done in several phases.

- Parlparse creates XMLs for Public Whip to parse.
- Parlparse downloads popolo json from Public Whip.
- TheyWorkForYou ingest some of these jsons from ParlParse.

The simplest approach would be to adjust parlparse to pull from a new source (might need a new json feed from twfy-votes to give a list of available policies (twfy has a list of IDs, but parlparse doesn’t).

However, the XML export from PW to TWFY does not do this double loop. Do we want to similarly switch to TheyWorkForYou directly querying the JSON and cut it out from ParlParse? 

(The current approach does mean the json is then publicly available).

dracos:
List of policies is presumably linked with https://github.com/mysociety/theyworkforyou/issues/1747

If the data is going to be available elsewhere anyway, I don't see an issue with changing it. It looks like the fact it requests policy XML for the MP info and then vote info from the JSON is purely historical, I see no reason to maintain that if it can be simplified. I assume ideally you'd want to request one policy JSON URL that returned the MP information and the vote information together that could then be imported, does that make sense with the twfy-votes repo?

ajparsons:
Yeah, that'd be fine. 

So we'd get towards something like:

```
policy_details: {meta information about the policy}
votes: [information about how each person voted for each vote]
alignment: [overall information about alignment for each person with the policy]
```

The existing popolo-esque view can be adapted to add that.

dracos:
Using 810 as a comparison, the outputs are different at present, I'm afraid. For example, 2005-06-14 division 12 in PW JSON has aye/no/both/absent of 199/292/0/153, whereas votes has 233/312/0/105. The 'text' of the motion is also different, PW has "National Lottery Bill (Reasoned amendment on second reading)", votes only has "National Lottery Bill". Or 2008-04-28/155 has 223/311/0/11 vs 265/335/0/50 and "Finance Bill - Clause 21 - Amusement Machine licence duty" vs "Orders of the Day - Clause 21 - Amusement Machine licence duty".

For the alignments, checked the first dozen or so...
[...]

PR #1752, opened 2023-12-15
Fetch policy/alignment data from new location.

dracos:
For https://github.com/mysociety/theyworkforyou/issues/1743 (I ignored the comparison stuff in the JSON for the moment, assume I'll try and understand that at some later point; hope I've picked the right fields).
This stops fetching dream MP XML and policy JSON from PW, instead fetching JSON only from twfy-votes that contains the policy votes and alignment information. The import script no longer imports divisions, assuming they've already been imported elsewhere, only policies (if missing), policy/division links, and policy alignments.

Also fixes #1747.


dracos:
This creates a local JSON file of the policies from the list, rather than change the DB at present.
It keeps 'summary' being hard-coded at present.

ajparsons:
Putting my notes back in the PR for history.  Following review comment explains new commits. 

All policy information is now loaded through the json2db.pl - running this on new instanance seems to work fine. 
 

## Reference to twfy-votes

URL is not hardcoded - but added as an CONF setting 'TWFY_VOTES_URL'.

When we have a semi-permenemnt home we should set a default on this - although still useful to have as a config setting for testing. 

## Meta infomration

There is a new file /scrapedjson/policies.json that stores the polices data. 

This is a reprocessed version of the /policies/commons/active/all.json in twfy-votes. 

This specifically stores the descriptions of sets - and the assignment of specific polices to sets. This is referenced when the Policies object is constructed, and which then can be used as the old manual assigments were to display information in views. 

Note: it would be useful for testing if twfy-votes let you reference both active+candidate in this URL - now https://github.com/mysociety/twfy-votes/issues/24

## Loading policy information

Assume that no new division information is being returned, and so no longer do update checks based on that.

Person policy information is set in the 'personinfo' table as keys - continuing the dreammp setup currently used.

For each policy_id - fetching the twfy-compatitible popolo and reading the info in.

Note: we are no longer storing this json locally as part of parlparse.

Process motions - stash in database the policydivisions information (connection between policy and division) - either adding or update existing. 

Note: maybe here the function should be process_policydivision for clarity? This is following the existing 'motionadd, motionupdate' quieries - both of which refer to policydivisions table and could also be renamed. Addressed in follow up commit. 

Process alignments - for each person per vote - create or update distance, both_voted, and absent num numbers in the person info table.

Meta note: We're loading policies one at at a time from twfy-votes, mirroing pw API - in principle this could be one big policies file. I don't see a massive reason to do this at this point but could. 

Meta note: Still pulling in attendance/rebellion info from Public Whip - which I don't think we're displaying anywhere.

PR #1754, opened 2023-12-18
Upgrade phpunit

dracos:


dracos:
Not sure about the SQL change for the party cohort group by...

dracos:
Though if we're dropping that query entirely, guess that doesn't matter so much!

ajparsons (approved):
Meta-yes - the cohort query is redundant soon - but I think it makes sense. 

As I understand it:

This is fixing a problem in the query where we want to ensure the cohort key is membership dates and party IN ORDER - and the extra subquery here is needed to ensure that (the order by is the wrong way to do it). 

This makes sense to me - according to past me it doesn't matter for party switchers that their cohort key is the original party (because it's just trying to create a unique set), so that's fine. 

I've rerun the party policy generation on this branch and on a quick check it puts out the right number of cohorts and public numbers seem to be ok. So if that was the only unclear bit (proof of the rest of it is that the tests pass?) - can be merged.

ajparsons:
Fixup commit for getRecentDebatesWithDivisions: where changing some to mins affected the field names used later on.

Translation currently fails here because of 'Maximum term length is 255 characters' in master. Can be fixed separately, but not an error here.

classes/Divisions.php
```diff
@@ -161,9 +161,9 @@ public function getRecentDebatesWithDivisions($number = 20, $majors = null) {
             $debate_gid = fix_gid_from_db($debate['debate_gid']);
             $anchor = '';
             if ($debate['c'] == 1) {
-                $anchor = '#g' . gid_to_anchor(fix_gid_from_db($debate['gid']));
+                $anchor = '#g' . gid_to_anchor(fix_gid_from_db($debate['debate_gid']));
```
dracos, line 164: I think this needs to remain gid and the "min(h.gid)" needs an "as gid" adding, rather than switch to a different gid.

PR #1755, opened 2023-12-19
Remove party cohort code.

### LINKED ISSUE #1742: Switch TheyWorkForYou’s internal calculation of comparisons to twfy-votes
The XML files from twfy-votes also contain additional fields around comparison with parties. 

This can in principle replace the ‘PartyCohort’ generation approach in TheyWorkForYou with a faster and more maintainable approach, and remove complexity from the TheyWorkForYou codebase.  

Implementing this requires changing the current sourcing of party comparison scores from TheyWorkForYou’s database, to using entries stashed in the key lookup from the XML import. 

This is not quite a prerequisite (but would help a lot) with the change to the public whip comparison algorithm.

Options include:

1. Don’t do this. Make any adjustments to the comparison algorithm in TheyWorkForYou at same time and maintain both approaches. 
2. Change the data source, but leave the PartyCohort approach in the code base. 
3. Change the data source, remove duplicate approaches from the code base.

I think there’s a good case for at least 2, and would leave 2 vs 3 up to people with a better sense of the pros and cons. 

--

A reason to keep the current approach intact would be as a third cross-check against the two approaches used in TWFY-votes. But this is practically a bit awkward - and wouldn’t be subject to the same automated cross-checking. There is a broader question about improving the twfy-votes test suite based on the TheyWorkForYou examples. TheyWorkForYou has scenario tests on fake data - whereas twfy-votes tests by applying two different approaches (following the same principles) on live data and checking for alignment.

dracos:
We store/use the figures returned externally. Some of the tests are therefore now only testing that what is stored in the fixture is displayed, rather than any calculations involved. Should fix #1742. Is on top of other PRs.


ajparsons:
This looks good - if you're happy with my amends:

- added autocommit settings when loading alignment information to speed that up. 
- readded vote/division populator for dev setup from json.

ajparsons (approved):

dracos:
Feels like I have to recombine/refactor those two scripts, given how much they do overlap :)

ajparsons:
> Feels like I have to recombine/refactor those two scripts, given how much they do overlap :)

I've recombined these with a switch in a fixup commit - if that's good I can squish before merging.

scripts/json2db.pl
```diff
@@ -189,5 +231,151 @@ sub process_alignments {
...
+            $mp_id_num = $vote->{id};
+            $mp_id_num =~ s:uk.org.publicwhip/person/::;
+            next unless $mp_id_num;
+            if ( $mp_id_num !~ /^[1-9]\d+$/ ) {
+                print "$mp_id_num doesn't look like a valid person id - skipping vote for $motion_id - " . ($dreamid || "") . "\n";
+                next;
+            }
+
+            # if we've seen this motion before then don't process it, however we want
+            # to make sure that the strong vote processing below happens so we still
```
dracos, line 350: This comment doesn't apply as the strong vote processing is no longer present below. (Don't really know if that means you could simplify this code somehow)

dracos (commented):
One query, looks fine otherwise. More generally, though, just to check I understand - what's the reason to use this over the xml2db.pl import of divisions? Do you want to get votes in separately?

scripts/json2db.pl
ajparsons: Hmm, I have no idea. Logically yes, but I don't really want to pull this bit I don't really understand apart and at worse it's just taking slightly longer.

ajparsons:
Sure - basically it's a speed of creating a working environment from scratch so the codespace prebuild takes an hour rather than a day.

For debates, I don't mind having a subsection (loading just one month), but for double checking the voting records I want to have all relevant votes, and to do that  means downloading and importing thousands of days of debates and votes, which significantly increases the build time. 

Ideally, I'd just do that - but given the dev setup was working with the json import, and twfy-votes has continuity with the old format - may as well keep it.

dracos:
I can't approve my own PR, but happy with it!

PR #1756, opened 2023-12-19
1738 move votes from profile page

### LINKED ISSUE #1738: Move ‘significant’ votes from profile page to be top element on voting summary  page.
Currently, the voting records are in two places. 

- There is a voting records page for each MP, which contains all policies and party comparisons. 
- On the MP profile page there is a section that pulls out significant votes - and will reset to random votes when that isn’t available (or for past MPs).

I want to lose this summary on the home page and move all voting records information to the same view. 

Reasoning:

- We default to ‘random votes’ for quite a lot of MPs (new MPs, or loyal MPs) - this is not an important enough bit of information to justify prime position on the summary page.
- Having all the information in one place reduces duplication in the helper text on what we’re highlighting and why. 
- We can lose the random votes fallback - as more information immediately follows. 
- Keeping this in one view lets us more easily handle multiple voting records when people change chamber in future.

dracos:
Fixes #1738, I think.

### DIFF
--- a/classes/Member.php
+++ b/classes/Member.php
@@ -472,7 +472,7 @@ class Member extends \MEMBER {
         }
     }
 
-    public function getPartyPolicyDiffs($partyCohort, $policiesList, $positions, $only_diffs = false) {
+    public function getPartyPolicyDiffs($partyCohort, $policiesList, $positions) {
         $policy_diffs = array();
         $party_positions = $partyCohort->getAllPolicyPositions($policiesList);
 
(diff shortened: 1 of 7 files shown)


ajparsons:
This works as I'd expect - the one thing I'm thinking about is if we need to keep the boilerplate when someone has no significant differences - e.g. to say the 'they are a blah MP, and generally vote like other blah MPs. In all policy areas below, they have generally voted the same as their party' bit.

Will have a think.

ajparsons (approved):
Added an extra panel for MPs with no differences that we might expand in future.

Otherwise this looks great! Approving, but thinking might not want to merge this one until the data update is at least partially in place (so some disclaimers can be dropped).

ajparsons:
Closing this one in favour of https://github.com/mysociety/theyworkforyou/pull/1774

PR #1762, opened 2024-02-06
Display agreements

ajparsons:
This PR adds basic support for displaying agreements in the Policy-for-MP page:

image

This doesn't directly deal with if agreements score for policies - that can be enabled after https://github.com/mysociety/theyworkforyou/pull/1755 is merged however.


www/includes/easyparliament/templates/html/mp/divisions.php
```diff
@@ -50,17 +52,29 @@
                         # does all divisions, but generally this is an array of one
                         # (the current policy)
                         foreach ($policydivisions as $policy) { ?>
-
+                        
                             <?php
+                                $current_policy_agreements = $policyagreements[$policy['policy_id']] ?? [];
```
dracos, line 57: I realise these aren't perfect, but generally I'd try and put the "?? []" bit of this in the business code that fetched the data, so the template can worry less about whether things are defined or not.

dracos (changes requested):
Couple of tiny things. Nice to have a way to include these!
Only minor structural thing, and much easier the way you've done it, is that it feels like you'd want to have a Policy object containing its divisions and its agreements together. But I don't suggest trying to rejig everything to enable that :)

www/includes/easyparliament/templates/html/mp/divisions.php
ajparsons: So retracing my steps, I think I do that in this bit here when we know the policy_id: 

https://github.com/mysociety/theyworkforyou/blob/8d5c0e389604dfed0d8d63c8c044d1b9e64e3101/www/docs/mp/index.php#L443-L455

BUT - there'd then be a problem for the 'all policies' page (which I think is mostly unused but theoretically there) -because not every policy has agreements and so most of that array is unset and so empty. 

So rather than :

```php
$data['policyagreements'] = $policiesList->all_policy_agreements;
```

It would be the following to make sure that policy_ids without agreements get empty lists:

```php
$policy_to_agreements = [];
foreach ($policies->getPolicies() as $policy_id => $policy_text) {
    $policy_to_agreements[$policy_id] = $policiesList->all_policy_agreements[$policy_id] ?? [];
}

$data['policyagreements'] = $policy_to_agreements;
```

And actually that could be put all the way back where `all_policy_agreements` is first defined, and lose a lot of these tests - because they'll all have empty lists to start with. 

I'll test that out later - but just checking the basic idea.

ajparsons:
> Only minor structural thing, and much easier the way you've done it, is that it feels like you'd want to have a Policy object containing its divisions and its agreements together. But I don't suggest trying to rejig everything to enable that :)

Just for revisiting when this isn't fixed in 10 years time - do you mean like a `policy->decisions` that is a list of both?

In twfy-votes, there's a PolicyDecisionLink that can be specialised for Divisions or Agreements - the Policy object maintains two lists, but they can be combined easier.  So could pull that concept back in at some point. 

https://github.com/mysociety/twfy-votes/blob/main/src/twfy_votes/apps/policies/models.py#L163-L167

dracos:
> do you mean like a policy->decisions that is a list of both?

Sorry, no, wasn't clear, I didn't mean they'd be together as one list, but that they'd both be under one object for the policy, rather than be in two different things under Policies

www/includes/easyparliament/templates/html/mp/divisions.php
dracos: OIC, yes, as you're just passing in the whole thing then. It's fine in the template :)

ajparsons:
Fixup commit fixes indenting

dracos (approved):

PR #1763, opened 2024-02-09
Seperate out phpdoc to docker based action

ajparsons:
There's something wrong with the current phpdoc process - I *think* it's a dependency versioning issue that will go away when we upgrade the PHP version to 8 at some point and can use the newer version.

BUT - we can also just move all of this into github actions. This creates a new action that uses the docker version of phpdoc and uploads directly to github pages without pushing back to a branch.

(In principle if we stick with this we could remove the dependencies from the repo, but also it seems fine)


dracos (approved):
Woo, much better, yes. Happy to clear out the dependencies!

ajparsons:
I've removed the dependency and all the tests seem fine - so will merge that as well.

dracos:
You could also have removed build-docs.sh :)

PR #1764, opened 2024-02-09
Convert voting information help page to markdown document

ajparsons:
Add approach for rendering of a markdown document as a help page. 

For the moment, the text is unchanged - the page itself isn't an exact replica as it loses the multiple cards, but that's fine. 

The end goal is making it easier for changes and tweaks to the text , making it easier to add similar pages to the site, and an approach that can be taken over to WriteToThem too.


classes/Renderer/Markdown.php
```diff
@@ -0,0 +1,43 @@
...
+class Markdown
+{
+    public function markdown_document($this_page){
+        // This function takes a markdown file and converts it to HTML
+    
+        $markdown_file = '../../../markdown/' . $this_page . '.md';
+        $Parsedown = new \Parsedown();
+        
+        $text = file_get_contents($markdown_file);
+        $html = $text;
```
dracos, line 20: This line isn't needed

classes/Renderer/Markdown.php
```diff
@@ -0,0 +1,43 @@
...
+        $html = $Parsedown->text($text);
+        
+        # title is the first h1
+        preg_match('/<h1>([^<]+)<\/h1>/i', $html, $matches);
+        
+        $title = $matches[1];
+        
+        $html = preg_replace_callback('/<h([1-3])>([^<]+)<\/h[1-3]>/i', function($matches) {
```
dracos, line 28: The h1 uses js-toc-title, not item, looks like?

classes/Renderer/Markdown.php
```diff
@@ -0,0 +1,43 @@
...
+            $level = $matches[1];
+            $htitle = $matches[2];
+            $slug = slugify($htitle);
+            return "<h$level id=\"$slug\" class=\"js-toc-item\">$htitle</h$level>";
+        }, $html);
+
```
dracos, line 34: If we're making adjustments, seems a shame to lose the `<div class="full-width">` around the images (and also lose their width/height), if that's easy to do, perhaps could have direct HTML for those bits if we cared...

dracos (changes requested):
Few small things

ajparsons:
This should now be almost exactly in sync style wise. Using markdown's HR support to create new panels.

dracos (approved):
Nice, can probably all be squashed

PR #1774, opened 2024-03-05
Voting copy changes

ajparsons:
This supersedes and includes https://github.com/mysociety/theyworkforyou/pull/1756 - rebased over more recent changes.

Basically this is making the "voting summary" page more complete in itself, with better descriptions and better interlinking with the documentation pages. 

## Update voting infomation page

Building on the change to make this a markdown page - new documentation page reviewed by Myf and Julia (although opinions still welcome). 

image

## Always display party comparison information

We always want this information about party to appear - even when there's no actual differences - and link to the info on party comparisons in the help page.

image

## Simplify vote count information

As we now only *score* strong votes, this is now only counting them here. Because it's harder to do this for the numbers we were showing for the party comparison, I've just taken these out as the two should now be in sync anyway (looking at the same set of votes) - so it can be simplified. 

image

## Top level explanation of voting summary

New box to explain the page and link to help pages.

Having little boxes about us in here gives better places for donation asks (which could be bigger than this in future)

image

## Link to feedback form in policy-divisions page

Link to new alchemer form asking for new votes:

image

## Divisions votes side menu doesn't display unnecessary things

Annoyingly this means looping through this twice - but I don't think that's that expensive because it's all already queried..

As some point could just say "this page is only ever for one policy" and get rid of the structure that enables it to do multiple policies. But also, it's messy but fine. 

image
image

# Link to help for informative votes

image


www/includes/easyparliament/templates/html/mp/profile.php
```diff
@@ -69,137 +62,6 @@
                 </div>
               <?php endif; ?>
 
-                <?php if ( !$current_member[HOUSE_TYPE_COMMONS] ) { ?>
```
dracos, line 72: I think the commit message for this commit should explain more that it isn't just moving, it's removing non-MP voting records entirely (as far as I read it?) (Also need a space before my email address)

dracos: Also should some of the 'default' section of www/docs/mp/index.php be removed now if it's no longer needed in this template? (that would impact upon my other comment)

www/docs/mp/index.php
```diff
@@ -383,7 +383,7 @@
 
         $policiesList = new MySociety\TheyWorkForYou\Policies;
         $divisions = new MySociety\TheyWorkForYou\Divisions($MEMBER);
-        $policySummaries = $divisions->getMemberDivisionDetails();
+        $policySummaries = $divisions->getMemberDivisionDetails($strong_only = true);
```
dracos, line 386: There's another use of getMemberDivisionDetails in the default section some lines down, but presumably that's no longer used, so could be removed in previous commit?

www/includes/easyparliament/templates/html/mp/votes.php
```diff
@@ -85,13 +85,17 @@
                         $party_position = $key_vote['party_position'] ;
                         $comparison_party = $data["comparison_party"];
                         $current_party_comparison = $data["current_party_comparison"];
-                        $party_voting_line = sprintf("%s, %s", $party, $diff['party_voting_summary']);
+                        $unslugified_comparison_party = ucwords(str_replace('-', ' ', $comparison_party));
```
dracos, line 106: Feels like we have a lookup somewhere for this, but can't find it.

www/includes/easyparliament/templates/html/mp/votes.php
```diff
@@ -174,7 +174,7 @@
                                     $party_position = $diff['party_position'];
                                     $party_score_difference = $diff["score_difference"];
                                     if ($sorted_diffs[$policy_id]['score_difference'] > $min_diff_score && $party_member_count > 1) {
-                                        $party_voting_line = sprintf( 'Comparable %s MPs %s (%s).', $comparison_party, $diff['party_position'], $diff['party_voting_summary']);
+                                        $party_voting_line = sprintf( 'Comparable %s MPs %s.', $unslugified_comparison_party, $diff['party_position']);
```
dracos, line 197: Afraid I don't really understand this commit message

dracos (changes requested):
A few tiny questions is all, I think

www/includes/easyparliament/templates/html/mp/profile.php
ajparsons: Ah yes - I was assuming we didn't have them but we do for the House of Lords sometimes or for ex-MPs who become something else. 

Will investigate the default seciton.

www/includes/easyparliament/templates/html/mp/votes.php
ajparsons: It's in principle in people.json but I couldn't see it was made available anywhere. Nothing says this approach *has* to work, but does seem to for parties we care about.

www/docs/mp/index.php
```diff
@@ -1162,11 +1156,11 @@ function policy_image($data, $MEMBER, $format) {
 }
 
 // generate party policy diffs
-function person_party_policy_diffs($MEMBER, $policiesList, $only_diffs) {
+function person_party_policy_diffs($MEMBER, $policiesList) {
     global $data;
 
     $divisions = new MySociety\TheyWorkForYou\Divisions($MEMBER);
-    $policySummaries = $divisions->getMemberDivisionDetails();
+    $policySummaries = $divisions->getMemberDivisionDetails($strong_only = true);
```
dracos, line 1163: Sorry, didn't notice this before - this should just be `(true)` (and the other same line), you don't want to set a variable here while calling the function.

dracos (changes requested):
Thanks (fixup commits generally easier to re-review :) ), only the one thing I noticed.

www/docs/mp/index.php
ajparsons: Doh - fixed in fixup.

dracos (approved):
Woo

PR #1776, opened 2024-03-12
Fix loose name appearing on voting records

ajparsons:
If rebellion rate isn't loaded you just get this:

image


dracos (approved):
`it's` to `its` in the commit message :)

PR #1781, opened 2024-03-22
Python division tweaks

dracos:


ajparsons (changes requested):
I was avoiding the commonlib mainly because the sys.path.append approach worked badly with my IDE. But! There's a solution to that.

Two fixup commits, and this is good if you're happy with them:

* The commonlib config module needs to use get and strips off the OPTION_ parts. 
* Update commonlib submodule to one with a pyproject.toml - which means it can then be added as a local package in requirements.txt and accessed as such in the code.

dracos:
Oh yes, I saw the `get` and then forgot to do it, ta. You haven't updated commonlib, it's a submodule, so you'll need to make whatever change you're describing there and then update where the submodule points to here.

dracos:
Oh yes, you did, ignore me!

PR #1782, opened 2024-03-25
Voting copy tweaks

ajparsons:
Series of small fixups and tidy-ups - I don't think anything controversial.

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -299,6 +299,24 @@ class Divisions {
           $policy_divisions[$policy_id] = $summary;
         }
 
+        // for each key in $policy_divisions, we want to add agreement information
+    
+        $policies_list = new \MySociety\TheyWorkForYou\Policies();
+        foreach ($policy_divisions as $policy_id => &$summary) {
+            $agreement_details = $this->member->member_agreements($policy_id, HOUSE_TYPE_COMMONS, $policies_list );
+            $summary["agreements_for"] = 0;
+            $summary["agreements_against"] = 0;
+            foreach ($agreement_details as $agreement){
+                if ($strong_only == true & $agreement["strength"] != "strong") {
+                    continue;
+                }
+                if ($agreement["alignment"] == "agree") {
+                    $summary["agreements_for"] += 1;
+                } else {
+                    $summary["agreements_against"] += 1;
+                }
+            }
+        }
         return $policy_divisions;
     }
 
@@ -467,6 +485,14 @@ class Divisions {
             $votes['for'] . ' ' . make_plural('vote', $votes['for']) . ' for',
             $votes['against'] . ' ' . make_plural('vote', $votes['against']) . ' against'
         );
+
+        if ( $votes['agreements_for']) {
+            $actions[] = $votes['agreements_for'] . ' ' . make_plural('agreement', $votes['agreements_for']) . ' for';
+        }
+
+        if ( $votes['agreements_against']) {
+            $actions[] = $votes['agreements_against'] . ' ' . make_plural('agreement', $votes['agreements_against']) . ' against';
+        }
 
         if ( $votes['both'] ) {
             $actions[] = $votes['both'] . ' ' . make_plural('abstention', $votes['both']);
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -375,6 +375,10 @@ $data['current_assembly'] = $country[2];
 $data['policy_last_update'] = MySociety\TheyWorkForYou\Divisions::getMostRecentDivisionDate();
 
 $data['comparison_party'] = $MEMBER->cohortParty();
+$data['unslugified_comparison_party'] = ucwords(str_replace('-', ' ', $data['comparison_party']));
+
+// is the party we're comparing this MP to different from the party they're currently in?
+$data['party_switcher'] = (slugify($data['current_party_comparison']) != slugify($data["comparison_party"]));
 
 // Do any necessary extra work based on the page type, and send for rendering.
 switch ($pagetype) {
@@ -447,13 +451,7 @@ switch ($pagetype) {
 
         if ( $policyID ) {
             $data['policydivisions'] = $divisions->getMemberDivisionsForPolicy($policyID);
-            $rel_agreements = $policiesList->all_policy_agreements[$policyID] ?? [];
-            // filter down to where 'date' is within the member's term
-            // This won't be perfect where the member has been in and out of the house
-            // But it doesn't affect their score.
-            $rel_agreements = array_filter($rel_agreements, function($agreement) use ($MEMBER) {
-                return $MEMBER->date_in_memberships(HOUSE_TYPE_COMMONS, $agreement['date']);
-            });
+            $rel_agreements = $MEMBER->member_agreements($policyID, HOUSE_TYPE_COMMONS);
             $data['policyagreements'] = array($policyID => $rel_agreements);
         } else {
             $data['policydivisions'] = $divisions->getAllMemberDivisionsByPolicy();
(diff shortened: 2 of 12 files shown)


www/includes/easyparliament/templates/html/mp/votes.php
```diff
@@ -77,16 +77,29 @@
                 <div class="panel">
                     <a name="votes"></a>
                     <h2><?= $full_name ?>&rsquo;s voting in Parliament</h2>
-
-                    <p>
-                    <?= $full_name ?> is a <?= $party ?> MP, and on the <b>vast majority</b> of issues <a href="/voting-information/#party-and-individual-responsibility-for-decisions">follow instructions from their party</a> and vote the <b>same way</b> as other <?= $party ?> MPs.
-                    </p>
+                    <?php $unslugified_comparison_party = ucwords(str_replace('-', ' ', $comparison_party)); ?>
```
dracos, line 80: you don't need this, if you've now set it in mp/index.php?

www/docs/mp/index.php
```diff
@@ -375,6 +375,10 @@
 $data['policy_last_update'] = MySociety\TheyWorkForYou\Divisions::getMostRecentDivisionDate();
 
 $data['comparison_party'] = $MEMBER->cohortParty();
+$data['unslugified_comparison_party'] = ucwords(str_replace('-', ' ', $data['comparison_party']));
+
+// is the party we're comparing this MP to different from the party they're currently in?
+$data['party_switcher'] = (slugify($data['current_party_comparison']) <> slugify($data["comparison_party"]));
```
dracos, line 381: Think `!=` is more usual.

classes/Divisions.php
```diff
@@ -257,10 +257,16 @@ public function getMemberDivisionDetails($strong_only = false) {
             GROUP BY policy_id, policy_vote, vote",
             $args
         );
+        
+        $allowed_policies = (new \MySociety\TheyWorkForYou\Policies())->getPolicyIDs();
 
         foreach ($q as $row) {
           $policy_id = $row['policy_id'];
-
+          // deleted policies may still be in the database - ignore them
```
dracos, line 265: This feels like it's a bit more than the commit message implies - mention this in the commit?

classes/Divisions.php
```diff
@@ -299,6 +305,23 @@ public function getMemberDivisionDetails($strong_only = false) {
...
+                if ($strong_only == true & $agreement["strength"] != "strong") {
+                    continue;
+                }
+                if ($agreement["alignment"] == "agree") {
+                    $summary["agreements_for"] += 1;
+                } else {
+                    $summary["agreements_against"] += 1;
+                }
+            }
+            $policy_divisions[$policy_id] = $summary;
```
dracos, line 323: I don't think this line is needed, because you're altering the $summary object directly above? Might need checking.

www/docs/mp/index.php
```diff
@@ -452,12 +452,7 @@ function($k) { return $k['score_difference'] >= 2; }
         if ( $policyID ) {
             $data['policydivisions'] = $divisions->getMemberDivisionsForPolicy($policyID);
             $rel_agreements = $policiesList->all_policy_agreements[$policyID] ?? [];
```
dracos, line 454: This line should be removed.

www/includes/easyparliament/member.php
```diff
@@ -242,6 +242,17 @@ public function date_in_memberships($house, $date){
         return false;
     }
 
+    public function member_agreements($policyID, $house = HOUSE_TYPE_COMMONS){
```
dracos, line 245: Could we pass in the Policies object rather than only a policy ID? Only each time it's instantiated it has to load in the big JSON file, so eg the call from Divisions.php is doing that over and over once per policy ID, but you've already created such an object on new line 261 (and the other call, you already have $policiesList).

www/includes/easyparliament/member.php
```diff
@@ -242,6 +242,17 @@ public function date_in_memberships($house, $date){
...
+        // agreements that a member has been present for on a specific policy
+        $policiesList = new MySociety\TheyWorkForYou\Policies($policyID);
+        $rel_agreements = $policiesList->all_policy_agreements[$policyID] ?? [];
+        $MEMBER = $this;
```
dracos, line 249: More out of interest, is this needed, does `use ($this...` not work okay?

dracos (changes requested):
Some small things

www/includes/easyparliament/member.php
ajparsons: I got an error for it - this looks like the documentation for that. Nicer approaches welcome! https://www.php.net/manual/en/functions.anonymous.php#:~:text=Any%20such%20variables%20must%20be%20passed%20to%20the%20use%20language%20construct.%20As%20of%20PHP%207.1%2C%20these%20variables%20must%20not%20include%20superglobals%2C%20%24this%2C%20or%20variables%20with%20the%20same%20name%20as%20a%20parameter

dracos: I think $this is automatically available through the closure without `use`, e.g.
`php -r'class Foo { public $allow = 2; function foo() { $a = [1,2,3,4]; $a = array_filter($a, function($n) { return $n === $this->allow; }); return $a; } }; print_r( (new Foo)->foo() );'`
Outputs 2.

So hopefully:
```
        $rel_agreements = array_filter($rel_agreements, function($agreement) use ($house) {
            return $this->date_in_memberships($house, $agreement['date']);
        });
```
would work.

www/includes/easyparliament/member.php
ajparsons: Thinking about this - is the core problem that it's become expensive to create that object? Would putting json-based data in a 5 minute memcache sidestep this? (appreciate that's adding complexity to an already unideal approach).

Otherwise yes, this method could accept an optional policies object and only create a new one if it didn't get it.

dracos: I don't know if it's actually expensive, it just felt wrong to repeatedly open a file. And sticking in a cache also feels wrong for the same reason :) Feels like you'd only ever need to instantiate the policies once each request, but meh :) opitonal policies object sounds good.

classes/Divisions.php
ajparsons: Yes, if $summary is adjusted to being passed by ref.

www/includes/easyparliament/templates/html/mp/votes.php
ajparsons: Yep.

www/docs/mp/index.php
ajparsons: Yep.

www/includes/easyparliament/member.php
ajparsons: Yep, that works as expected!

www/includes/easyparliament/member.php
ajparsons: Null approach added!

www/docs/mp/index.php
```diff
@@ -447,13 +451,7 @@ function($k) { return $k['score_difference'] >= 2; }
...
         if ( $policyID ) {
             $data['policydivisions'] = $divisions->getMemberDivisionsForPolicy($policyID);
-            $rel_agreements = $policiesList->all_policy_agreements[$policyID] ?? [];
-            // filter down to where 'date' is within the member's term
-            // This won't be perfect where the member has been in and out of the house
-            // But it doesn't affect their score.
-            $rel_agreements = array_filter($rel_agreements, function($agreement) use ($MEMBER) {
-                return $MEMBER->date_in_memberships(HOUSE_TYPE_COMMONS, $agreement['date']);
-            });
+            $rel_agreements = $MEMBER->member_agreements($policyID, HOUSE_TYPE_COMMONS);
```
dracos, line 454: Could this pass in `$policiesList` as well?

dracos (commented):
Thanks, all the changes look good :) One possible change to the other member_agreements call is all.

ajparsons:
Ok, I've addressed original issues - and then decided to just do some tidying up in the Policies class while it was open - which then needed to remove 7.3 from testing. I will stop expanding the scope of this PR...

dracos (approved):

PR #1797, opened 2024-06-02
Separate register page

ajparsons:
Move register of interests to separate sub profile page.

- Clearer top navigation from overview page (We should do the same to debates at some point)
- Track interest in this feature
- More room for text about whofundsthem. 

Room for improving this with new boilerplate text - but this does a few things quickly. 

As we want to do this with recent debates/future features(!?) i've moved the list of allowed sub-profile pages from httpd.conf into the profile php page to have more of the logic in one place. But we add new features rarely enough if the other approach is better for some reason can switch back. 

image

### DIFF
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -475,6 +475,10 @@ switch ($pagetype) {
 
         break;
 
+    case 'register':
+        // Send the output for rendering
+        MySociety\TheyWorkForYou\Renderer::output('mp/register', $data);
+
     case 'policy_set_svg':
         policy_image($data, $MEMBER, 'svg');
         break;
--- a/www/docs/mp/index.php
+++ b/www/docs/mp/index.php
@@ -40,9 +40,14 @@ include_once '../api/api_getGeometry.php';
 include_once '../api/api_getConstituencies.php';
 
 // Ensure that page type is set
+$allowed_page_types = ['divisions', 'votes', 'policy_set_svg', 'policy_set_png', 'recent', 'register'];
+
 if (get_http_var('pagetype')) {
     $pagetype = get_http_var('pagetype');
 } else {
+    $pagetype = 'profile';
+}
+if (!in_array($pagetype, $allowed_page_types)) {
     $pagetype = 'profile';
 }
 if ($pagetype == 'profile') {
--- a/www/includes/easyparliament/templates/html/mp/register.php
+++ b/www/includes/easyparliament/templates/html/mp/register.php
@@ -1,0 +1,78 @@
+<?php
+include_once INCLUDESPATH . "easyparliament/templates/html/mp/header.php";
+?>
+
+<div class="full-page">
+    <div class="full-page__row">
+        <div class="full-page__unit">
+            <?php include '_person_navigation.php'; ?>
+        </div>
+        <div class="person-panels">
+            <div class="sidebar__unit in-page-nav">
+                <div>
+                    <h3 class="browse-content"><?= gettext('Browse content') ?></h3>
+                    <ul>
+                      <li><a href="https://www.mysociety.org/2024/01/17/improving-the-register-of-mps-interests/">Download a spreadsheet</a></li>
+                      <?php include '_featured_content.php'; ?>
+                      <?php include '_donation.php'; ?>
+                </div>
+            </div>
+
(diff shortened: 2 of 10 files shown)


www/includes/easyparliament/templates/html/mp/register.php
```diff
@@ -0,0 +1,94 @@
...
+                    <?= $register_interests['data'] ?>
+
+
+                    <p>
+                         <a class="moreinfo-link" href="https://www.publications.parliament.uk/pa/cm/cmregmem/100927/introduction.htm">More about the register</a>
+                    </p>
+                </div>
+                <?php endif; ?>
+
+                <div class="panel panel--secondary">
```
dracos, line 61: There already is a _vote_footer.php that seems similar (though not identical) to this - could we factor all the MP pages to use it?

www/includes/easyparliament/templates/html/mp/register.php
```diff
@@ -0,0 +1,94 @@
...
+                    <ul>
+                      <li><a href="https://www.mysociety.org/2024/01/17/improving-the-register-of-mps-interests/">Download a spreadsheet</a></li>
+                      <?php include '_featured_content.php'; ?>
+                      <?php include '_donation.php'; ?>
+                </div>
+            </div>
+
+            <div class="primary-content__unit">
+
+              <?php if ($party == 'Sinn Féin' && in_array(HOUSE_TYPE_COMMONS, $houses)): ?>
```
dracos, line 23: Presumably this block isn't needed, because there won't be a register for sinn fein or newly elected MPs.

www/includes/easyparliament/templates/html/mp/register.php
```diff
@@ -0,0 +1,94 @@
...
+                    </p>
+                    <?php if ($register_interests['date']): ?>
+                        <p>Last updated: <?= $register_interests['date'] ?>.</p>
+                    <?php endif; ?>
+
+                    <?= $register_interests['data'] ?>
+
+
+                    <p>
+                         <a class="moreinfo-link" href="https://www.publications.parliament.uk/pa/cm/cmregmem/100927/introduction.htm">More about the register</a>
```
dracos, line 56: Shall we update this to a more up-to-date page?

dracos (changes requested):
A few suggestions and tidyings :)

www/docs/mp/index.php
```diff
@@ -40,11 +40,16 @@
...
 
 // Ensure that page type is set
+$allowed_page_types = ['divisions', 'votes', 'policy_set_svg', 'policy_set_png', 'recent', 'register'];
+
 if (get_http_var('pagetype')) {
     $pagetype = get_http_var('pagetype');
 } else {
     $pagetype = 'profile';
 }
+if (!in_array($pagetype, $allowed_page_types)) {
```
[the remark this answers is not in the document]
ajparsons: So the issue here is without this check, the full version of the URL with the constituency in it gets the constituency assigned to the $pagetype - which might be an argument just for using the httpd.conf to manage this? Your call.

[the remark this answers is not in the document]
dracos: Ah, right - hmm, I guess the get_person_by_id check redirects if it doesn't match a known thing anyway, so actually this works okay (just with a redirect, rather than a 404 - I just didn't want /mp/N/name/const/random to display a page, which it presuambly still doesn't).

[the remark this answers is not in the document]
ajparsons: Yes - that's right, it redirects back to the main profile page if you add things after the constituency not in this list.

www/includes/easyparliament/templates/html/mp/register.php
ajparsons: Added some of the missing bits in extra commit and moved it all to a common `_profile_footer.php`.

www/includes/easyparliament/templates/html/mp/register.php
ajparsons: Yep, out.

www/includes/easyparliament/templates/html/mp/register.php
ajparsons: Linked to main parliament register page - in general we'll want to come back and add better text.

dracos (approved):

PR #1798, opened 2024-06-09
Improve codespace secrets approach

ajparsons:
Features like the donate page and election lookup need external API keys, which doesn't work nicely with the docker setup because the conf/general is recreated each time the docker is opened (rather than created). 

This PR:

- Adjusts docker setup so that system environmental variables populate the `conf/general` file.
- Adjust 'devcontainer.json' to prompt codespaces users to setup the (optional) keys in their user codespace secrets store.

image


docker-compose.yml
```diff
@@ -24,6 +24,13 @@ services:
       TWFY_TEST_DB_USER: twfy
       TWFY_TEST_DB_PASS: password
       TWFY_VOTES_URL: ${TWFY_VOTES_URL}
+      MAPIT_URL: ${MAPIT_URL:-https://mapit.mysociety.org/}
+      MAPIT_API_KEY: ${MAPIT_API_KEY:-}
+      DEMOCRACYCLUB_TOKEN: ${DEMOCRACYCLUB_TOKEN:-}
+      RECAPACHA_SITE_KEY: ${RECAPACHA_SITE_KEY:-}
```
dracos, line 30: Some spelling mistakes in here

dracos (changes requested):
Some typos to fix I think is all

docker-compose.yml
ajparsons: Should be fixed!

dracos: And still mispelled here

ajparsons:
Sorry sorry - should all be fixed.

dracos (approved):
Thanks. Didn't want that to cause some confusion down the line somewhere, if unlikely :)

PR #1799, opened 2024-06-10
Election copy changes

ajparsons:
Editing pass updating the copy on postcode and MP page linking to more resources.

Trying putting the election box beneath the former MP box to make the navigation a bit clearer at the top. I think the election box is still fairly clear (especially as now a bit bigger with more buttons). 

image

image

### DIFF
--- a/www/includes/easyparliament/templates/html/postcode/index.php
+++ b/www/includes/easyparliament/templates/html/postcode/index.php
@@ -134,20 +134,12 @@ include("repexplain.php");
                   <?php } ?>
                 </ul>
 
-<h3><?= gettext('Donate') ?></h3>
-<p>
-For the election mySociety are busy helping people to understand what their
-election candidates stand for with <a href="https://www.theyworkforyou.com/">TheyWorkForYou</a>; supporting people to have
-informed conversations with candidates around climate using <a href="https://www.localintelligencehub.com/">Local Intelligence
-Hub</a>; and ensuring all our democracy services are ready to go on day one of a
-new government.
+<h3>We can make politics better together</h3>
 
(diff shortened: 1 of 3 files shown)


www/includes/easyparliament/templates/html/postcode/index.php
```diff
@@ -134,20 +134,12 @@ function member_image_box(string $person_id, string $person_url, string $person_
...
-
-<p><a href="/support-us/?utm_source=theyworkforyou.com&amp;utm_content=postcode+donate&amp;utm_medium=link&amp;utm_campaign=postcode" class="button">Donate</a>
-
-<p>Whoever is elected, they need to understand the importance of
-transparency and accountability — and we’ll be making sure that happens.
-
+<h3>We can make politics better together</h3>
 
+<p>We want MPs to meet the standards and expectations of the people who elected them - <strong>you</strong>!</p>
+<p>Learn about <a href="/support-us/?utm_source=theyworkforyou.com&utm_content=postcode+donate&utm_medium=link&utm_campaign=postcode&how-much=5">our current work</a>, and <a href="https://www.mysociety.org/democracy/who-funds-them/">our new project WhoFundsThem</a> - looking into MPs' and APPGs' financial interests.</p>
```
dracos, line 140: Can we have curly apostrophes please :)

dracos (changes requested):
Sorry, one pretty and prettifying pedantic change is all!

www/includes/easyparliament/templates/html/postcode/index.php
ajparsons: Done! Will wait for tests, and then squash the PR.

PR #1800, opened 2024-06-11
Upgrade Stripe library/API version to current

dracos:
Similar (hopefully identical?) to https://github.com/mysociety/mapit.mysociety.org/pull/158
so we can get on current Stripe API version, and improve downgrade behaviour.

### DIFF
--- a/classes/Subscription.php
+++ b/classes/Subscription.php
@@ -64,7 +64,7 @@ class Subscription {
                     'latest_invoice.payment_intent',
                 ],
             ]);
-        } catch (\Stripe\Error\InvalidRequest $e) {
+        } catch (\Stripe\Exception\InvalidRequestException $e) {
             $this->db->query('DELETE FROM api_subscription WHERE stripe_id = :stripe_id', [':stripe_id' => $id]);
             $this->delete_from_redis();
             return;
@@ -90,7 +90,7 @@ class Subscription {
 
         try {
             $this->upcoming = $this->api->getUpcomingInvoice(["customer" => $this->stripe->customer->id]);
-        } catch (\Stripe\Error\Base $e) {
+        } catch (\Stripe\Exception\ApiErrorException $e) {
         }
     }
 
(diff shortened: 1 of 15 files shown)


classes/Subscription.php
```diff
@@ -98,33 +100,74 @@ private function update_subscription($form_data) {
...
+            $args = [
+                'payment_behavior' => 'allow_incomplete',
+                'plan' => $form_data['plan'],
+                'metadata' => $form_data['metadata'],
+                'cancel_at_period_end' => false, # Needed in Stripe 2018-02-28
+                'proration_behavior': 'always_invoice',
+            ];
+            if ($form_data['coupon']) {
+                $args['coupon'] = $form_data['coupon'];
+            } elseif ($this->stripe->discount) {
```
struan, line 164: I'm assuming this is just saying if there was a discount and there is no longer a discount then clear the discount?

struan (approved):
one tiny question but assuming that's fine it looks good.

classes/Subscription.php
dracos: Yeah, exactly, I think update only changes the things you provide.

ajparsons:
https://github.com/mysociety/theyworkforyou/tree/stripe-upgrade-format < should be this but via the new formatting.

dracos:
The new branch lost a couple of things:
```
git diff 714afc790..6e894f05e > ORIG
git diff a4cbec8ff..387bf32eb > NEW
diff ORIG NEW|grep -v @@|grep -v '\---'|grep -v '^. index'|grep -v '^[0-9]*c'
< -                'tax_percent' => 20
> -                'tax_percent' => 20,
371,374c371,374
<      <?php if ($stripe) { echo 'data-has-subscription="1"'; } ?>
<      src="<?= cache_version('js/payment.js') ?>"></script>
<
>      <?php if ($stripe) {
>          echo 'data-has-subscription="1"';
>      } ?>
436,439c436,439
<  <script id="js-payment"
<      <?php if ($subscription->has_payment_data) { echo 'data-has-payment-data="1"'; } ?>
<      <?php if ($stripe) { echo 'data-has-subscription="1"'; } ?>
>      <?php if ($stripe) {
>          echo 'data-has-subscription="1"';
>      } ?>
```

dracos:
Thanks, my ignorance on previous comment!

dracos:
```diff
-$subscription->stripe->cancel(['at_period_end' => true]);
+\Stripe\Subscription::update($subscription->stripe->id, ['cancel_at_period_end' => false]);
```
true somehow became false here!

PR #1803, opened 2024-07-01
Use constituency data from main people.json file.

dracos:
Due to changes in https://github.com/mysociety/parlparse/pull/173


ajparsons (approved):
Makes sense to me!

dracos:
Have added a commit to speed up running when we can run it as `load-people --from 2024-07-04`

PR #1804, opened 2024-07-02
Update LIH link format for new cons

ajparsons:
For new memberships, use WMC23: e.g. https://www.localintelligencehub.com/area/WMC23/Carshalton%20and%20Wallington


struan (approved):
👍

PR #1805, opened 2024-07-03
Add WhereDoIVote link

ajparsons:
Very basic link to polling station: uses the UPRN if we have it, postcode if we don't.

image


dracos (commented):
Given the parent commit, please don't use $_GET in any code :) Better to set the variables in the 'controller' and use in template directly (e.g. $pc will already and always be set here).

Issue #1791, opened 2024-04-23
Create mockup for Homepage


lucascumsille:
With heavy text section we should avoid using green as background.

lucascumsille:
Close with https://github.com/mysociety/theyworkforyou/pull/1809

Issue #1789, opened 2024-04-10
Button contrast colours are not accessible


lucascumsille:
I know this is part of a wider discussion about our colours across all our websites and applies not only to the buttons but other sections as well. But I'm raising this issue because I think we could add an easy provisory fix to improve the experience.

![Image](https://github.com/mysociety/theyworkforyou/assets/13790153/5b4e8c11-0e5a-4ba0-b77f-6fcddb22119d)

While we figure a permanent solution, we could add the following fixes:

- [ ]  green button use `colour: #333333`
- [ ] `.btn--negative` and  `.donate-button` to use a the same `background-colour `. Not sure if they use different colours on purpose to make the user aware they have different intentions, but the difference is quite subtle. 
- [ ] For the `.btn--negative` and `.donate-button` use a slightly darker version for the current red we are using, we can keep the white `colour` and pass the contrast test.

This is how it would look:

![Image](https://github.com/mysociety/theyworkforyou/assets/13790153/d4d35d9f-2d5d-4fe1-ab34-32b9eac1ff63)

@ajparsons let me know if this sounds ok with you and I'll go ahead.

ajparsons:
Cool, these are fine except for the search button.

If you scroll down on the homepage you get the big search box - here black text works fine because it's black > green > black. 

But at the top, white > green > black seem very low contrast. 

Why can't we just defer to the bootstrap colours here? https://getbootstrap.com/docs/5.0/components/buttons/ That seems to be doing the same job ok. 

In general I'd be really good to some proper time thinking about the general picture - I'm still not wild about the black on green in the big text on the homepage (https://github.com/mysociety/theyworkforyou/pull/1699) but it was the quicker fix at the time.

lucascumsille:
Closed with https://github.com/mysociety/theyworkforyou/pull/1809

PR #1809, opened 2024-07-05
Home page box

ajparsons:
related to https://github.com/mysociety/theyworkforyou/issues/1789 and  https://github.com/mysociety/theyworkforyou/issues/1791

Quick proof of concept of making the search box the main intro box:

image

Comments from Matthew on slack about switching popular searches and quick links - exploring some version of a grid here.


ajparsons:
I think this is hitting the point where it needs a designer to have a think - @lucascumsille ?

lucascumsille:
TODOS:

- [ ] Get rid of comments and placeholder text
- [ ] Responsive for smaller sizes screen

lucascumsille:
@ajparsons I have included the new layout for the "Quick links" sections. Here is a video of how it would look on mobile and desktop:

https://github.com/user-attachments/assets/5aa3a944-2a15-4724-90d8-bdee86c418d6

I notched down the size and fonts of cards so they don't look so prominent, but they are still noticeable.

Apologies for the amount of commits, I would normally rebase, but trying to run `git rebase -I master` causes the following issue:
```
hint: Waiting for your editor to close the file... error: cannot run editor: No such file or directory
error: unable to start editor 'editor'
```
I think is the same problem we had in the past, let me know if you know how to fix it, and I'll rebase =)

Regarding the images for the other parliaments, due to timing, I didn't have much time to keep looking, but during the week I might have some extra time. In the meantime let me know what you think of the current layout and if you need me to make some changes.

ajparsons:
Ok, I've rebased that and fixed the tests. 

On images: if you could give me an example of how you'd make the UK image responsive - I can apply that to the others.

I'm a bit stuck on the images and putting it down for the day. 

NI Assembly is fine - it's a nice horizontal building that bits behind the banner. 

The trouble with the other two:

Scottish Parliament: to get clean sky behind it, you're usually looking at quite a *busy* bit of the building architecturally - which makes it hard to layer the search box on top of it. 

Like this is fine: 

image

But is less *nice* than the others. Don't have to get it right first time. 

Senedd I've tried a few, but the current one is the best. Again a problem where the architecturally distinctive bit is hard to layer a box onto of. 

image

lucascumsille:
@ajparsons I added the CSS rules + Images for the UK Parliament.

A few considerations:

- I decreased the quality while doing the compression of the images to 92% quality.

- For mobile and tablet I used images sizing from their minimum viewpoint. For example, for Tablet the image has a width of 768px because that's where medium starts. While for Desktop and Huge(1136px) I didn't create images using the minimum, I added some extra pixels. The reason behind, it's that on Desktop and Huge screens users can appreciate the image more, because there is more space. So it makes sense the quality is better for those ones.

I have added a rule for `huge-screen`, but considering that rule applies for screen above > 1136px, while $large from 1040px, probably is not very useful at the moment, unless we create a new variable for truly 'huge screens'

Let me know what you think

ajparsons:
Ok, I this now does everything I think is needed - thanks for the work on those devolved parliament images Lucas.

@dracos : Is there anything in this you want to look at in more detail? 

Lucas was a bit concerned that the devolved parliament images were a bit heavy even compressed. They're off the main homepage so I think that's more ok, but flagging.

classes/SectionView/SeneddView.php
```diff
@@ -98,11 +98,43 @@ protected function display_front_senedd() {
...
+        $search_box->search_section = "senedd";
+        $search_box->quick_links = [];
+        if (count($data["regional"])) {
+            // get all unique constituencies
+            $constituencies = array();
+            foreach ($data["regional"] as $member) {
+                $constituencies[$member["constituency"]] = 1;
+            }
+            $constituencies = array_keys($constituencies);
+            if (LANGUAGE == 'cy'){
```
dracos, line 125: Feels like this could be done with a string and an actual translation call to keep the translations in the translation file, but no biggie.

classes/SectionView/SeneddView.php
```diff
@@ -98,11 +98,43 @@ protected function display_front_senedd() {
...
+        if (LANGUAGE == 'cy') {
+            $search_box->homepage_subhead = "Senedd";
+        } else {
+            $search_box->homepage_subhead = "Senedd / Welsh Parliament";
+        }
+        $search_box->homepage_desc = "";
+        $search_box->search_section = "senedd";
+        $search_box->quick_links = [];
+        if (count($data["regional"])) {
+            // get all unique constituencies
```
dracos, line 119: Done different from Scotland for some reason?

dracos (changes requested):
Couple of tiny questions.
On the images - firstly, we need to take out the photos we're not using that are stock? And secondly, running e.g. the Scottish huge one through squoosh.app gets it from 85.2KB to 40.4KB kept as a webp or 59.4KB JPEG, so would like them to at least go through that. Or Senedd huge goes from 274KB to 147 or 135KB.

classes/SectionView/SeneddView.php
```diff
@@ -122,11 +118,7 @@ protected function getSearchBox(array $data): \MySociety\TheyWorkForYou\Search\S
                 $constituencies[$member["constituency"]] = 1;
             }
             $constituencies = array_keys($constituencies);
-            if (LANGUAGE == 'cy'){
-                $search_box->add_quick_link('Darganfod mwy am eich Aelodau Cynulliad dros ' . $constituencies[0] . ' a ' . $constituencies[1], '/postcode/?pc=' . $data["mp_data"]['postcode'],'torso');
-            } else {
-                $search_box->add_quick_link('Find out more about your MSs for ' . $constituencies[0] . ' and ' . $constituencies[1], '/postcode/?pc=' . $data["mp_data"]['postcode'],'torso');
-            }
+        $search_box->add_quick_link(gettext('Find out more about your MSs for ') . $constituencies[0] . gettext(' and ') . $constituencies[1], '/postcode/?pc=' . $data["mp_data"]['postcode'],'torso');
```
dracos, line 121: Sorry, I realise I asked for this, but good knowledge to share - unless you know the translated language has the same construction as English (I don't know if Welsh does here), this should be `sprintf(gettext("Find out more about your MSs for %s and %s"), $constituencies[0], $constituencies[1])` so that the sentence can reorder if the language requires that.

dracos:
Also, just one thought - check the background colour is okay if the images don't load, I see WEBP is pretty well supported nowadays but just to be nice to anyone that doesn't get it.

ajparsons:
> Also, just one thought - check the background colour is okay if the images don't load, I see WEBP is pretty well supported nowadays but just to be nice to anyone that doesn't get it.

Quick check, it's not the prettiest but is all legible.

image

classes/SectionView/SeneddView.php
ajparsons: Yes that seemed the wrong way (fine for welsh, but good to learn!)

ajparsons:
Ok, I need to merge that last set of fixups - but I've:

* Squished the images more
* Replaced the Scottish Parliament mobile image
* Did translation change in above comment (seems to work locally)

dracos (approved):
All looks good - only thing is why is parliament-photo.jpg in there, it doesn't appear to be used, but there may be a reason for that

PR #1812, opened 2024-07-23
Update queries to better match pw export

ajparsons:
Adjusts the export for twfy-votes to more closely mirror the original public whip data table dumps.

### DIFF
--- a/scripts/division_io.py
+++ b/scripts/division_io.py
@@ -273,6 +273,16 @@ def export_division_data(verbose: bool = False):
         ON persondivisionvotes.division_id = divisions.division_id
     JOIN member
         ON persondivisionvotes.person_id = member.person_id
+        -- division.house is a string, so we need to convert it to correct integer
+        AND CASE 
+                WHEN divisions.house = 'pbc' THEN 1
+                WHEN divisions.house = 'commons' THEN 1
+                WHEN divisions.house = 'lords' THEN 2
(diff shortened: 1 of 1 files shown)


scripts/division_io.py
```diff
@@ -223,13 +223,66 @@ def export_division_data(verbose: bool = False):
...
+        persondivisionvotes.division_id as division_id,
+        vote,
+        proxy
+    FROM
+        persondivisionvotes
+    JOIN divisions
+        ON persondivisionvotes.division_id = divisions.division_id
+    JOIN member
+        ON persondivisionvotes.person_id = member.person_id
+        AND divisions.division_date BETWEEN member.entered_house AND member.left_house;
```
dracos, line 276: This will have an issue if someone is in two houses at once, that both have divisions, won't it?

dracos (commented):
one query is all

scripts/division_io.py
ajparsons: Ah right - it needs to also have a `divisions.house = member.house` I think? I'll give that a go at some point and check it outputs correctly.

ajparsons: Ok, fix in a new commit which I'll squish. divisions.house is string and member.house is int so it needs the conversion in the query.  Adding this slightly decreased the number of divisions being exported as you'd hope.

dracos (approved):
Think this is okay

PR #1814, opened 2024-08-23
PHP formatter

ajparsons:
Adds [php-cs-fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer) as a formatter for php files. 

- Reasonably quick - stores a cache after first run so not horrible when just editing a few files. 
- Has rules by php version - so can initially get us consistent on `array()` vs `[]` - and we can adjust this as we go into php 8. 

I've separated the two rules into two commits to show what it's doing at each step. This includes removing older style syntax.

When approved, the last two commits will be undone and rerun on whatever the state of master is at that point. Ideally this is when there are few PHP touching PRs outstanding.


ajparsons:
Now has a rule to put braces on the same line for classes and functions.

dracos (approved):

PR #1815, opened 2024-08-23
Python add poetry

ajparsons:
Stage one of https://github.com/mysociety/theyworkforyou/pull/1813

- requirements.txt replaced with Poetry (will need feature enabled before deploy to server)
- Creates a src/twfy_tools that acts as the common library for features.
- Config for ruff as a formatter/linter.
- Wrapper for config (lets IDE see values and type)
- Common DB access function.


ajparsons:
Before deploy need to add poetry to the relevant role in puppet so then it’s installed on the server.

dracos (approved):

PR #1817, opened 2024-09-12
Homepage and Navbar Improvements

lucascumsille:
Screenshot 2024-09-12 at 12 06 36

I tried decreasing the width to 28% and It works using my name.


ajparsons (approved):
Seems ok to me, will merge now.

PR #1820, opened 2024-09-24
Re-enable popular searches

ajparsons:
Ok, this is adding a few more parameters to the recent searches option.

I've added explicitly some things that were loosely managed through the search caching. 

- Explicit number of IP addresses required to hit the threshold for inclusion. 
- Reduce number of items listed to 5.
- Easier to add exclusions.

Ideally, we'd be able to add exclusions through the admin interface - but this means just adding to an array if we need to make a change. 

In the interest of getting it back up - I've stopped halfway through adding it back to the devolved homepages. 

There is now support for limiting by section - so Scotland could just show Scottish searches. This works fine - but where it needs more work is pulling out when searches for people are devolved politicians (and given the lower daily search volume) having a more useful full back for the search (I think we could have something configured in the admin - but adding a few new things to the admin felt like pushing the scope too much).

### DIFF
--- a/www/includes/easyparliament/searchlog.php
+++ b/www/includes/easyparliament/searchlog.php
@@ -153,7 +153,7 @@ class SEARCHLOG {
         $visible_name = preg_replace('/"/', '', $query);
 
         if (isset($section_string)) {
-            $visible_name = preg_replace('/' . $section_string . '/', '', $visible_name);
+            $visible_name = preg_replace('/' . preg_quote($section_string) . '/', '', $visible_name);
         }
 
         $rowarray = $row;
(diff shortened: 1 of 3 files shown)


www/includes/easyparliament/searchlog.php
```diff
@@ -119,6 +152,10 @@ public function _db_row_to_array($row) {
         }
         $visible_name = preg_replace('/"/', '', $query);
 
+        if (isset($section_string)) {
+            $visible_name = preg_replace('/' . $section_string . '/', '', $visible_name);
```
dracos, line 156: shouldn't ever happen but put a preg_quote around the $section_string anyway.

dracos (approved):
just quote the regex is all

PR #1830, opened 2024-10-03
New alert page

lucascumsille:
I left comments in the code to further explain the logic for all the views/scenarios. I left some in the front end so they are easier to see, but in the code they have the class `mockup-internal-comment`

### View if the user has some alert groups created, 

https://github.com/user-attachments/assets/f7291f6c-44f8-42da-ade7-e9b744345582

Screenshot 2024-10-03 at 11 00 39

Screenshot 2024-10-02 at 12 10 50

### Creating a new alert
https://github.com/user-attachments/assets/5aec736f-d592-43a1-ba27-b9adee6fdf94

### No alert for Keywords, but the user has a subscription to Keir Starmer and doesn't have subscription to his own MP
Screenshot 2024-10-03 at 10 47 58

### No alert for Keywords, but the user has a subscription for Keir Starmer and his own MP

Screenshot 2024-10-03 at 10 47 50

### DIFF
--- a/www/docs/js/main.js
+++ b/www/docs/js/main.js
@@ -423,6 +423,77 @@ function wrap_error($message){
   return '<div class="donate-form__error-wrapper"><p class="donate-form__error">' + $message + '</p></div>';
 }
 
+function createAccordion(triggerSelector, contentSelector) {
+  var triggers = document.querySelectorAll(triggerSelector);
+  
+  triggers.forEach(function(trigger) {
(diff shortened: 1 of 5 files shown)


lucascumsille:
@struan and @ajparsons Thanks for the feedback @ajparsons. This is the current mockup with the latest from feedback included. I left some comments about the logic and how the elements should interact + scenarios, but @struan let me know if you want to jump on a meeting =)

PR #1834, opened 2024-10-09
Homepage class tidy and tweaks

### LINKED ISSUE #1832: Suggestion: "Recent votes" on homepage should indicate in which chamber the vote occurred
image

It feels a little odd that there's no "Lords" or "Commons" reference underneath these (and that it doesn't specifically mention that there *are* votes not shown in this box from other parliaments).

dracos:
Don't know who should review something like this currently :) The only actual change should be displaying Welsh votes at all, moving Scottish votes underneath Scottish debates, and including Commons/Lords on UK (fixes #1832) on their homepages. The rest is just trying to tidy all these homepages to use the same homepage base class, which seemed sensible whilst I was there, they're mostly the same now (more could probably be done, combining templates and whatnot, but seemed enough for now).

### DIFF
--- a/classes/Divisions.php
+++ b/classes/Divisions.php
@@ -171,6 +171,7 @@ class Divisions {
                 'url' => $url->generate() . $anchor,
                 'title' => "$debate[section_body] : $debate[subsection_body]",
                 'date' => $debate['hdate'],
+                'major' => $debate['major'],
             ];
         }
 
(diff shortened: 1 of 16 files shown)


ajparsons:
I'll have a look this week - seems sensible.

struan (changes requested):
One actual bug and one only a bug if there is no content which does not happen outside of dev environment bugs.

Otherwise it all looks to make sense.

struan (approved):
yay!

ajparsons:
I've fixed the rebase issue - I think it was unhappy about the popular search item being removed from homepage in master, while that was all being removed and renamed here. 

I can just merge this if no reasons not to.

PR #1835, opened 2024-10-09
use configured smarthost for PHP emails

struan:
This updates the utility class to use the configured smarthost, plus adds a port configuration, and then updates docker to use mailhog and configure php to use this. Largely to make testing things like alert signups etc easier in docker.


dracos (approved):
👍  For my reference as much as anything, OPTION_SMTP_SMARTHOST was previously only being used by scripts/ calling EmailUtil::send_email.

Issue #1833, opened 2024-10-07
Usability: Sidebar nav calendars' links lack contrast


handelaar:
![image](https://github.com/user-attachments/assets/81ea38e9-f589-46dd-b162-f0cb5c228770)

I'm old, sure, but the linked dates in these are not at all well-distinguished visually from the dates which are not linked.  

Maybe add a background tint to days with data?  

Or even just make the links bold (tried that in Inspector and it's still not great but definitely a lot better)?

handelaar:
eg: 

![image](https://github.com/user-attachments/assets/7b5f3cde-83b9-4f26-a15d-e228525c2e4e)

lucascumsille:
Thanks for raising this issue @handelaar we just merged to master a fix to make the links in the calendar more noticeable and also improve their hover and focus state so at least two properties are different compare to the default state.

PR #1836, opened 2024-10-09
Improved day links accessibility for calendar component

lucascumsille:
Fixes: https://github.com/mysociety/theyworkforyou/issues/1833

https://github.com/user-attachments/assets/29b09377-c00a-4383-b489-85ed8a505e9e

Just in case the "Green one" is the active day

### DIFF
--- a/www/docs/style/sass/pages/_business.scss
+++ b/www/docs/style/sass/pages/_business.scss
@@ -224,10 +224,13 @@ a.business-list__title {
 
     th {
         text-align: inherit;
+        // This min-width minimises the odd uneven day width.
+        min-width: calc(100% / 7);
+        padding: 0;
     }
 
     td {
-        padding: 0;
+        padding: 0.1em;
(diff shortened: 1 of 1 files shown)


ajparsons:
Good fix - but I think this specific blue looks like a "you have selected the following date range".

A lighter one maybe (example below, but that then has problems with the white text...)? Otherwise seems fine.

image

lucascumsille:
@ajparsons  I was hoping the space would be enough to make obvious that is not a range, but it didn't work. Instead of going for adding background, I went for text-decoration + font-weight to make it pop up more. I think now they are quite noticeable a link. I added some off-set for the underline so the numbers are easier to read as well.

Let me know what you think.

lucascumsille:
@ajparsons I updated the video at the top.

ajparsons (approved):
Ah, that's a great fix - loaded it up and looks fine!

Just need to squish the fixup commit when merging.

Issue #1824, opened 2024-09-26
Improved alerts creation flow and dashboard


ajparsons:
This ticket is about improving the user interface of managing reports, and the database changes that might be needed to support that.

The main goals are to:

- Make it easier to add more complicated queries (including prompting on possible alternate terms)
- Make it easier to review and manage existing queries. 

Worth referring to the [user stories from past research](https://docs.google.com/document/d/1Jx4wTlyr69HJXzTEHKCTpjMhKCy4V1MuWW3MxPetAkI/edit#heading=h.dy7eu8vu6ke5) .

## Adding more complicated queries.

We want a better flow for creating keyword email alerts. 

This might double as an alternative/improvement to the advanced search interface. 

What we specifically want is to make it simpler to do ORs - both manually and using suggestions from our similarity data (as in cape).  

The thinking here is for political monitoring, it's less complicated AND NOTs, and more 'here is a list of related terms I want to know about'. 

One possible flow is:

- I want to create an email alert
- enter terms and filters (new form)
- would you like to also include any of the following suggested terms (tickboxes)
- search preview of the results, would you like to revise?
- set frequency, and name of alert (name might be suggested from first term).

Doesn't have to be this exactly, but we want a clear flow - some screenshots with notes on existing process in [this doc](https://docs.google.com/document/d/1Jx4wTlyr69HJXzTEHKCTpjMhKCy4V1MuWW3MxPetAkI/edit#heading=h.dy7eu8vu6ke5) - and especially something that's very clear how to make it work for devolved parliament searches. 

This flow should also work to modify existing alerts. 

For person based alerts - need to be able to select if it's speeches AND votes or just one of those. 

## Managing multiple alerts

Having more complicated alerts - we want an interface that helps us manage and view them. This dashboard should in a minimal way make the alerts useful even without checking emails. 

We want to be able to:

- View all queries - and be able to sort by name, date created, or the last time the alert found something. 
- Have a link to modify a query rather than just removing it.
- Be able to click through from the query to the search results (see below, but this might include some customisation of the search page). 

## Additional information this might mean storing:

- date of last match (this might be single last match date - or alternatively a json dict of key statistics - e.g. hits all time, last year, last month, last day, last_date matched)
- store longer queries (complex queries might plausibly be longer than 255).
[...]

PR #1837, opened 2024-10-09
Improve design of alerts page

struan:
This is the first PR in a series to update the look of the alerts page. It only updates the display of existing alerts and not the creation etc. It splits alerts into keyword alerts and representative alerts, although _x is mentioned_ alerts still count as keyword alerts for now.

It doesn't really touch any of the back end other to add some extra properties to alert objects.

It's also missing any translations for the new text.

I'm not proposing this is merged, but is used as the first part of several which will include changes to the alert creation UX.

Part of #1824

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get -qq update && apt-get -qq install \
       php-xdebug \
       gettext \
       rsync \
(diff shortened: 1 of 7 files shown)


struan:
Closing in favour of #1851

Issue #1825, opened 2024-09-26
Political monitoring backend changes


ajparsons:
This ticket is around changes to the database and email sender to be able to customise the content further. 

We want to be able to do:

- customisable period of email digests (daily/weekly)
- include/don’t include votes for a speaker alert (boolean, or future proof with list of accepted values?)
- use the announcement system to add content to the top of a message.

PR #1842, opened 2024-10-29
enable using announcements to add content to alerts

struan:
If an announcement has a location of 'alerts' then include that at the top of all alerts. The HTML version either uses "Read more" or the `button_text` property for the link.

Announcements are added at the top of the alerts.

Part of #1825

### DIFF
--- a/scripts/alertmailer.php
+++ b/scripts/alertmailer.php
@@ -13,6 +13,25 @@ include_once '../www/includes/easyparliament/init.php';
 ini_set('memory_limit', -1);
 include_once INCLUDESPATH . 'easyparliament/member.php';
 
+$announcement_manager = new \MySociety\TheyWorkForYou\Model\AnnouncementManagement();
+$banner = $announcement_manager->get_random_valid_item('alerts');
+
+$text_banner = '';
+$html_banner = '';
+if ($banner) {
+    $text_banner = $banner->title . "\n\n" . $banner->content . "\n\n";
+    $html_banner = '<p><strong>' . $banner->title . "</strong></p><p>" . $banner->content . '</p>';
+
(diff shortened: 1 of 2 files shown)


scripts/alertmailer.php
```diff
@@ -13,6 +13,20 @@ function mlog($message) {
...
+
+$text_banner = '';
+$html_banner = '';
+if ($banner) {
+    $text_banner = $banner->title . "\n\n" . $banner->content . "\n\n" . $banner->url . "\n\n";
+    $link_text = "Read more";
+    if (property_exists($banner, 'button_text')) {
+        $link_text = $banner->button_text;
+    }
+    $html_banner = '<strong>' . $banner->title . "</strong><p>" . $banner->content . '</p><p><a href="' . $banner->url . '">' . $link_text . '</a></p>';
```
dracos, line 27: Could we only output the second p if link_text isn't blank? People could put their own links inside the content presumably, and then not need a "Read more"? e.g. the last time we did this manually, the added text was:
```
 <p style="font-size: 16px;">
The UK Parliament is back – read our blog post on
<a href="https://www.mysociety.org/2024/07/04/theyworkforyou-understanding-the-new-parliament/">resources to help you understand the new Parliament</a>,
and if you enjoy your alerts,
<a href="https://www.theyworkforyou.com/support-us/">donate to support our work</a>.
            </p>
```

dracos (commented):
When we've manually done this, we've put the HTML p in the header under the linked logo, before the h1 - do we want it there, or under the h1 as is done here?

dracos (approved):
Oh, thinking about it, as banner->content is same in both text and HTML, we couldn't really put HTML in it anyway, unless we had two different contents 🤷

PR #1844, opened 2024-10-30
Add option to allow users to hide speaker votes in alerts

struan:
This lets users set a flag on an alert to avoid including recent votes in an alert for "when x speaks"

Part of #1825


lucascumsille:
@struan I added some commits to tidy up indentation and class name convention, but I think I still need some time to work on this.

lucascumsille:
@struan I think I'm done pushing commit here. I did some rebasing, let me know if it gets too confusing and we can jump on a meeting.

Not sure if it is too much faff, but I noticed when there is only one keyword the accordion works well, but when you have multiple keywords the reading becomes a bit unclear, because there is no separation. For example here:
Screenshot 2025-01-21 at 10 55 12
I think if we could split the multiple keywords with commas or each keyword wrapped on a span elements it would allows to have more control over that.

There is one thing I noticed when creating an alert. I tried adding multiple keywords to exclude, but when creating the alert it will only leave one exclusion and all the rest would go to the list of **included** keywords.

https://github.com/user-attachments/assets/421df686-8c23-463d-a3e1-09f058c7c336

PR #1846, opened 2024-10-31
new alerts front end

struan:

MP alerts continue to use the existing mechanism.

Part of #1824


lucascumsille:
@ajparsons and @struan not sure if I'll be able to finish this before going on holiday so I'll leave a few notes for when I'm back and I can have a catch up with Struan:

Just in case I'll keep updating this comment during the day.

Don't need to reply to this comment now, happy to have a catch up when I'm back.

### Alert page
- [ ] Is it possible to move the "MP is mentioned" from the keywords section. If it's too much hassle I should add a message in the MP section telling users the MP mention is in the key word section.

### Create Alert: Keywords

- [ ] I want to check with Struan if I'm missing something in the journey, but so far so good.
- [ ] 


### Create Alert : MPs

- [ ] I'm a bit confused when I add a postcode and then press search, I'm not actually searching, is instantly creating the alert. Is this the behaviour we expect? if so I'll probably change the name
Screenshot 2024-11-18 at 10 14 28

- [ ] Double checking but currently if I add a postcode I create an alert for my MP, 

- [ ] Related to the previous point I can see there is a 'Subscribe' button but is wrapped on a conditional. I couldn't understand the logic behind it though
```php

        <?php if ($pid || $keyword) { ?>
        <button type="submit" class="button" name="mp_step" value="mp_confirm">
          <span><?= gettext('Subscribe') ?></span>
          <i aria-hidden="true" class="fi-megaphone"></i>
        </button>
        <?php } else { ?>
``` 

- [ ] Double checking, but am I right to think the create alert only has one step at the moment?

struan:
We previously immediately created an alert if you put in a post code so I retained that behaviour.

struan:
Closing in favour of #1851 which tidies all this up into some sensible commits.

PR #1847, opened 2024-11-04
Export debates without GID

ajparsons:
Joining with the other tables to map GIDs to debates loses the (few, but all relevant to policies) - pre 2004 votes. 

As a result, doing a union with just the divisions in there with no GID.


dracos (approved):
Have you tested this works okay? It feels like there should be a nicer way to do this than a UNION, but anything I tried was incredibly slow, perhaps you've found the same.

ajparsons:
Yeah, I tried a few different join approaches but they were very slow. Fully concede this is a horrible query. 

(I guess the long term alternative it is to go in and add the gids, but this is the path of least resistance).

PR #1849, opened 2024-12-12
Add basic formatting to new regmem data

ajparsons:
The new regmem data has more markup for formatting, which is good because we're also dumping a lot more data into the display now.

This is a first pass at styling that data to make the structure (especially of parent/child items - which are new)- clearer. 

image

### DIFF
--- a/www/docs/style/sass/pages/_mp.scss
+++ b/www/docs/style/sass/pages/_mp.scss
@@ -481,7 +481,6 @@
 
     .register {
         .regmemcategory {
-            border-top: 1px solid $borders;
             font-weight: $weight_semibold;
             margin-top: em-calc(16);
             margin-bottom: em-calc(8);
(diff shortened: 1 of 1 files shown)


www/docs/style/sass/pages/_mp.scss
```diff
@@ -790,4 +789,26 @@ a[href^="https://www.publicwhip.org"] {
...
+    font-weight: normal;
+    text-decoration: none;
+}
+
+.interest-details-list {
+    font-size: 0.8em;
+}
+
+.child-item-header {
+    text-decoration: underline;
```
dracos, line 810: This doesn't feel to visually fit - if you want a smaller heading then I'd make it bold and smaller font (or make the other one bigger), we need a heading structure. Or indent it maybe. (Also, with conversation in other PR, really this should be a hN if it's a heading, not a div, but I realise that might not be doable now.)
Feels like all the metadata could be hidden behind a details element and only shown if requested? It reads quite dry/technical (ISO dates, repetition of amount without pound symbol, etc). And then it wouldn't need the heading so much.

dracos (changes requested):

www/docs/style/sass/pages/_mp.scss
ajparsons: Based on revised heading structure from other PR: image

Agree on the detail bit but I think that needs to wait for another pass - ideally we want to be passing storing the json in the database and get the formatting into our normal template (it's in three places at the moment)

dracos: Yep!

dracos (approved):

PR #1850, opened 2024-12-16
Add mailchimp opt-in uploader

ajparsons:
This PR is a simplified version of https://github.com/mysociety/theyworkforyou/pull/1813

Currently we have contact optins when people create accounts, but we don't automatically move them to mailchimp.

This adds uploading new registrations from yesterday to the mailchimp account. 

This also adds a basic django ORM approach to accessing the TWFY database from python.

It needs a new MAILCHIMP_API_KEY in the config.


src/twfy_tools/db/models.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import Optional
+
+from django.db import models
+
+from twfy_tools.db import django_setup as django_setup
+
+from ..common.enum_backport import StrEnum
+from .model_helper import UnmanagedDataclassModel, field
+
+datetime_min = datetime.datetime(1, 1, 1, 0, 0, 0)
```
dracos, line 18: `datetime.min` exists :)

src/twfy_tools/utils/contact_io.py
```diff
@@ -0,0 +1,91 @@
...
+app = Typer()
+
+
+def get_internal_optin_id(
+    interest_group: str, interest_name: str
+) -> InterestInternalId:
+    client = MailChimpHandler(config.MAILCHIMP_API_KEY)
+
+    # get internal id for the mailing list
+    mailing_list_id = client.list_name_to_unique_id(mailing_list_name)
```
dracos, line 28: I mean, it doesn't matter apart from the fact I have to deal with inefficient bots bringing down our services and so I don't want to ever be thought of as something like that, but can we call this once rather than 4 times (once per option, then again in line 65 before the batch add? :)

src/twfy_tools/db/model_helper.py
```diff
@@ -0,0 +1,59 @@
+from typing import Any, Callable, TypeVar
```
dracos, line 1: I don't understand the typing but I'm sure it's lovely

src/twfy_tools/common/mailchimp.py
```diff
@@ -0,0 +1,686 @@
+import datetime
```
dracos, line 1: Sorry, this is quite a big file and lots of it doesn't appear to be used by this PR. Do you want me to review the whole file? Seems to be quite a lot of repetitive boilerplate, e.g. list_name_to_unique_id/ segment_name_to_unique_id/ template_name_to_unique_id all seem to do the same thing, just on a different thing; I don't see why MailchimpHandler has to call top-level functions of exactly the same name with all the arguments having to be repeated, that kind of thing. I'm sure it's all fine 🤷

dracos (changes requested):
Couple of small things is all

src/twfy_tools/common/mailchimp.py
ajparsons: Sorry yeah, it's a general helper file I've built up for projects that use mailchimp - the duplication was for backwards compatibility. Could be simplified for this but I do know it at least works in this form.

src/twfy_tools/db/models.py
ajparsons: ooh, added.

src/twfy_tools/utils/contact_io.py
ajparsons: Moved to a cached function!

PR #1851, opened 2024-12-17
new alerts front end

struan:

MP alerts continue to use the existing mechanism.

Part of https://github.com/mysociety/theyworkforyou/issues/1824

### DIFF
--- a/tests/SearchTest.php
+++ b/tests/SearchTest.php
@@ -121,7 +121,7 @@ class SearchTest extends FetchPageTestCase {
     public function testSearchPageMP() {
         $page = $this->fetch_page([ 'q' => 'Mary Smith' ]);
         $this->assertStringContainsString('Mary Smith', $page);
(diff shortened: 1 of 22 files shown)


tests/SearchTest.php
```diff
@@ -121,7 +121,7 @@ public function testSearchPage() {
     public function testSearchPageMP() {
         $page = $this->fetch_page([ 'q' => 'Mary Smith' ]);
         $this->assertStringContainsString('Mary Smith', $page);
-        $this->assertStringContainsString('MP, Amber Valley', $page);
+        $this->assertMatchesRegularExpression('/MP *for Amber Valley/', $page);
```
dracos, line 124: Should it be a `+` to make sure "MPfor Amber Valley" doesn't get through? :)

struan:
Closed in favour of #1890

PR #1853, opened 2025-02-10
[alerts] Update welsh terms from fuzzy

ajparsons:
Some fuzzy terms have snuck in and the test is failing.

### DIFF
--- a/locale/cy_GB.UTF-8/LC_MESSAGES/TheyWorkForYou.po
+++ b/locale/cy_GB.UTF-8/LC_MESSAGES/TheyWorkForYou.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-02-06 16:34+0000\n"
+"POT-Creation-Date: 2025-02-10 20:19+0000\n"
 "PO-Revision-Date: 2023-03-20 17:59-0000\n"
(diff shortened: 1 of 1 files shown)


locale/cy_GB.UTF-8/LC_MESSAGES/TheyWorkForYou.po
```diff
@@ -525,7 +525,17 @@ msgstr "Cwestiynau Maer Llundain"
 msgid "Senedd debates"
 msgstr "Dadleuon y Senedd"
 
-#: scripts/alertmailer.php:278
```
dracos, line 528: This seems to have been inserted in an odd place, we've lost this comment line for the "as a teller"

ajparsons: So at this point  the comment lines that are generated are a big diff from what's committed. I think I should just do another commit catching it up, having separated out the main change this time.

ajparsons: Added that

dracos: The source .po file should also be committed then?
Thanks, yes, you shouldn't be manually editing .po files :)

struan (approved):
I mean, I think this is fine and passes tests so sure, yes.

PR #1854, opened 2025-02-14
Twfy votes connections

ajparsons:
This is a basic PR to connect to the new votes site rather than the prototype one. 

Two changes:

- Some adjustments to fields in the output of the new data.
- After divisions are exported, it sends a message to the votes site to import the past days data. 

This needs the config updated to the new votes site, the updater will say something but not fail if the refresh token isn't present.

### DIFF
--- a/scripts/json2db.pl
+++ b/scripts/json2db.pl
@@ -71,7 +71,7 @@ sub get_url {
     # simplify retrying urls if they fail
     my ($url, $retries) = @_;
     my $ua = LWP::UserAgent->new;
-    $ua->timeout(10);  # 10 second timeout
+    $ua->timeout(30);  # 30 second timeout
     my $response = $ua->get($url);
     if ($response->is_success) {
         return $response->decoded_content;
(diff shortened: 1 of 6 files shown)


scripts/json2db.pl
```diff
@@ -141,17 +141,23 @@ sub fetch_policies {
     # so can be just be simply stored for reference in policy page.
     my $out_agreements = {};
 
-    foreach my $policy (@{$policies->{policies}}) {
+    foreach my $policy (@$policies) {
         my $policy_id = $policy->{"id"};
         foreach my $agreement (@{$policy->{"agreement_links"}}) {
             my $decision = $agreement->{"decision"};
-            my $chamber = $decision->{"chamber"};
+            my $decision_date = $decision->{"date"};
```
struan, line 148: why not use the everywhere there is `$decision->{"ref"}`?

scripts/json2db.pl
```diff
@@ -141,17 +141,23 @@ sub fetch_policies {
...
+            # adjust decision_ref - split by . and remove the last part and put it back together
+            # this is because the decision ref is to a specific line, rather than a linkable section.
+            my @parts = split(/\./, $decision->{"decision_ref"});
+            pop @parts if @parts > 1;
+            my $decision_ref = join(".", @parts);
```
struan, line 153: I feel like maybe calling this `$decision_url_ref` would be a bit clearer or `$division_section_ref` maybe?

struan (approved):
Looks good. Feel free to ignore my tiny niggles :)

I've added an extra commit just because I noticed the issue while testing this and figured I might as well fix it.

scripts/json2db.pl
struan: oh, thinko on my part I meant `$decision->{"date"}` but my brain had already moved on to the ref bit.

ajparsons:
Those are in! Will work out a good timeline to merge because this will now add the new votes content.

PR #1857, opened 2025-02-25
Format alternative syntax

ajparsons:
This isn't necessarily to be merged but just building on some points in https://github.com/mysociety/theyworkforyou/pull/1856

This applies a rule to enforce the braces as opposed to 'if' 'endif' syntax.

This ends up changing relatively few files - all templates - which makes sense as I was using it because it felt a little more readable in mixed files, but also the IDE is good at finding the matching brace. 

Basically if there are two options and we have a preference, we could have a rule.


dracos:
Yeah, happy to have matching brackets, I think it's clearer too (boo python ;) ).

PR #1859, opened 2025-02-26
Wft content 2024

ajparsons:
This is the actual WFT content.

There's still some design tweaks, but I want to review the basic data loading and site navigation.

I'll make the two jsons available to you - they need to go in universal_format_regmem/misc.

To load data:
```
scripts/personinfo upload-enhanced-2024-regmem
```

Questions:

- I'm loading the highlighted interests json direct from disc - is this ok or does it benefit from a cache somewhere.
- highlighted page URL format - have gone for highlighted-interests/2024 to be future safe - but could just be 'highlighted-interests
- I'm duplicating some of the code for the entry and field display because I wanted to tweak this a bit but not feedback into the other one. My intention is aspects of this move back into the main display in the long run, and this enriched register might get removed.


classes/DataClass/DataFrame.php
```diff
@@ -0,0 +1,103 @@
...
+            // make sure all columns are strings rather than ints
+            $this->rows[] = new Row($rowData);
+        }
+    }
+
+    public function toArray(): array {
+        return array_map(fn($row) => $row->toArray(), $this->rows);
+    }
+
+    public function toHTML(?string $url_column = null): string {
```
struan, line 68: I think I would prefer not to have things generated like this because it makes it harder to style the data and in this case the election_register template is quite hard to understand where the content is coming from and how to style it. Probably not for now but I think we should probably change this in the future to move this into the templates.

www/includes/easyparliament/templates/html/misc/_register_entry.php
```diff
@@ -0,0 +1,49 @@
...
+            <?php endif; ?>
+            <?php if ($entry->date_published): ?>
+                <li class="published-date">Published Date: <?= htmlspecialchars($entry->date_published) ?></li>
+            <?php endif; ?>
+            <?php if ($entry->date_updated): ?>
+                <li class="last-updated-date">Last Updated Date: <?= htmlspecialchars($entry->date_updated) ?></li>
+            <?php endif; ?>
+        </ul>
+
+    <?php else: ?>
```
struan, line 25: indentation is not helpful here. And is there a reason this and the following are combined into an else if?

www/includes/easyparliament/templates/html/misc/highlighted_2024.php
```diff
@@ -0,0 +1,95 @@
...
+        <div class="toc-page__col">
+
+            <div class="panel">
+                <h1>Highlighted interests</h1>
+
+                <a name="intro"></a>
+                <p>As part of our 2024/25 WhoFundsThem project, we are highlighting declared interests that we feel merit being more visible, based on known public attitudes towards certain industries.</p>
+                <p>You can read more about our research and choices in the accompanying <a href="http://research.mysociety.org/html/beyond-transparency/">Beyond Transparency report</a> </p>
+
+                <p>Nothing in this section was against the rules for MPs to accept and they have been correctly declared.</p>
```
struan, line 33: I wonder if we should make this more prominent just to head off potential complaints.

www/docs/mp/index.php
```diff
@@ -473,6 +474,20 @@ function ($k) { return $k['score_difference'] >= 2; }
...
 
+    case 'election_register':
+        // Send the output for rendering
+
+        $highlighted_register = MySociety\TheyWorkForYou\DataClass\Regmem\Register::getMisc("highlighted_interests.json");
+
+        $str_id = "uk.org.publicwhip/person/" . $MEMBER->person_id();
+        $highlighted_for_this_mp = $highlighted_register->getPersonFromId($str_id);
+
+        $data['mp_has_highlighted_interests'] = (bool) $highlighted_for_this_mp;
```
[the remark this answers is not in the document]
ajparsons: Yeah that's a good minimum

www/includes/easyparliament/templates/html/mp/election_register.php
```diff
@@ -0,0 +1,204 @@
...
+  word-wrap: break-word;
+overflow: hidden;
+word-wrap: break-word;
+white-space: normal;
+}
+
+td.wrap {
+  white-space: normal;
+}
+</style>
```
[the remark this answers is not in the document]
ajparsons: Oh I forgot about this - yes I can move it into the real css.

classes/DataClass/DataFrame.php
ajparsons: Yeah this was just quickly moving it from the prototype (which was in python, so just used the pandas grids.)

ajparsons:
Will discuss with group monday when the final ror data is in.

- Have added a new register of interests overall page to navigation (which also gives the highlighted page something to sit beneath).
- Various style fixes.
- Possibly gone too far with emojis for navigation. Can review.

struan (approved):

PR #1860, opened 2025-02-26
Fix empty register error

ajparsons:
- bots keep hitting old register pages that aren't linked
- this stops that causing an error


dracos:
Ah, sorry, didn't see this; did basically the same in e1519824e.

PR #1862, opened 2025-03-06
add suppress-progress option to personinfo import

struan:
Only suppresses the progress bar for using in cron jobs

### DIFF
--- a/scripts/dailyupdate
+++ b/scripts/dailyupdate
@@ -29,7 +29,7 @@ chdir $FindBin::Bin;
 system './mpinfoin.pl';
 
 # Update register of interests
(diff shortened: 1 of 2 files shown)


dracos (commented):
Might be easier to set TQDM_DISABLE=1 in the script? :)

ajparsons:
I did not know about that flag. I don't have a preference but might be good to have an example of that lying around because it'll simplify similar problems.

struan:
So, because we have a quiet option I had explicitly handle the TQDM_DISABLE in the script but it is a much tidier approach anyway.

dracos (approved):

PR #1865, opened 2025-03-11
Devolved register launch

ajparsons:
This is the PR to make live the devolved register changes - tidying up non gettext features for wales and adding a by category view feature.

Notes:

- I've marked up the text and will put in a translation request. (have just noted i missed marking up on the time comparision page, but have submitted the relevant text).
- The category page works by loading the latest json. This needed a few extra defaults adding to the data structure (the python loader for the regmem data adds the defaults before storing in the database).
- There's a asymmetry in the devolved parliaments have a menu link *direct* to the category page, while the UK wide one goes to the general explainer page. I'm ok with this because that page has details I want from the main heading. 
- Welsh language version looks fine - as it loads from what's stored in the json all the categories names load ok.
- Outstanding task is to create and link the spreadsheets - but this will just be a bit in the markdown.

image

image
image

### DIFF
--- a/classes/DataClass/Regmem/Category.php
+++ b/classes/DataClass/Regmem/Category.php
@@ -15,8 +15,33 @@ class Category extends BaseModel {
     public ?EntryList $summaries = null;
     public ?EntryList $entries = null;
 
-    public function emoji() {
+    public static function emojiLookup(string $value): string {
         $emoji_lookup = [
+            "Remuneration or other material benefit" => "💼",
+            "Membership/Chairmanship of bodies in receipt of Senedd funds" => "👥",
+            "Record of the employment of family members" => "👪",
+            "Directorships" => "👔",
+            "Gifts, hospitality, material benefit or advantage" => "🎁",
+            "Remunerated employment, office, profession etc" => "💼",
+            "Financial sponsorship" => "💳",
+            "Overseas visits" => "🌍",
+            "Land and property" => "🏠",
+            "Record of membership of societies" => "🤝",
+            "Donations and other support" => "💳",
+            "Gifts, benefits and hospitality" => "🎁",
+            "Visits" => "🌍",
+            "Shareholdings" => "📈",
+            "Land and Property" => "🏠",
+            "Miscellaneous" => "🏷️",
+            "Unremunerated interests" => "🤝",
+            "Family members who benefit from Office Cost Expenditure" => "👪",
+            "Voluntary" => "🤝",
+            "Controlled transactions" => "📊",
+            "Gifts" => "🎁",
+            "Heritable property" => "🏠",
+            "Interest in shares" => "📈",
+            "Overseas visits" => "🌍",
+            "Remuneration and related undertaking" => "💼",
             "Donations and other support (including loans) for activities as an MP" => "💳",
             "Gifts, benefits and hospitality from UK sources" => "🎁",
             "Employment and earnings - Ad hoc payments" => "💼",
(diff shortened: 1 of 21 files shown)


www/docs/interests/category.php
```diff
@@ -0,0 +1,75 @@
...
+} else {
+    $register = MySociety\TheyWorkForYou\DataClass\Regmem\Register::getLatest($chamber);
+}
+
+
+// get the relevant categories from that register
+// and if there is a selected_category_id reduce the register here
+$categories = [];
+
+foreach ($register->persons as $person) {
```
struan, line 41: I understand that we have to do it this way because of the way the data is structured but it seems quite inefficient to do this instead of including the list of categories in the data. And does it not also mean that if no one has declared an entry in a category for the current register we'll not show the category?

www/docs/interests/category.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+// Logic page for the register of interests by category view
+// if no selected category, will just show a list of the categories
+
+include_once '../../includes/easyparliament/init.php';
+
+$chamber = get_http_var("chamber", "house-of-commons");
+$date = get_http_var("date", null);
```
struan, line 9: I might be missing something but there doesn't seem to be an interface to this?

www/includes/easyparliament/templates/html/interests/category.php
```diff
@@ -0,0 +1,137 @@
...
+                            <li><a href="?chamber=<?= $chamber_slug ?>&category_id=<?= $category_id ?>"><?= $category_name ?></a></li>
+                        <?php }; ?>
+                    </ul>
+                <?php }; ?>
+
+
+                <?php foreach ($categories as $category_id => $category_name) { ?>
+                    <?php if ($selected_category_id != $category_id) {
+                        continue;
+                    }; ?>
```
struan, line 47: would it not make more sense to just set a `$selected_category_name` at the data stage and pass that in to the template and then you could avoid this loop?

classes/DataClass/Regmem/Category.php
```diff
@@ -9,14 +9,48 @@
...
+            "Miscellaneous" => "🏷️",
+            "Unremunerated interests" => "🤝",
+            "Family members who benefit from Office Cost Expenditure" => "👪",
+            "Voluntary" => "🤝",
+            "Controlled transactions" => "📊",
+            "Gifts" => "🎁",
+            "Heritable property" => "🏠",
+            "Interest in shares" => "📈",
+            "Overseas visits" => "🌍",
+            "Remuneration and related undertaking" => "💼",
```
struan, line 53: I understand that this is to help indicate which categories are the equivalent across chambers but I'm not entirely convinced by it. However, I am old so maybe not the target market :)

struan (approved):
few small comments that I do not think should hold this up, otherwise this looks good.

www/docs/interests/category.php
ajparsons: There isn't - it felt like a bolt on later and a bit of a tool for our purposes now.

www/docs/interests/category.php
ajparsons: Yeah, this is consequence of choosing a tiered data format (where all the useful data sits under the person and can be sliced for our database). 

Although nothing stops us adding something to the data format (an extra set of category objects under the register) if we thought that was worth sorting out at this point?

www/docs/interests/category.php
struan: was just checking it wasn't a missing thing.

www/docs/interests/category.php
struan: I don't think it's worth delaying anything at the moment. I do wonder if it would make it a bit more helpful for anyone re-using the data but again, we could presumably add this in later.

www/includes/easyparliament/templates/html/interests/category.php
ajparsons: Yep! Done a change on this

PR #1872, opened 2025-04-11
Supporters page

ajparsons:
This changes the supporters page:

* Simplifies copy before donate form.
* Simplifies choices - leans harder into a default value (currently £10 a year). This is the only option visible unless another option is selected. 

On desktop:

image

On mobile, button at top are immediately visible:

image

Clicking the default amount gives a simplified form:

image

Clicking the  other gives the normal form:

image

### DIFF
--- a/markdown/support-us.md
+++ b/markdown/support-us.md
@@ -1,26 +1,31 @@
-# Support TheyWorkForYou and mySociety
+# Support TheyWorkForYou for {{ verbose_amount }}
 
-[Skip to donate form.](#donate-form)
+**Or less, or more.**
+ 
 
-We want MPs to meet the standards and expectations of the people who elected them - **you**!
+<button type="button" class="default-donate-button button button-primary button-mobile button--large" style="margin-right: 10px" onclick="restrict_to_default('{{ how_much }}', '{{ payment_type }}')">
+        Donate {{ verbose_amount }}
+</button>
 
-TheyWorkForYou started twenty years ago as a volunteer project, when a bunch of people got together to make Parliament more transparent. We don’t have to wait for a better politics to be given to us – **we can work together to make it happen now**.
+<input type="button" value="Donate another amount" class="open-form-button button button-primary button-mobile  button--large">
 
-These sites need attention every day to:
+For over 20 years, TheyWorkForYou has been making our democracy more transparent and our politicians more accountable. We need your support to: 
 
-* Keep the data flowing and ensuring that the comprehensive data about politicians is up to date.
-* Make improvements such as [recent changes to voting records](https://www.mysociety.org/2024/04/23/improving-the-focus-of-theyworkforyous-voting-summaries/).
-* Provide support on how to use the sites and the data we provide - we’re here for you.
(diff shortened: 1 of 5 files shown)


www/docs/support-us/index.php
```diff
@@ -11,5 +11,74 @@
...
+    ],
+    'annually' => [
+        '5' => '£5',
+        '10' => '£10',
+        '50' => '£50',
+    ],
+    'one-off' => [
+        '5' => '£5',
+        '10' => '£10',
+        '50' => '$50',
```
dracos, line 31: $

www/docs/support-us/index.php
```diff
@@ -11,5 +11,74 @@
...
+    'monthly' => '5',
+    'annually' => '10',
+    'one-off' => '10',
+];
+
+
+$default_type = 'annually';
+
+# use the how-often parameter if set, if not default to option at end of line (options are 'monthly', 'annually', or 'one-off')
+$payment_type = get_http_var('how-often', $default_type);
```
dracos, line 46: Please add a check here that it is one of the valid options, otherwise we can get errors when bots spider with random crap in the parameters.

www/includes/easyparliament/templates/html/donate/_stripe_donate.php
```diff
@@ -1,53 +1,20 @@
...
-<form class="donate-form" method="post" name="donation_form">
+
+<button type="button" id="default_donate_button" class="button button-primary button--large" style="margin-right: 10px" onclick="restrict_to_default('<?= $how_much ?>', '<?= $payment_type ?>')">
+        Donate <?= $verbose_amount ?>
+</button>
+
+<input type="button" value="Donate another amount" class="open-form-button button button-primary button--large" >
+    
+
+<form class="donate-form" method="post" name="donation_form" style="display: None";>
```
dracos, line 17: Please have the default be visible, and use `html.js` in a stylesheet parent to set it to display none if JS is available. Otherwise there's no way to see the form if the JS doesn't work.

markdown/support-us.md
```diff
@@ -1,26 +1,28 @@
...
+</button>
 
-We want MPs to meet the standards and expectations of the people who elected them - **you**!
+<input type="button"value="Donate another amount" class="open-form-button button button-primary button-mobile  button--large" >
 
-TheyWorkForYou started twenty years ago as a volunteer project, when a bunch of people got together to make Parliament more transparent. We don’t have to wait for a better politics to be given to us – **we can work together to make it happen now**.
+For over 20 years, TheyWorkForYou has been making our democracy more transparent and our politicans more accountable. We need your support to: 
 
-These sites need attention every day to:
+* **Hold Power to Account** - With your help, we highlight what our representatives are doing – so they know the public is watching. By making votes more visible, the register of interests easier to explore, and sending email alerts on issues you care about, we create accountability that changes.
```
dracos, line 11: en dashes everywhere please, this one has one and also a hyphen :)

www/includes/easyparliament/templates/html/donate/_stripe_donate.php
ajparsons: There's a general thing here that the stripe form needs javascript - but can't remember if I moved the warning about this *out* of the now invisible form. Either way, yes , can fix.

ajparsons:
Forgot to re-request review - but should have addressed!

dracos (approved):
Thanks :)

PR #1874, opened 2025-04-11
Dynamic open graph

ajparsons:
This PR the ability to create dynamic open graph images for pages - improving how our site appears when shared on social media. 

There's a function that takes a header, subheader and returns a link including a hash of that with a salt in the settings. 

(The salt stops arbitrary title generation from the outside - unclear how much of a problem this is, but means it returns a 404 rather than pushing unvalidated external input into the image generator). 

This will then create something like the following:

image (2)

image (3)

There are four versions of the base image for the different Parliaments. 

This is set for the new register of interest pages, markdown pages, and representative pages. Others will default to a new style image. In principle, can come back and do the debate titles, but these have a more variable length and can be fiddly. 

The font has been loaded onto the servers.


www/docs/opengraph/image.php
```diff
@@ -0,0 +1,97 @@
...
+// Truncate the hash to the first 10 characters for validation
+$expectedHash = substr(hash_hmac('sha256', $heading . $subheading, OPENGRAPH_IMAGE_SALT), 0, 10);
+if (!hash_equals($expectedHash, $hash)) {
+    http_response_code(403);
+    echo 'Invalid hash';
+    exit;
+}
+
+// Wrap the title text to a new line if it exceeds 16 characters
+if (strlen($heading) > 16) {
```
struan, line 56: FWIW this means roughly 10% of MPs won't ever get a subheading

classes/Url.php
```diff
@@ -190,4 +190,52 @@ public function generate($encode = "html", $overrideVars = []) {
...
+     * @param string $subheader The subheader text for the social image.
+     * @param string $parliament The parliament to use (default is "uk").
+     * @return string The URL to the generated social image.
+     */
+    public static function generateSocialImageUrl(string $header, string $subheader = "", string $parliament = "uk"): string {
+
+
+        function chamber_to_parliament($chamber) {
+            switch ($chamber) {
+                case 'house-of-commons':
```
struan, line 205: these also needs to handle 'uk-commons', 'uk-lords' and 'wales' otherwise you get an "Invalid parliament" error from the opengraph image generator. (or the open graph generator needs to accept those but that would mean more images).

www/docs/mp/index.php
```diff
@@ -492,27 +494,22 @@ function ($k) { return $k['score_difference'] >= 2; }
...
         MySociety\TheyWorkForYou\Renderer::output('mp/election_register', $data);
 
         // no break
     case 'register':
+        $data['og_image'] = \MySociety\TheyWorkForYou\Url::generateSocialImageUrl($member_name, 'Register of Interests', $data['current_assembly']);
         // Send the output for rendering
         MySociety\TheyWorkForYou\Renderer::output('mp/register', $data);
 
         // no break
-    case 'policy_set_svg':
```
struan, line 506: given we're removing this we should remove the code under votes that sets the og_image to use these as it now links to the profile page.

www/docs/mp/index.php
```diff
@@ -383,10 +383,10 @@
 // is the party we're comparing this MP to different from the party they're currently in?
 $data['party_switcher'] = (slugify($data['current_party_comparison']) != slugify($data["comparison_party"]));
 
-// Do any necessary extra work based on the page type, and send for rendering.
+// Update the social image URL generation logic
 switch ($pagetype) {
-
     case 'votes':
+        $data['og_image'] = \MySociety\TheyWorkForYou\Url::generateSocialImageUrl($member_name, 'Voting Summaries', $data['current_assembly']);
```
struan, line 389: instead of repeating this we could set an `$og_image_subtitle` variable here and then after the switch have a single place that the `$data['og_image']` is set. Would also need to remove the special case setting this for single policy pages but as that link no longer works that seems fine.

www/includes/easyparliament/templates/html/header.php
```diff
@@ -29,8 +29,8 @@
   <?php if ($og_image) { ?>
     <meta property="og:image" content="<?= $og_image ?>">
     <meta property="og:image:type" content="image/png">
-    <meta property="og:image:width" content="1000">
-    <meta property="og:image:height" content="500">
+    <meta property="og:image:width" content="1200">
+    <meta property="og:image:height" content="630">
   <?php } else { ?>
     <meta property="og:image" content="https://<?= DOMAIN ?>/images/og/social_<?= $current_assembly ?? 'uk' ?>.jpg">
     <meta property="og:image:type" content="image/jpeg">
```
struan, line 36: There is no `social_uk.jpg` image so the extension here should be a png. Some of the social graphics are jpgs and some are png so need to standardise for this to work.

`$current_assembly` does not seem to be consistently set so e.g. the scottish homepage gets the uk parliament link.

www/docs/mp/index.php
ajparsons: Not sure - the rendering happens inside the switch so not sure we can tidy it that way.

www/includes/easyparliament/templates/html/header.php
ajparsons: All of the social graphics *should* be pngs. I've copied over the old London Assembly as a jpg, and I'll just replace that with the UK one as a holder if we end up coming back to the LA.

www/docs/mp/index.php
struan: oh, doh! so it does. ignore this.

www/docs/opengraph/image.php
ajparsons: I've set this to 25 and reduced total font size - looks good for current MPs.

www/includes/easyparliament/templates/html/header.php
ajparsons: I've added the current_assembly in a few obvious places. It's not the end of the world to fall back to the UK one, but this should catch the obvious ones.

ajparsons:
Ok! Good issues. Should have cleared them. I think the 'current_assembly' change might interact with the outstanding homepage tidy-up PR, but should be easy to sort. 

The twfy-votes version of this dynamically adjusts font size for different lengths - but the quick approach fixed here should be ok.

classes/Homepage.php
```diff
@@ -47,6 +47,7 @@ public function display() {
         $data['topics'] = $this->getFrontPageTopics();
         $data['divisions'] = $this->getRecentDivisions();
         $data['search_box'] = $this->getSearchBox($data);
+        $data['current_assembly'] = "scotland";
```
struan, line 50: I don't think this is correct as it means the uk page gets scotland as the current assembly.

Maybe make current_assembly into a property and then override that in SPHomepage?

struan (changes requested):
One more change I'm afraid.

We should maybe also think about caching but possible the load balances will do this well enough for our purposes.

ajparsons:
How would the caching work for this by default?

(I guess one approach would be to store a file for a set of input somewhere and then serve that if it came up again - would be nice to avoid as fiddly, but doable)

struan:
well, we could send cache headers with a reasonably long expiry date given the content of these images isn't likely to change. I think, and sam would confirm, that the load balancers will cache things for a bit too. I think things like facebook already cache opengraph images so it might not be that much of an issue, but would need to look into whether all the query string stuff upsets that.

ajparsons:
I have fixed the homepage approach (latest fixup) - and I think the view is otherwise we don't need special caching?

struan (approved):
yay!

PR #1875, opened 2025-04-17
improve handling of missing/bad registers

struan:
If someone puts a bad value in the chamber arg then throw a 404, otherwise carry on throwing a 500 so we are warned if expected registers aren't there.


ajparsons (approved):
Sorry this is more fiddly than i thought! Looks good though and works when I tested it.

PR #1878, opened 2025-05-15
Alert front end rebase ap

ajparsons:
Ok, this is looking really really good! I've just gone in and tweaked stuff rather than making a list.

This is mostly a set of formatting/copy changes to the alerts display and view - but I've also made some changes to the logic. 

- The search results bit wasn't working for the main view - so I've tided up connections between the two places that is used. Similarly, copied the way the results view url worked inside the form itself. 
- I've removed the 'don't include speaker votes' bit from the *keyword* search because this didn't seem relevant (but may have mis-understood).
- For the MP alert search,  I've taken away the shortcuts that automatically subscribe if there is only one result. This is because I was getting unexpected behaviour when i did a typo (got someone's name wrong,  but it *was* a match for a constituency). It's just one extra button and I think works fine.


ajparsons:
I think those new commits make sense to me

struan:
closed in favour of #1890

PR #1882, opened 2025-05-16
Fix-mpinfoin

ajparsons:
Important thing is it fixes the bad if statement introduced by the votes PR.

Building on that, I've made it so the verbose argument works on its own (for debugging).

I've also removed the regmem loader - this was left in for some overlap while the new data was coming in, but now can be removed and speed up the ingest process.

### DIFF
--- a/scripts/mpinfoin.pl
+++ b/scripts/mpinfoin.pl
@@ -27,10 +27,7 @@ my %action;
 my $verbose;
 my $action_count = 0;
 foreach (@ARGV) {
-    if ($_ eq 'regmem') {
-        $action{'regmem'} = 1;
-        $action_count++;
-    } elsif ($_ eq 'links') {
+    if ($_ eq 'links') {
         $action{'links'} = 1;
         $action_count++;
     } elsif ($_ eq 'compile') {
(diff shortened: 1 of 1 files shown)


scripts/mpinfoin.pl
```diff
@@ -58,26 +54,18 @@
...
-find sub { push(@regmemfiles, $_) if /^regmem.*\.xml$/ }, 'scrapedxml/regmem/';
 
 # Read in all the files
 my $twig = XML::Twig->new(
     twig_handlers => {
         'memberinfo' => \&loadmemberinfo,
         'personinfo' => \&loadpersoninfo,
-        'consinfo' => \&loadconsinfo,
-        'regmem' => \&loadregmeminfo
+        'consinfo' => \&loadconsinfo
```
dracos, line 65: Just for info, you can leave final commas on, makes the diff bit easier

scripts/mpinfoin.pl
```diff
@@ -53,26 +54,18 @@
 my $personinfohash;
 my $consinfohash;
 
-# Find latest register of members interests file
+# Change to raw data directory
 chdir mySociety::Config::get('RAWDATA');
```
dracos, line 58: I think this can be removed now as well? It was only doing this to find the regmem files.

dracos (changes requested):
Looks good - one additional thing to remove, I think is all :) 
And in the commit message for regmem, add a sentence saying "it's now done by personinfo upload-all-regmem" or something like that?

ajparsons:
There you go!

(meta note that the social media loading could also go towards the personinfo script at some point, but also it loads quickly and is fine).

PR #1887, opened 2025-06-11
[PBC] Change import of names to use content rather than attribute

dracos:
This matches the debate divisions, and will mean any non-ASCII will be HTML escaped (which is obviously stupid to do nowadays, but it at least matches).


ajparsons:
Then I have no problem!

PR #1888, opened 2025-06-18
Adds new options to the verbose vote description

ajparsons:
This is handling a few new trends becoming more prominent.

 - more one vote policies.
 - highlighted party differences that *narrowly* cross the 60% threshold for inclusion. 

Changes made:

- Drops 'consistently' for single vote policies - can't consistently do things once. 
- Adds 'tended to vote' for votes narrowly above the threshold - there are very few of these, but important to have clear language.
- Includes alignment score for middle-ish descriptions (gives some more context on the "it's just a bit mixed").

End result looks like this:

image

### DIFF
--- a/classes/PolicyDistribution.php
+++ b/classes/PolicyDistribution.php
@@ -90,25 +90,49 @@ class PolicyDistribution {
     }
 
     public function getVerboseScore(): string {
+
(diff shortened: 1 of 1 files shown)


ajparsons:
Walking back and forth in the footsteps of giants.

PR #1890, opened 2025-07-07
New alerts front end

struan:

A final small feature is allowing people to turn off getting MP vote results as part of MP speaks alerts

Part of https://github.com/mysociety/theyworkforyou/issues/1824

### DIFF
--- a/classes/AlertView/Standard.php
+++ b/classes/AlertView/Standard.php
@@ -491,8 +491,8 @@ class Standard extends \MySociety\TheyWorkForYou\AlertView {
             }
             $member_count = count($this->data['members']);
             $cons_count = count($this->data['constituencies']);
-            $mem_cons_count = count($this->data['member_constituencies']);
-            if ($member_count == 0 && $cons_count == 0 && $mem_cons_count == 0) {
+            $mem_cons_count = isset($this->data['member_constituencies']) ? count($this->data['member_constituencies']) : 0;
+            if (!$this->data['alertsearch'] && $member_count == 0 && $cons_count == 0 && $mem_cons_count == 0) {
                 $errors["representative"] = gettext("No matching representative found");
             }
         } elseif ($this->data['pid']) {
@@ -732,7 +732,7 @@ class Standard extends \MySociety\TheyWorkForYou\AlertView {
             $this->data['alertsearch'] = '';
         }
 
-        if ($this->data['alertsearch'] && !(isset($this->data['mistakes']['postcode_and']) || $this->data['members'] || $this->data['pid'])) {
+        if ($this->data['alertsearch'] && !(isset($this->data['mistakes']['postcode_and']) || $this->data['members'] || $this->data['pid'] || $this->data['constituencies'])) {
             $this->data['step'] = 'define';
             $this->data['words'] = [$this->data['alertsearch']];
             $this->data['keywords'] = [$this->data['alertsearch']];


dracos (changes requested):
The last commit feels like it should perhaps have been a number of commits :) It was a large one to review, though it was nice having the set up stuff in previous commits.
Overall, works well; seems to be some bugs with no provided email, not showing all regional reps.

www/includes/easyparliament/templates/html/alert/_alert_step_define.php
```diff
@@ -39,7 +39,7 @@
                   <hr />
                   <div class="alert-form__section">
                     <label for="exclusions"><?= gettext('Are there any phrases you would not like to receive alerts about? (optional)') ?></label>
-                    <input type="text" id="exclusions" name="exclusions" aria-required="true" value="<?= _htmlspecialchars($exclusions) ?>" placeholder="Eg. 'Information Rights'">
+                    <input type="text" id="exclusions" name="exclusions" aria-required="true" value="<?= _htmlspecialchars($exclusions) ?>" placeholder="Eg. Information Rights">
```
dracos, line 42: Could we be consistent in `e.g.`? :)

www/includes/easyparliament/templates/html/alert/_alert_controls.php
```diff
@@ -32,16 +32,17 @@
       <span><?= gettext('Edit alert') ?></span>
     </button>
     <?php } else { ?>
-    <input type="hidden" name="mp_step" value="mp_confirm">
```
dracos: Could we please have some documentation of the flows for future reference, maybe in Standard.php I guess? What is `mp_step` for? I see with this removed there's only one use in a template of `mp_confirm` that's not actually then checked anywhere, is that because it uses the default behaviour? Was this actually creating a new alert each time (that I didn't spot last time)?
(I guess this is a continuation of my 'confused' comment from last review, I'm happy this is fixing all the bugs, but I'm not confident looking back at this in years I will understand what is going on.)

classes/AlertView/Standard.php
```diff
@@ -409,6 +410,10 @@ private function searchForConstituenciesAndMembers() {
             $members_from_words = \MySociety\TheyWorkForYou\Utility\Search::searchMemberDbLookupWithNames($text, true);
             $this->data['members'] = array_merge($members_from_words, $members_from_names);
             [$this->data['constituencies'], $this->data['valid_postcode']] = \MySociety\TheyWorkForYou\Utility\Search::searchConstituenciesByQuery($text, false);
+            if (!$this->data['valid_postcode']) {
+                $this->data['member_constituencies'] = \MySociety\TheyWorkForYou\Utility\Search::searchMembersByConstituency($text);
+            }
+            [$this->data['constituencies'], $this->data['valid_postcode']] = \MySociety\TheyWorkForYou\Utility\Search::searchConstituenciesByQuery($text, false);
```
dracos, line 432: This line is a repeat of line 412.

classes/AlertView/Standard.php
```diff
@@ -409,6 +410,10 @@ private function searchForConstituenciesAndMembers() {
             $members_from_words = \MySociety\TheyWorkForYou\Utility\Search::searchMemberDbLookupWithNames($text, true);
             $this->data['members'] = array_merge($members_from_words, $members_from_names);
             [$this->data['constituencies'], $this->data['valid_postcode']] = \MySociety\TheyWorkForYou\Utility\Search::searchConstituenciesByQuery($text, false);
+            if (!$this->data['valid_postcode']) {
+                $this->data['member_constituencies'] = \MySociety\TheyWorkForYou\Utility\Search::searchMembersByConstituency($text);
```
dracos, line 430: I realise /search looks like it only searches MPs by constituency name at the moment, but (not for now, but please ticket), really we should only have the one way to search for reps by constituency name, not this new function and an entirely different way on search results page.

classes/AlertView/Standard.php
```diff
@@ -454,7 +459,20 @@ private function searchForConstituenciesAndMembers() {
...
                         $cons[] = [ 'member' => $MEMBER, 'constituency' => $constituency ];
                     }
-                } catch (\MySociety\TheyWorkForYou\MySociety\TheyWorkForYou\MemberException $e) {
+                } catch (\MySociety\TheyWorkForYou\MemberException $e) {
+                    // do nothing
+                }
+            }
+            $this->data['constituencies'] = $cons;
+        }
+
```
dracos, line 484: Something not right here - so if there's match on constituency name (`constituencies`) we loop through and find all the matching people, but then if there was a `member_constituencies` match (which presumably there would also be if constituencies matched), we throw all that away and do it all again?

classes/Utility/Postcode.php
```diff
@@ -178,4 +178,25 @@ public static function canonicalisePostcode($pc) {
...
+     */
+
+    public static function mapitTypeToHouse($type) {
+        $map = [
+            'WMC' => HOUSE_TYPE_COMMONS,
+            'SPC' => HOUSE_TYPE_SCOTLAND,
+            'SPE' => HOUSE_TYPE_SCOTLAND,
+            'WAC' => HOUSE_TYPE_WALES,
+            'WAE' => HOUSE_TYPE_WALES,
+            'LGU' => HOUSE_TYPE_NI,
```
dracos, line 192: I don't think LGU is a thing, NIE?

www/includes/easyparliament/templates/html/alert/index.php
```diff
@@ -237,143 +231,106 @@
...
-                    <?= sprintf(gettext('Sign up for an alert when %s speaks.'), $name) ?>
-                </h3>
-              <?php } elseif ($keyword) { ?>
-                <h3>
-                    <?= sprintf(gettext('Sign up for an alert when %s.'), _htmlspecialchars($display_keyword)) ?>
-                </h3>
-              <?php } elseif ($alertsearch) { ?>
-                <h3><?= gettext('Not quite right? Search again to refine your email alert.') ?></h3>
+              <div class="alert-section__header">
+              <h3><?= gettext('Create an alert when an MP, MSP or AM speaks') ?></h3>
```
dracos, line 289: Still not full list, sorry - MP, MSP, MS, MLA

classes/AlertView/Standard.php
```diff
@@ -438,10 +438,22 @@ private function searchForConstituenciesAndMembers() {
 
         if (isset($this->data['constituencies'])) {
             $cons = [];
-            foreach ($this->data['constituencies'] as $constituency) {
+            foreach ($this->data['constituencies'] as $type => $constituency) {
```
dracos, line 462: But isn't this all overridden by the member stuff? There's probably things about the canonical constituency names that mean it isn't, I guess, whereby something you search for is an alternative name which would then match constituency but no members. But this seems very convoluted - should we not a) canonicalise constituency name, the b) do member lookups - rather than both?

www/includes/easyparliament/templates/html/alert/_mp_alert_form.php
```diff
@@ -64,7 +64,7 @@
           <input type="text" class="form-control" name="mp_search" id="mp_search" value="<?= _htmlspecialchars($keyword) ?>">
         <?php } else { ?>
           <label for="mp-postcode">Search postcode, or MP name</label>
-          <input id="mp-postcode" type="text" class="form-control" placeholder="<?= gettext('e.g. ‘B2 4QA’ or ‘John Doe’') ?>" name="mp_search" id="mp_search" value="<?= _htmlentities($search_text) ?>" style="min-width:300px;">
+          <input id="mp-postcode" type="text" class="form-control" placeholder="<?= gettext('e.g. ‘B2 4QA’ or ‘John Doe’') ?>" name="mp_search" id="mp_search" value="<?= _htmlentities($search_term) ?>" style="min-width:300px;">
```
dracos: If I submit with this filled in and no email, the page now gives an error, but this is now blank. Not important, but just checking this was the right thing put here.

www/includes/easyparliament/templates/html/alert/_mp_alert_form.php
```diff
@@ -0,0 +1,114 @@
...
+          </button>
+          <?php } else { ?>
+          <button type="submit" class="button" name="mp_step" value="mp_search">
+            <span><?= gettext('Search') ?></span>
+            <i aria-hidden="true" class="fi-magnifying-glass"></i>
+          </button>
+          <?php } ?>
+          <button type="submit" class="button button--red" name="action" value="Abandon">
+            <i aria-hidden="true" class="fi-trash"></i>
+            <span><?= gettext('Abandon changes') ?></span>
```
dracos, line 92: It's still saying "Abandon changes" if I go to /alert/ and click the rep alert button...

dracos (changes requested):

www/includes/easyparliament/templates/html/alert/_mp_alert_form.php
dracos: (I think the setting this back to search_text might have been put in the wrong commit, the gettext one, but assume this will all be squashed/tidied anyway).

classes/AlertView/Standard.php
```diff
@@ -155,37 +448,135 @@ private function checkInput() {
...
+            $this->data['members'] = array_merge($members_from_words, $members_from_names);
+            [$this->data['constituencies'], $this->data['valid_postcode']] = \MySociety\TheyWorkForYou\Utility\Search::searchConstituenciesByQuery($text, false);
+            # if there's no postcode match then constituencies will only have a list of names of westminster constituencies
+            # so do a lookup using the member table to get members in all constituency types
+            if (!$this->data['valid_postcode']) {
+                $this->data['member_constituencies'] = \MySociety\TheyWorkForYou\Utility\Search::searchMembersByConstituency($text);
+            }
+            $member_count = count($this->data['members']);
+            $cons_count = count($this->data['constituencies']);
+            $mem_cons_count = isset($this->data['member_constituencies']) ? count($this->data['member_constituencies']) : 0;
```
dracos, line 494: `$mem_cons_count = count($this->data['member_constituencies'] ?? []);` perhaps?

dracos (approved):
Just in the new commit, saw a tiny thing. Thanks for explaining the things; it feels like this could be simplified, but you understand it and I'm sure if I spent more time with it I could also follow it, and has had enough :)

PR #1891, opened 2025-07-08
Limit imported policies to active policies

ajparsons:
We're currently importing the draft policies from votes.theyworkforyou.com.

The other file shouldn't display them in the first place - but we need to adjust the script here so it removes non-present items.

### DIFF
--- a/src/twfy_tools/utils/votes2db.py
+++ b/src/twfy_tools/utils/votes2db.py
@@ -177,6 +177,22 @@ def load_vote_distribution():
 
     df = pd.read_parquet(distribution_url)
 
+    starting_distribution_count = len(df)
+
+    active_policies = [
+        int(x) for x in Policy.objects.all().values_list("policy_id", flat=True)
+    ]
+
+    # filter to only active policies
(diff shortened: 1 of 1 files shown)


src/twfy_tools/utils/votes2db.py
```diff
@@ -292,11 +311,20 @@ def process_policy_json():
     to_update = [
         policy for policy in policies if policy.policy_id in existing_policy_ids
     ]
+    # remove any policies that are in the database but not in the json
+    # these are likely to be retired policies
+    to_delete_ids = [
+        policy.policy_id
+        for policy in Policy.objects.all()
+        if policy.policy_id not in [pol.policy_id for pol in policies]
+    ]
```
struan, line 320: This is fine but you could do this in the database with something like (I have not tested this ;):
```python
to_delete_ids = Policy.object.exclude(policy_id__in=[pol.policy_id for pol in policies]).values_list("policy_id", flat=True)
```
you might need to do something to make sure that's a list. Six and half a dozen and it's fine as it is. Could also construct the list of ids as you loop over data above to save you looping twice but all this is premature optimisation probably.

PR #1902, opened 2025-07-29
member interests page

### LINKED ISSUE #1895: Revised MPs Memberships page
Possibly framed as a "What is your MP interested in?" - this is giving more insight into MPs activities that are more individual and less party based (reflecting specific interests or concerns). 

A few elements that would be good to add/bring together in a new view:

- Select Committee Memberships
- APPG memberships
- Recent EDMs (via twfy-votes)
- Recent Open Letters (via twfy-votes)

This is combining parliamentary information with information we are building through other routes. 

# Select Commitees

Currently we display these on the main profile page:

Image

But what would be good here is to include more description about what each committee actually does (there are official descriptions we can use). 

# APPG memberships

We're working on building this APPG membership database, and it would be good to bring into this page. 

This probably lives in the personinfo table rather than a new membership information because we have very poor information about the start and end date of these memberships. 

https://github.com/mysociety/theyworkforyou/issues/1876

https://pages.mysociety.org/appg-membership/

# EDMs and Open Letters

We now have these in twfy-votes - and it would be good to link through. 

Parliament's site does for the EDMs like this: https://members.parliament.uk/member/4212/earlydaymotions

If we start thematic tagging of EDMs - we could create per person "what kind of EDMs do they often sign" and display that here. 

Open letters would be a new data source that is not on the parliament site.

struan:
Fixes #1895


ajparsons (changes requested):
A few small things to get the structure going at the start - but looks like it works ok!

Just to check, would splitting out the commitees so we can explain what each one is come in a subsequent PR after get the basic page working here?

ajparsons (approved):
Looks good!

I guess my one remaining question isn't a code one but a naming thing - this is about 'MPs interests' but also 'MPs interests' is often about MPs financial interests.

Is this a "Profile" page? The landing bit is now 'overview' - so could be.

PR #1904, opened 2025-08-07
Load and display MP Statments on interests tab

struan:
Displays 10 most recent EDM signatures from last three months and all open letter signatures from the past year that votes.twfy knows about.

Screenshot 2025-08-07 at 16 25 35

Part of #1895

### DIFF
--- a/classes/DataClass/Statements/Signature.php
+++ b/classes/DataClass/Statements/Signature.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * Mirrors pydantic model for deseralisation in a PHP context.
+ * For adding display related helper functions.
+ * @package TheyWorkForYou
+ */
+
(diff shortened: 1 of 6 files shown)


ajparsons:
Aside whatever's going on in test town - this looks good! 

My only structure comment is along the lines of the other PR: if you wanted to use define a pydantic structure for the output of `make_statment_object` (actually just spotted typo there). It wouldn't do anything substantial - but would be a bit clearer on the expected structure - or store the results in a DataClass on the php end. I think there's a bit less prospect of bringing more future data through so it's more consistency with the other PR.

Although that said, 'this person signed alongside x others' might be good to know? So possibly we want to store a value of 'total signatures' in here?

struan:
merged

PR #1905, opened 2025-08-22
Record emails sent in GA

ajparsons:
For better tracking of "hey how many emails did we send this month/year", we want to stash the amount of emails sent in google analytics.

This uses the measurement protocol to send events - the python version of this function is here: https://github.com/mysociety/research-repository/blob/7180805ee9522c4272019357df6de4c9dc738aa7/repository/views.py#L399

In the long run, we might want to add a few more stats from the alerts (keep track of keyword/representative ones separately) - but this gets the basic approach in.


struan (approved):
📈

Issue #1901, opened 2025-07-29
Too many sub-pages on representative page for menu structure


ajparsons:
We have this top menu structure for sub-pages on the representative page (which is a slow restructure from one very long page):

<img width="1018" height="410" alt="Image" src="https://github.com/user-attachments/assets/bb05de8b-97a6-49ac-bd85-9aca91a86e29" />

This already looks a bit rubbish on mobile and we want to add a few more pages and make this worse: 

<img width="512" height="117" alt="Image" src="https://github.com/user-attachments/assets/c60876b4-6081-4d4a-a155-16c7d1286272" />

I had a go at this in this PR https://github.com/mysociety/theyworkforyou/pull/1873, but it didn't go well, need a designer to have a think about what's actually the good approach here so we can add a few more views, and especially so it works ok on mobile.

PR #1907, opened 2025-09-02
New layout for MP Subpages

lucascumsille:
Fixes: https://github.com/mysociety/theyworkforyou/issues/1901

- [x] Add table of contents from other subpages
- [x] Add `Back to top` and `go to bottom` button
- [x] Mobile navbar for subpages
- [x] Update `aria-labels` for `nav` elements
- [x] I haven't been able to test the page `2024 election donations`

### Desktop preview:

https://github.com/user-attachments/assets/2f7a0aa4-b1c3-4cc8-8762-fe79a944ddd1

### Mobile preview:

https://github.com/user-attachments/assets/26a1b939-de6d-4b2f-88a4-9cca0a7930cb

### Notes

- I added a commit that allows users to scroll to the top and bottom of the page. Some of the subpages are really long, and personally, I always appreciate these buttons so I don’t have to scroll all the way up or down again.
- As we discussed, on mobile I have hidden the "Table of Contents" so that we only allow top navigation across the subpages.
- Let me know what you would prefer instead of "Browse Content" and, for the mobile button, "Table of Contents."
- Unfortunately I couldn't test the `2024 election donations` I get the following error:

Screenshot 2025-09-03 at 08 30 41

### DIFF
--- a/locale/TheyWorkForYou.pot
+++ b/locale/TheyWorkForYou.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-08-22 11:44+0000\n"
+"POT-Creation-Date: 2025-09-10 19:37+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -605,8 +605,8 @@ msgstr ""
 #: www/docs/interests/category.php:93
 #: www/includes/easyparliament/metadata.php:119
 #: www/includes/easyparliament/metadata.php:120
-#: www/includes/easyparliament/templates/html/mp/_person_navigation.php:15
-#: www/includes/easyparliament/templates/html/mp/register.php:51
+#: www/includes/easyparliament/templates/html/mp/_person_navigation.php:106
+#: www/includes/easyparliament/templates/html/mp/register.php:62
 msgid "Register of Interests"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


lucascumsille:
It looks good, Alex and I like the extra pages.

The one thing I'm a bit wary of is the up and down buttons at the moment. I find them useful for easily scrolling up and down in pages that are very long, for example, the voting summary, especially the up button once you have reached the bottom of the page and want to go back to the top with no effort. The current scenario is unclear about how much it actually helps with navigation, as it may scroll you down/up by almost nothing, depending on the section's length. I wonder if a feature like this could be more helpful for people with mobility issues, where scrolling takes more effort than pressing an area on the screen. But for a user without a mobility condition, I don't think it is that useful. I personally would just scroll.

ajparsons:
Oh I don't mind losing the up and down buttons altogther - i just thought the version where it just went all the way to the top and all the way to the bottom didn't really work. Should we just take out? the menu is a big improvement

ajparsons:
@struan : I've been reviewing this with Lucas but wanted to tag you in if you wanted to have a look.

www/includes/easyparliament/templates/html/mp/_person_navigation.php
```diff
@@ -50,7 +50,7 @@
         <?php if (array_key_exists('letters_signed', $memberships) || array_key_exists('edms_signed', $memberships) || array_key_exists('topics_of_interest', $memberships) || array_key_exists('eu_stance', $memberships)): ?>
             <li <?php if ($pagetype == "signatures"): ?>class="active"<?php endif; ?>>
                 <a href="<?= $member_url ?>/signatures" class="person-navigation--subpage-heading">
-                    <h2>✍️ <?= gettext('Signatures') ?></h2>
+                    <h2>✍️ <?= gettext('Signatures') ?>🆕</h2>
```
struan, line 53: I think these need some sort of spacing as they are a bit jammed against the text:

Screenshot 2025-09-15 at 14 23 13

struan (approved):
One small cosmetic change but otherwise looks fine

www/includes/easyparliament/templates/html/mp/_person_navigation.php
ajparsons: I'll actually just roll this back for now - and can re-add when we do the launch.

Issue #1892, opened 2025-07-17
Better display of lost whip information


ajparsons:
Losing the whip entirely is being used more frequently as a tool. This has historically been rarer and it doesn't have special handling. 

It would be good to show where an MP was elected as part of a party, and a link to any news story that gives more information.

Issues:

- We currently record all whip changes start_reason as changed_party - and this is one of the allowed enums in the database (this list of enums is not enforced by mysoc-validator).
- The popolo standard supports both 'source' and 'links' - the mysoc-validator tool supports source. Neither of these are in the member model.

A minimal change would be :

- add 'whip_removed', 'whip_restored' as valid reasons (this is different from 'changed_party' when someone resigns). 
- add a 'source' column to the member table and amend the ingest to pull it from people.json. 

We would then need to backtrack to add the richer data into people.json (I've been logging all of this in the PRs). 

This should then make all the information available for a small 'whip removed - read here for why' bubble.

PR #1918, opened 2025-09-15
Whip storage

ajparsons:
First part of https://github.com/mysociety/theyworkforyou/issues/1892

Makes basic changes to the membership table so it doesn't get upset if the source data using different entered/left reasons. 

Source info is not displayed anywhere at this point.


struan (approved):
looks good

PR #1919, opened 2025-09-16
Mp menu copy and styling

lucascumsille:
Hi @ajparsons I was testing this new page on my phone yesterday, it all seem to be working well, but I find the "Menu" label a bit confusing because it doesn't provide much context of what it does, especially considering it has the same copy as the one in the header.

I added two commits:

- One that add "MP" to the button, but "Representative" could also be a good alternative.
- The other commit adds an icon, to improve visibility, hopefully draws also more attention to the element.

Preview
Screenshot 2025-09-16 at 05 41 21

Screenshot 2025-09-16 at 05 38 34


ajparsons:
I've amended this slightly so the 'MP' is responsive the rep type. If tests pass i'll merge quickly.

PR #1920, opened 2025-09-16
add annotations permission to accounts

struan:
Adds a `can_annotate` boolean to users, plus updates the admin front end to allow editing this.

Also adds an `organisation` column to users and updates associated properties in user mode.

Fixes 1909

### DIFF
--- a/classes/User.php
+++ b/classes/User.php
@@ -68,6 +68,8 @@ class User {
         $data["status"] = $user->status();
         $data["facebook_id"] = $user->facebook_id();
         $data['facebook_user'] = $user->facebook_user();
+        $data['can_annotate'] = $user->can_annotate();
+        $data['organisation'] = $user->organisation();
         return $data;
     }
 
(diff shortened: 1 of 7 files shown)


classes/User.php
```diff
@@ -88,6 +90,9 @@ public function getUpdateDetails($this_page, $user) {
 
             $details["url"] = trim(get_http_var("url"));
 
+            $details["can_annotate"] = get_http_var("can_annotate") == "true" ? true : false;
```
ajparsons, line 93: I think this needs to be further down the file behind the `($this_page == "otheruseredit")` check.

ajparsons (changes requested):
## What works

- schema update runs good.
- a non superuser can't see this in their own edit screen.
- A superuser can edit a user, add it, it's there in the form after, can be removed.

## What's probably fine

- on my environment had to delete some stuff around facebook_user to get the forms to work- assuming that's all good on live.

## To change

Just need to run the linter and get the translation in for the tests - can use an auto-translation line because it's in an admin console. 

I've added one comment where I think the org/annotate save needs to be done  in a different place so it's covered by a superuser check. 

Just to check the logic, who can annotate: Moderators, Administrators, Superusers, any use who has the can_annotate option set. Do we have any moderator level accounts? This might be a good time to audit that quickly.

PR #1926, opened 2025-09-18
Use common register formatting

ajparsons:
This tidies up some different approaches, the register page on a representative page and in the category view will now share underlying templates, have the 'show new', etc buttons.


struan (approved):
no notes!

PR #1927, opened 2025-09-18
Add constituency info section for MPs

ajparsons:
This adds a much more prominent constituency info section for MPs - with the goal of linking and increasing traffic to the LIH.

@zarino : For your consideration on if you like this and what the copy should be.


zarino:
@ajparsons sure, go ahead!

I guess I wanted to make it super clear to people that, although the site looks like it’s all about climate and nature, there’s data there that’s more generally useful.

Feel free to edit as you wish :-)

PR #1938, opened 2025-10-01
editing interface for glossary entries

### LINKED ISSUE #1935: Add an editing interface for existing glossary entries

struan:
Fixes #1935


ajparsons:
So I haven't had time to go through the code, but at a 'does this work' level I've tried the various editing screens and the json export/import and it works as I'd expect. 

Quick thoughts:

- If you have permission to edit, can we add a link to the admin interface from the glossary? Just so it's all grouped there. 
- How hard would it be so the json import *could* do adding items? Like if the ID was set to -1, or there was an exact match on phrase? Just thinking adding a bunch at once might be something we do, so if it's simple would be nice to have.

dracos (changes requested):
Just some minor things

Issue #1937, opened 2025-09-24
Improve in page glossary UX


struan:
Instead of the full content of the entry being on a separate page have some way to view the entry without leaving the current page.

This will need a tiny bit of code to enable it in that we only add the first 80 characters of an entry to the link. The place to change this is in the `glossarise` function in `www/includes/easyparliament/glossary.php`. I guess we could stick the whole content in a data attribute? Will also need to run it through the markdown processor.

lucascumsille:
https://research.mysociety.org/html/improving-oversight/#different-forms-of-oversight-and-appeal

PR #1952, opened 2025-10-13
buttons and in page glossary definitions

struan:
Fixes part of: https://github.com/mysociety/theyworkforyou/issues/1937


struan:
superseded by #1955

PR #1955, opened 2025-10-15
Added Glossary drawer

### LINKED ISSUE #1937: Improve in page glossary UX

lucascumsille:
Preview:

https://github.com/user-attachments/assets/84fb7013-3ed0-4d45-ac51-f3c3c098aa57

Important: Markdown format is not available so at the moment is just plain text.

Fixes #1937

### DIFF
--- a/www/docs/js/main.js
+++ b/www/docs/js/main.js
@@ -868,6 +868,7 @@ class GlossaryDrawer {
       // Set initial ARIA attributes
       button.setAttribute('aria-expanded', 'false');
       button.setAttribute('aria-haspopup', 'dialog');
+      button.setAttribute('role', 'button');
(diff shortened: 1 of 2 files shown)


ajparsons:
Cool! Happy for that to go to struan

struan:
oh, whoops, my bad, I left if some debug.

PR #1961, opened 2025-10-16
Make banner button smaller on mobile

ajparsons:
I feel this is the slightly wrong approach but @lucascumsille might have a better one.

On mobile the banner button sits a bit weirdly:

image

This shunts it to a newline and shrinks a bit (but still bigger footprint than currently)

image

### DIFF
--- a/www/docs/style/sass/layout/_header.scss
+++ b/www/docs/style/sass/layout/_header.scss
@@ -428,13 +428,15 @@ $badge-horizontal-shadow-desktop: -8px;
             display: flex;
             flex-direction: column;
             align-items: center;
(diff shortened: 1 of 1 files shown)


lucascumsille:
Hey Alex
I made a couple of changes, but happy to keep it as you did. The only thing I replaced that probably is better to keep it that way is replacing `margin-top: 0.75em !important` with `row-gap: 0.5em` that accomplish the same thing without the `!important`, I decreased a bit the vertical spacing.

Screenshot 2025-10-21 at 09 25 12

ajparsons:
Lovely - I'm happy with that.

PR #1969, opened 2025-10-28
[alerts] Don't change step based on vector form

ajparsons:
https://github.com/mysociety/theyworkforyou/pull/1965 was along the right lines. - the issue was the form setting the step after that extra processing. 

This just adds an exception for that specific step - given the progression from this step is handled elsewhere.

### DIFF
--- a/classes/AlertView/Standard.php
+++ b/classes/AlertView/Standard.php
@@ -829,7 +829,10 @@ class Standard extends \MySociety\TheyWorkForYou\AlertView {
             $this->data['delete_token'] = $this->data['alerts'][0]['token'];
         }
         if ($this->data['addword'] != '' || ($this->data['step'] && count($this->data['errors']) > 0)) {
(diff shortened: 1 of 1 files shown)


dracos (changes requested):
This doesn't fix the issue, it just hides it. You only get here if "Add word" has been clicked, or there's an error. The error causing this to happen (with my example of add all on "disabled people") is "That search is too long for our database; please split it up into multiple smaller alerts." but this change just means that is ignored and you're taken to the review page anyway. And if you then create the alert, none of the extra options are included, only the main one.

Presumably it *should* be showing the same page if there's an error, so this bit of code is okay, and the issue is that the suggestions aren't being set up to be re-displayed (along with the error)? (so whatever https://github.com/mysociety/theyworkforyou/pull/1965 was meant to fix isn't, because the data isn't available at that point)

PR #1981, opened 2025-12-22
[api] Do not double escape json values

ajparsons:
We're storing more json values in the personinfo table - this stops double escaping these values when requested as an json export from the api by trying to decode it first.


struan (approved):
👍

PR #1991, opened 2026-02-23
Add export-translation script

ajparsons:
Quick LLM-assisted export tool to make working with translations a bit easier:

- Easier csv export to make back and forth with heloblod easier.
- Flag translations we've marked as auto (which feels like a good compromise to keep things going with small tweaks while doing updates in batches)


src/twfy_tools/utils/export_translation.py
```diff
@@ -0,0 +1,540 @@
...
+        table.add_row("Fuzzy", str(fuzzy_count))
+    if include_auto:
+        table.add_row("Auto-translated", str(auto_count))
+    table.add_row("Total", str(len(filtered_entries)))
+
+    print(table)
+
+
+@app.command()
+def import_translations(
```
struan: this could just be import to match export and be less wordy

struan (changes requested):
When I export and import it seems to introduce a lot of additional `msgid ""` and equivalent msgstr lines on anything that's a multiline string which I'm not sure is helpful and produces diffs that are hard to follow.

The json path seems to work, the csv side not so much. And modulo the above it seems to work fine.

PR #2003, opened 2026-04-09
Add featured annoucements to senedd/sp homepages

ajparsons:
Currently can only do this on main homepage, this adds some extra items for senedd and sp - do to be able to add a message about election.


struan (approved):
looks good

PR #2006, opened 2026-04-16
Track gift aid expansion

ajparsons:
I'd like to check how often this is being expanded to see if having the extra text box is discouraging people.


struan (approved):
looks good

PR #2007, opened 2026-04-16
Rep page donate banner

ajparsons:
Hopefully small tweak to auto expand banner on rep page. 

I've made this time limited to see if it makes a difference in the next week. 

This is optional if it over-complicates it , but a cookie is set to remember if this is minimised.


struan (approved):
Looks good

PR #2008, opened 2026-04-27
Regmem permalink

ajparsons:
Creates a permalink to an entry on a date in the register of interests

(linting problem unrelated, now fixed in master)

### DIFF
--- a/classes/DataClass/Regmem/Register.php
+++ b/classes/DataClass/Regmem/Register.php
@@ -60,7 +60,8 @@ class Register extends BaseModel {
         # sort most recent to least, find the first one before the date
         rsort($files);
         foreach ($files as $file) {
-            $file_date = basename($file, ".json");
+            # filename ends with isodate (10 characters - (while starting chamber is inconsistent length)
+            $file_date = substr(basename($file, ".json"), -10);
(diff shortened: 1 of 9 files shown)


classes/DataClass/Regmem/Register.php
```diff
@@ -60,7 +60,7 @@ public static function latestAsOfDate(string $chamber, string $date): Register {
         # sort most recent to least, find the first one before the date
         rsort($files);
         foreach ($files as $file) {
-            $file_date = basename($file, ".json");
+            $file_date = substr(basename($file, ".json"), -10);
```
struan, line 64: a comment to say what this is doing might be nice.

struan (approved):
Could this also get added on the mp page? I don't think we should hold this up but certainly something to look at in future.

classes/DataClass/Regmem/Register.php
ajparsons: Done!

ajparsons:
I decided it was worth doing that quickly so the approach was tidy for both - see what you think but works fine for me.

struan:
looks good

PR #2017, opened 2026-05-17
Use Plaid Cymru's green for the party dot

### LINKED ISSUE #2004: Update colour used for Plaid Cymru
Currently, Plaid Cymru is represented by a yellow colour on TWFY, this is not consistent with the colours used by other sites e.g. BBC, Parliament UK

**MP on TWFY:**

Image

**MP on Parliament UK**

Image

The Green Party also use the colour green so images such as the one below may become difficult to interpret but I guess the same could be said with the existing colour and the Lib Dems

Image

c-tonneslan:
Closes #2004.

Swaps the Plaid Cymru party dot from `rgb(224, 134, 26)` (orange) to `rgb(0, 129, 66)`, the party's official Pantone 348C green. That's what Parliament UK and the BBC use too.

@SarahmySociety noted the trade-off vs the Green Party already; the new shade is darker than the existing Green Party `rgb(106, 176, 35)` so they're still distinguishable next to each other.

### DIFF
--- a/www/docs/style/sass/pages/_search.scss
+++ b/www/docs/style/sass/pages/_search.scss
@@ -595,7 +595,7 @@ table.search-results-grouped {
   }
 
   &.plaid-cymru:before {
-    background-color: rgb(224, 134, 26);
+    background-color: rgb(0, 129, 66);
   }
 
   &.green:before {


ajparsons:
Before closing this one we should open an equiv issue in [twfy-votes-django ](https://github.com/mysociety/twfy-votes-django)

c-tonneslan:
Opened mysociety/twfy-votes-django#101 for the equivalent change in division dots.

ajparsons:
Sorry that was more of a note for me on the other issue! Looks good, thanks for having a look.

PR #2018, opened 2026-05-17
Fix 'seperate' typos in two comments

c-tonneslan:
Caught these while skimming the source. Two unrelated comments, same misspelling.


ajparsons:
Super, thanks for this.

PR #2021, opened 2026-06-01
Allow empty/missing categories list in regmem

ajparsons:
Mismatch between the regmem json format and the PHP parser - we weren't allowing an empty categories list under a person (something to investigate in the data generator, but it's valid under the schema). 

This sets a null default and adds a new helper function to ensure there's always a list (even if empty).

### DIFF
--- a/classes/DataClass/Regmem/Person.php
+++ b/classes/DataClass/Regmem/Person.php
@@ -21,8 +21,14 @@ class Person extends BaseModel {
     public string $person_id;
     public string $person_name;
     public string $published_date;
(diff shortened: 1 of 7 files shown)


dracos (approved):
Should the two internal `->categories` in Person.php also be changed to `getCategories()`, or can it be guaranteed it'll be set when they're called?

(Oh, actually, I see neither are actually used anywhere, so guess it doesn't matter, but still, should they be removed then)

ajparsons:
I've updated those as well (and a few other references).

PR #2020, opened 2026-05-21
Added Newsletter sign up form to MP page

lucascumsille:
Fixes: https://github.com/mysociety/theyworkforyou/issues/2012

### Desktop

Screenshot 2026-05-21 at 08 48 15

### Mobile preview
https://github.com/user-attachments/assets/77a2c8a5-0f72-4b1e-81c0-fa6b320f96db

### Json example
```
{
  "title": "Love what we do?",
  "content": "Be the first to find out what's coming next.",
  "thumbnail_image_url": "https://www.mysociety.org/files/2026/03/wdtk-image-event.jpg",
  "thumbnail_image_alt_text": "",
  "button_text": "Subscribe",
  "button_class": "button--primary",
  "location": ["sidebar_newsletter"],
  "footer_html": "<a href=\"https://www.mysociety.org/privacy#newsletter\">Your data and how we use it</a>",
  "weight": 3,
  "published": true,
  "lang": "en",
  "start_time": "2025-01-01T00:00:00",
  "end_time": "2030-12-31T23:59:59"
}
```

PR #2022, opened 2026-06-03
Register of interest page layout

### LINKED ISSUE #2015: [2009] On the Register of Interests page:
Add donate box in right hand column, below current anchors:

**Support TheyWorkForYou.** 
We are a charity - donate today to invest in making UK politics more transparent and accessible. Even £5 goes a long way! Donate Now ♥️"
 
Add a newsletter prompt below it: 

**Find out what else we do.** [Subscribe to our newsletter] 

- If possible, the signup box is right on the page, not a click away
- it adds the 'TWFY' and 'democracy' tags in Mailchimp, thus kickstarting the [TWFY welcome journey](https://us9.admin.mailchimp.com/customer-journey/builder?id=186&stepModal=trigger).

lucascumsille:
Fixes: https://github.com/mysociety/theyworkforyou/issues/2015

The following changes have been made:

- Register of interest page uses the same layout as the MP page. So we don't have two different styles that at the same cover similar layout out content.
- Move the styling from `in-page-nav` to its own stylesheet. So we can reuse it in other pages.
- Added a component `.js-table-of-content-markdown` that creates from the markdown a table of content with the proper classes so it can use the current functionality/styles of `js-table-of-content`
- Added a newsletter and donation prompts to the sidebar, just like in the MP. Because the copy is different I decided to create new templates instead of all using the current templates.

Screenshot 2026-06-03 at 10 32 51
Note: the odd styling inside the newsletter should be fixes when merging [this previous PR needs](https://github.com/mysociety/theyworkforyou/pull/2020)


ajparsons:
So mostly looks good!

- This will need a bit of changing after merging my changes to the other PR - as that changes the approach for specifying something as a newsletter page.
- The static page donation/newsletter bubble are fairly similar to the ones in the MP page - I wonder about moving the display bit to sidebar/_donation.php etc - and then include that after loading the specific data for each location?

lucascumsille:
> The static page donation/newsletter bubble are fairly similar to the ones in the MP page - I wonder about moving the display bit to sidebar/_donation.php etc - and then include that after loading the specific data for each location?

Just to clarify the styling is the same, however the content/copy is different. I'm happy to go with this approach if we can have different content in the  MP and Register of Interest.

> This will need a bit of changing after merging my changes to the other PR - as that changes the approach for specifying something as a newsletter page.

Perfect I'll rebase this one with master once we have merge this [PR](https://github.com/mysociety/theyworkforyou/pull/2020) and make the changes.

lucascumsille:
@ajparsons I just added two commits that follows the same logic for the announcements in the MP page.

I also included some commits to:

- Keep the announcements from disappearing on mobile, instead they'll stack.
https://github.com/user-attachments/assets/e47bb3a8-15ef-4ff4-8960-95f51d1702ba

- I noticed we had a bug for the reusability of the announcements where the variables defined at the very top of donation and newsletter keep overriding what I had in the sidebar.php for static pages.

lucascumsille:
@ajparsons I think this should be ready for review:
JSON Example:

```
  {
    "title": "Support TheyWorkForYou",
    "content": "We are a charity - donate today to invest in making UK politics more transparent and accessible. <br> ♥️ Even £5 goes a long way!",
    "url": "https://www.mysociety.org/donate/",
    "button_text": "Donate now",
    "button_class": "button--negative",
    "location": ["sidebar_interests"],
    "type": "donation",
    "weight": 2,
    "published": true,
    "lang": "en",
    "start_time": "2025-01-01T00:00:00",
    "end_time": "2030-12-31T23:59:59"
  },
{
  "title": "Find out what else we do",
  "content": "Subscribe to our newsletter",
  "thumbnail_image_url": "https://www.mysociety.org/files/2026/03/wdtk-image-event.jpg",
  "thumbnail_image_alt_text": "",
  "url": "https://www.mysociety.org/subscribe/",
  "button_text": "Subscribe",
  "button_class": "button--primary",
  "location": ["sidebar_interests"],
  "type": "newsletter",  
  "footer_html": "<a href=\"https://www.mysociety.org/privacy#newsletter\">Your data and how we use it</a>",
  "weight": 3,
  "published": true,
  "lang": "en",
  "start_time": "2025-01-01T00:00:00",
  "end_time": "2030-12-31T23:59:59"
}
```

ajparsons:
This renders the new things ok, but I get this vertical sidebar in chrome on both the markdown and the MP pages
image

I've added two fix-up commits - one removes a 1px thing that stops the scrollbar (but may have been there for another reason?) the other fixes a problem where the recent votes page had the scrollbar on the right. 

If that makes sense to you - just merge in the fixups and it's good.

lucascumsille:
@ajparsons I just squashed those commits, they look good to me. But I could not replicate you issue with the vertical visible scrollbar. I tried on Chrome and Safari, I also tried limiting reducing the max-height and adding more items to the list, but I couldn't replicate.

lucascumsille:
@ajparsons let me know if you are happy for me to merge into master =)

ajparsons (approved):

PR #2023, opened 2026-06-08
Motion to annul text

ajparsons:
Explanation for this one also in the voting summaries update doc - but basically we're now including a few substantically similar EDMs that are not *technically* motions to annul.


struan (approved):
👍

PR #2024, opened 2026-06-08
Update redis settings

ajparsons:
Addressing a set of (it turns out unrelated) problems in docker and the test suite. 

- moved the docker to use the same bitnamisecure repo as the tests.
- for both added ALLOW_EMPTY_PASSWORD to let the sentinel start


dracos (approved):
It'd be good to have your explanation as to the change in the commit message

PR #2028, opened 2026-06-16
Added "Data, coding and development" to mailchimp form in footer

lucascumsille:
Fixes: https://github.com/mysociety/Design-Internal-tasks/issues/56

This only applies when users sign up on the API page(`/api`)


ajparsons:
Not sure on a hidden field - we usually would have it as a pre-checked tickbox that was visible to add to a segment

lucascumsille:
Screenshot 2026-06-22 at 06 28 20

I just added a fixup for the feedback above. Let me know if we need to tweak anything else.

lucascumsille:
@ajparsons I forgot to tag you.

ajparsons (approved):
all good when squashed!

PR #2029, opened 2026-06-19
Switch to current period view

ajparsons:
Change to set the default period for the voting summaries to the current 'labour 2024-'' gov.


struan (approved):
This looks good. One small thing is the # links by the topic headers don't respect the comparison period so if you click one of those you always get the default period.

PR #2031, opened 2026-06-23
Add memcache to alerts page query

ajparsons:
- Cache results for 6 hours to speed up refresh on pages with many alerts


struan (approved):
looks good

PR #2032, opened 2026-06-23
Better start/end memberships

dracos:
Use a new 'grouped memberships' (which joins together actual memberships if the ~~constituency is the same and~~ end/start dates are close enough) to cover distinct periods.
Use these in most places to allow better showing of 'current' start dates rather than only earliest start date.
Should fix Andy Burnham displaying as Labour MP for Makerfield ( 7 Jun 2001 – current) on search results page.

* Doesn't do anything with getEnterLeaveStrings which could be changed to loop through grouped_memberships and output all the relevant lines.
* ~~Makes things little bit worse (I guess) for everyone whose constituency changed in 2024 due to the boundary changes, because it now gives their start date as 2024 rather than when they actually started. But that does list the constituency name normally, so I guess is still accurate in that sense?~~

### DIFF
--- a/classes/AlertView/Standard.php
+++ b/classes/AlertView/Standard.php
@@ -563,7 +563,7 @@ class Standard extends \MySociety\TheyWorkForYou\AlertView {
             foreach ($this->data['member_constituencies'] as $pid) {
                 try {
                     $MEMBER = new \MySociety\TheyWorkForYou\Member(['person_id' => $pid]);
-                    $cons[] = [ 'member' => $MEMBER, 'constituency' => $MEMBER->constituency, 'rep_name' => $MEMBER->getMostRecentMembership()['rep_name'] ];
+                    $cons[] = [ 'member' => $MEMBER, 'constituency' => $MEMBER->constituency, 'rep_name' => $MEMBER->getMostRecentGroupedMembership()['rep_name'] ];
                 } catch (\MySociety\TheyWorkForYou\MemberException $e) {
                     // do nothing
                 }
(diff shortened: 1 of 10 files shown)


ajparsons:
It's tricky, I think the date range thing for constituency changes would make it slightly worse (technically accurate, but it's a useful 'how long have they been an MP thing).

I wonder if grouped memberships should be "loosely continuously member of the same chamber" - so jump through the role to the chamber_id? (This would need some different code for Lords - because the chamber conneciton is direct without a role)

The way to make it clear it's not about the specific would be something like:

Labour MP for Makerfield (MP from 2001 to 2016, 2026 onwards)
Labour MP for MadeUp (MP from 2001 to 2016)

Which is solving it just by stuffing more information in - but that might be the right thing to do when someone's done something complicated. What do you think?

Just doing the grouped membership change and not changing the text would I think improve burnham without massively altering others? (even if it doesn't catch the historical bit)

dracos:
> Just doing the grouped membership change and not changing the text would I think improve burnham without massively altering others? (even if it doesn't catch the historical bit)

Not sure what you mean, sorry; how would I do the change without changing the text? It's changing e.g. getEntryDate and getMostRecentMembership - getEntryDate is definitely improved because that's only used on the "recently elected MP on DATE" bit, and getMostRecentMembership dates is only used on search results page I think - so that would be better with the multiple ranges as you mention.

(Not going to ask why we have single use functions like this that rely upon context for understanding what they're for, makes it harder to follow through from source to output.)

I've removed the constituency check when joining in grouped_memberships, which I think does what you ask (so same house, less than 3 months apart membership => group them), no need for different code for Lords, it's already per-house.
It will show date ranges for same-house, so e.g. if someone was MP 2001-2005, MSP 2007-2011 and MP 2019-2024 it would only say (MP 2001-2005, 2019-2024) but that seems fine.

image

ajparsons:
Yep that's all looking good to me.

The only thing that might help us out in future is to rename getMostRecentMembership to getMostRecentGroupedMembership - so it's clear that's not actually grabbing the last membership (in case we try and use that function in something else in future). Nothing that currently uses it seems to do anything that this issue would break. 

(also don't think this one matters - but could drop down to just month year (Jun 2001) to save some space)

PR #2035, opened 2026-06-25
Group voting summaries by periods

ajparsons:
This PR adds a bit more time structure to the voting policies - the goal of this is to make recent votes easier to discover, give some structure to long running MPs, while reflecting that long ago votes can still be impactful today. 

- policies are grouped into five year periods (we don't have exact dates avalaible for start and end of policies, so can't sync exactly to parliaments). 
- start and end date displayed are adjusted depending on entry dates

Here's an example:

image

### DIFF
--- a/classes/PolicyDistributionCollection.php
+++ b/classes/PolicyDistributionCollection.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Party Cohort Class
+ * Policy Distribution Collection Class
  *
  * @package TheyWorkForYou
  */
@@ -9,7 +9,7 @@
 namespace MySociety\TheyWorkForYou;
 
 /**
- * Policy DistributionCollection
+ * PolicyDistributionCollection
  * This brings together a set of a person's policy distributions for a given period and party.
  * It covers items in the same 'group' (health, social, etc)
  */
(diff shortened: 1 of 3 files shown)


classes/PolicyDistributionBucket.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * Party Cohort Class
```
struan: comment needs updated

classes/PolicyDistributionBucket.php
```diff
@@ -0,0 +1,41 @@
...
+namespace MySociety\TheyWorkForYou;
+
+/**
+ * Policy DistributionBucket
+ * A set of policy pairs that share a recency window (based on the year the
+ * member most recently voted on the policy). Used to add a time structure
+ * beneath each policy group in the all-time voting summary.
+ */
+
+class PolicyDistributionBucket {
```
struan: Not sure about this class name. Doesn't really describe what it's doing. `PolicyDistributionYearGroup` or something similar might be a bit less generic.

classes/PolicyDistributionCollection.php
```diff
@@ -49,6 +49,35 @@ public function __construct(
...
+        // Votes can't be in the future, and shouldn't extend past when the
+        // person left office.
+        $current_year = (int) date('Y');
+        $cap_high = $person_end_year ? min($person_end_year, $current_year) : $current_year;
+
+        $buckets = [];
+        foreach ($this->policy_pairs as $pair) {
+            $year = $pair->member_distribution->end_year
+                ?: $pair->member_distribution->start_year;
+            $block = intdiv($year, $bucket_size) * $bucket_size;
```
struan: could we call this $bucket_start_year to be a bit more descriptive of that this is doing? Using block/bucket interchangeably is also a bit confusing.

struan (approved):
few small naming/comment things but otherwise looks good

ajparsons:
Yep, done those naming changes - makes sense.

PR #2036, opened 2026-06-25
Add voting appeal specific content

ajparsons:
This adds a specific voting summary update appeal that takes priority over the current 20% coverage banner for the time.

There's a blank link that'll need to be updated to a blog post when that's published (before this is deployed)

image


struan (approved):
If we're going to do more of these it might be sensible to pull all the boiler plate out to it's own file to save redundant code

PR #2041, opened 2026-07-07
fix alert form when linked to with pid and logged out

### LINKED ISSUE #2040: creating an alert from representative page does not work when logged out
If a logged out user clicks on the "Get email updates" button on the representative page, the "Subscribe" button on the following page does not work. It reloads the page but does not create an alert or ask for the user's email address so it can create an alert.

struan:
This now correctly asks you for your email address. Also update the MP page links to add the correct parameter to get you to directly to the right form

Fixes #2040

### DIFF
--- a/tests/AlertsPageTest.php
+++ b/tests/AlertsPageTest.php
@@ -30,6 +30,13 @@ class AlertsPageTest extends FetchPageTestCase {
         $this->assertStringContainsString('<input type="text" id="words0" name="words[]" aria-required="true" value="elephant"', $page);
     }
 
+    public function testPidOnly() {
+        $page = $this->fetch_page([ 'pid' => '2']);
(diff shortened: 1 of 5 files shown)


dracos (approved):
This fixes the issue, but as I think I've said before, I find this code very confusing and hard to follow, which seems odd for what it's then doing. Why is there a $members loop in two different places? When do you get to the one in alert/index.php or the one in alert/_mp_alert_form.php?
Would this be simplified if, I dunno "create rep alert" was at /alert/create-rep/ and "create keyword" was at /alert/create-keyword/ and /alert/ was just viewing/managing your existing alerts?

And just noticed that if there's an MP/constituency result when you come from a search results page, it only lets you sign up for them, there's no keyword search. Should we add a link to let you go to a keyword sign up as well, in case that's what you wanted?

PR #2044, opened 2026-07-19
Dev tooling, dependencies, linting.

Br3nda:
HELLO! I've been working for OpenAustralia on long-overdue maintenance of the Australian fork. We actually forked before git was invented, back in SVN/CVS days, but are looking into being a bit closer to the UK edition.

I'd like to share some housekeeping and "developer quality of life" improvements upstream to you. 

 - linting the Perl
 - versions in mise, to assist developers having the same versions of Ruby, PHP, Python, and Perl for this code (no messing with system packages)
 - cpan file to record Perl dependencies.
 - script to install dependencies

We hope you'll accept this.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,3 +55,12 @@ jobs:
           .php-cs-fixer.cache
           tools/php-cs-fixer
         key: ${{ runner.OS }}-${{ github.repository }}-phpcsfixer74
+
+    - name: Install Perl dependencies
+      run: cpanm --installdeps --with-develop .
+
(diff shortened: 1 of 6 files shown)


dracos (changes requested):
Hello! Thanks for this :) Looks good.

Could `install-deps` be instead called `bootstrap` to match our use in other repos and https://github.com/github/scripts-to-rule-them-all?

I think perlcritic looks for pl/pm files automatically, so could the lines that run that, run `perlcritic scripts search` instead? Otherwise it feels like it's going to produce a lot more output than we'd want :)

PR #2048, opened 2026-08-10
Find correct Senedd dir for registers

ajparsons:
We've been getting errors when an exact data is being queries for a senedd register.

This started happening because we added permalinks from the MS pages - but the helper function that map this to the right file didn't pick up on the language folders for senedd registers.

This is now consistent across all three times the path is constructed.


struan (approved):
🏴󠁧󠁢󠁷󠁬󠁳󠁿

PR #2050, opened 2026-08-12
Bookworm - PHP8.2

ajparsons:
This branch updates the docker to bookwork and mariadb 11.8.

A few small fixes needed but generally seems ok. 

The Claude scan for the string to number change is [here](https://docs.google.com/document/d/1L5PsusydGnPBuRN6YyU01PuG4dvLiaH_S7yeXl4jaeI/edit?tab=t.0) - a few things flagged but I haven't made any changes due to it because it all seems fine from our point of view. That's not conclusive obviously, but suggests we're not going to find massive issues due to that change.


scripts/mprss.php
```diff
@@ -9,7 +9,7 @@
 $rsspath = BASEDIR . '/rss/mp/';
 
 // Make things group writable.
-umask(002);
+umask(0o02);
```
sagepe, line 12: Shouldn't this be `0002` not `0o02`?

Gemfile.lock
```diff
@@ -25,4 +25,8 @@ PLATFORMS
   ruby
 
 DEPENDENCIES
+  chunky_png (>= 1.4)
   compass
+
+BUNDLED WITH
+   2.3.15
```
sagepe: I think we usually remove the `BUNDLED WITH` lines.

scripts/mprss.php
ajparsons: No that's right - it's a new convention for flagging it's a octal number - in principle no difference between how the old and the new is processed.  https://www.php.net/manual/en/migration81.new-features.php#:~:text=Integer%20Octal%20Literal%20Prefix%20%C2%B6

Gemfile.lock
ajparsons: go for it!

.github/workflows/default.yml
```diff
@@ -45,62 +54,59 @@ jobs:
...
+          --health-retries=5
 
     steps:
-#    - name: Stop any running MySQL
-#      run: sudo systemctl stop mysql.service
-#    - uses: chhsiung/github-actions-mysql@0f348ad54f14038adafed5914b5ce775806768ab
-#      with:
-#        rootPassword: 'password'
-#        sqlMode: 'ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'
+      - uses: niden/actions-memcached@v8
```
dracos, line 63: Can we use pinned commit hashes here, if we're updating this file anyway

dracos (commented):
One request only!

PR #2052, opened 2026-08-13
Add script to reconcile keys and quotas in Redis.

neprune:
For https://github.com/mysociety/sysadmin/issues/2125.

### DIFF
--- a/scripts/reconcile-redis.php
+++ b/scripts/reconcile-redis.php
@@ -1,0 +1,204 @@
+#!/usr/bin/php
+
+<?php
+
+// Ensures API keys and subscription quotas are configured correctly
+// in Redis.
+// Will also delete subscriptions in DB if stripe says they are cancelled.
+// Pass '--commit' to run for real.
+
+include_once dirname(__FILE__) . '/../www/includes/easyparliament/init.php';
+use Predis\Collection\Iterator\Keyspace;
+
+$commit = ($argv[1] ?? '' == '--commit');
+if ($commit) {
+    echo "Running for real.\n\n";
+} else {
+    echo "Dry run. Pass --commit to run for real.\n\n";
+}
+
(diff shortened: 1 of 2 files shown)


neprune:
Dry run against live: 
* 6 API keys in Redis with no DB record that would be deleted.
* 2 quotas in Redis with no DB record that would be deleted (seemed related to testing).
* All but 10 stripe subs were 'active'. These were 'incomplete_expired'. Most have active API keys and also quota set-up in Redis, but at least wouldn't have had their counts reset. I _think_ this is because stripe isn't firing `customer.subscription.deleted` in this scenario and we're not checking the status in our 'updated' hook logic.

`incomplete_expired` scenario needs a proper look but for now just updating this script to delete these as it already does for `canceled`.

scripts/reconcile-redis.php
```diff
@@ -0,0 +1,204 @@
...
+
+echo "\nGetting all API keys in DB.\n";
+
+$q = $db->query("SELECT api_key, user_id FROM api_key");  # Including disabled ones.
+
+$db_api_keys = [];
+foreach ($q as $api_key) {
+    $key = $api_key['api_key'];
+    $user_id = $api_key['user_id'];
+    if (!($key && $user_id)) {
```
dracos, line 51: Not sure this ever happens

dracos (approved):
Looks good, yeah, guess we must have assumed deleted event always got sent? Not sure. Presumably MapIt has the same potential issue?

neprune:
> Looks good, yeah, guess we must have assumed deleted event always got sent? Not sure. Presumably MapIt has the same potential issue?

I don't recall seeing any subs in `incomplete_expired` when doing the dry run in MapIt, but not 100%. Have created https://github.com/mysociety/societyworks/issues/5667 to take a look properly at some point.

PR #2055, opened 2026-09-01
Ongoing php 8.2 fixes

ajparsons:
The simpler ones in the latest batch (the alerts one looks more fiddly)


dracos (approved):
Guess hope nothing bad passed to sectionToTitle (presumably unlikely), as looks like blank string would end up in a join(' or ', ...) type thing.

PR #2056, opened 2026-09-02
speaker affiliations are often duplicated on TWFY

samsmith:
dedup affiliations of speakers

Doesn't fix it in the database, but that's not easy to find without state, so simply dedup at display


dracos:
Thanks for letting us know. This variable is also used in e.g. www/includes/easyparliament/templates/api/hansard_gid.php so better to be fixed in the _get_speaker_offices that creates it - hopefully that's done in 80f742ffc

PR #2057, opened 2026-09-10
Alerts creation can handle OR in section

ajparsons:
Currently the alerts creation process does not work when doing OR keywords and putting it in a section.

`a or b section:c` - effectively is `a OR (b section:C)`

We did have a function that fixed this by wrapping the keywords in brackets, but it was only applied as part of the preview interface. 

This PR moves it to the common conversion function (also simplifying some of the logic). 

Once live we'll need a script to adjust historical ones (have a candidate for this)


dracos:
This has been merged as 5e291c7ec

PR #2058, opened 2026-09-10
Stop error on allowed length query

ajparsons:
In principle we've allowed much longer queries since d7010246a4956ec088bcdb3ae169afbb557a339c

But we still had the error here which is my best candidate for where the errors are triggering in the alerts process (appearing after the related terms stage - which is a reasonable place for them to get longer).

Unless there's something I'm missing - it's safe to expand this. 

If we continue to get errors - need to have a go at getting the errors to at least surface in the right place.


struan (approved):
👍

